PR #220, opened 2015-08-28
Scrollable toplayer

matzimowski:
Changes in a toplayer:
- the toplayer is scrollable when its higher than browser height
- the toplayer vertical position is always 5px below mint-header (thanks to previous change the toplayer and its content is always available) => discussed with @JanuszSzewczyk 
- removed --centered-modal modifier - no cases of usage
- an overlay uses flexbox
- the toplayer as modal (pop-up) has to be located always inside overlay element

screen shot 2015-08-28 at 10 32 42
screen shot 2015-08-28 at 10 32 48

closes https://github.com/brainly/style-guide/issues/199


src/docs/examples/notlogged_toplayer.html
```diff
@@ -6,42 +6,43 @@
 
 </head>
 <body>
-<div class="mint-toplayer mint-toplayer--modal mint-toplayer--medium mint-toplayer--lead mint-toplayer--with-bugbox">
+<div class="mint-overlay">
```
matzimowski: i just put mint-toplayer inside mint-overlay

kdzwinel:
lgtm!

src/docs/examples/notlogged_toplayer.html
mr-mig: Why?

mr-mig:
> removed --centered-modal modifier - no cases of usage

Why?

matzimowski:
@mr-mig 
discussed with @JanuszSzewczyk 
position of all of our modals gonna be always the same => below mint-header element, we won't use this --centered-modal modifier.
if requirements change we will implement it.
wdyt?

src/docs/examples/notlogged_toplayer.html
matzimowski: @mr-mig thanks to this only one element has ```position:fixed```, one element has to be removed on close and we can use flexbox for positioning elements inside an overlay.

matzimowski: what's more,```flexbox``` is suppprted by Opera Mini, ```transform``z is not.

mr-mig:
I see, makes sense. :+1:

src/docs/examples/notlogged_toplayer.html
mr-mig: Smart move, :+1:

kdzwinel:
ack

PR #221, opened 2015-08-28
Added parameters to skip rdoc installation for ruby gems in Dockerfile

mr-mig:
... so that we do not download and build docsets during image creation


matzimowski:
:+1:

kdzwinel:
ok

Issue #237, opened 2015-09-02
Conditional html-class including


matzimowski:
Condtional html-class creating:
sth like:
```
$include-html-classes: true !default;

@if ($include-html-classes) {
    .mint-element {
    }
}
```

PR #238, opened 2015-09-02
Conditional including of html-classes

matzimowski:
closes https://github.com/brainly/style-guide/issues/237
`?w=1` makes life easier

### DIFF
--- a/src/components/bubble/_bubble.scss
+++ b/src/components/bubble/_bubble.scss
@@ -8,9 +8,9 @@ $borderSize: 3px;
 $triangleSize: 12px;
 $innerTriangleSize: $triangleSize - 2 * $borderSize + 1;
 
-$include-html-classes: true !default;
+$include-html: false !default;
 
-@if ($include-html-classes) {
+@if ($include-html) {
 
   .mint-bubble {
     @include component;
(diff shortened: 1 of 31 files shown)


src/components/bubble/_bubble.scss
```diff
@@ -8,48 +8,54 @@ $borderSize: 3px;
...
-    border: $triangleSize solid transparent;
-    border-bottom-color: $bubbleBorderColor;
-    position: absolute;
-    top: -(2 * $triangleSize);
-    right: 0;
-    left: 0;
-    width: 0;
-    margin: 0 auto;
-  }
+$include-html-classes: true !default;
```
mr-mig: Should it be moved to the main file?
It will defined whether partials should be included

matzimowski: we were thinking about that but we decided to make each file includable separately and do to that this variable has to be defined in each file.
@mr-mig wdyt?

mr-mig: Why ?

mr-mig: So, after discussion with @matzimowski @kdzwinel we've decided that this variable should be `false` by default. It will be set to true in `main.scss` in styleguide.

In other cases it will result in a very handy `import`, where you can be sure that you won't get classes from styleguide baked-in in your component.

Implementation details up to @matzimowski

mr-mig:
LGTM

kdzwinel:
OK

matzimowski:
:lollipop:

PR #239, opened 2015-09-02
Add new component  - mint-box . Closes #181

### LINKED ISSUE #181: Create box with border and header (see example)
screen shot 2015-08-18 at 11 47 07

kdzwinel:
screen shot 2015-09-02 at 13 49 51

As you can see in the "example usage" the spacing is wrong, but that's an issue of content-box (we have a separate ticket for that).

### DIFF
--- a/src/components/box/_box.scss
+++ b/src/components/box/_box.scss
@@ -1,0 +1,14 @@
+$boxBorderColor: $graySecondary;
+
+.mint-box {
+  @include component;
+
+  border-radius: 22px;
+  padding: $layoutDefaultPadding;
+  border: 2px solid $boxBorderColor;
+
+  &--full {
+    width: 100%;
+  }
+}
+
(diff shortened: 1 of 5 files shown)


src/components/box/_box.scss
```diff
@@ -0,0 +1,14 @@
+$boxBorderColor: $graySecondary;
+
+.mint-box {
+  @include component;
+
+  border-radius: 22px;
```
vergilius: extract to variable? <extract until you drop>

vergilius:
:+1:

src/components/box/_box.scss
matzimowski: :+1: for var here

matzimowski:
:+1:

src/components/box/_box.scss
mr-mig: I am wondering why it's `22px`. Does it have something related to `$defaultLayoutPadding`?

mr-mig:
:+1:

src/components/box/_box.scss
kdzwinel: Asia told me that `22px` and `10px` are only two valid values for `border-radius`. I don't think there any correlation with `defaultLayoutPadding`.

Issue #13, opened 2015-07-02
Vertical rhytm


mr-mig:
It would be really cool to define vertical rhytm in the styleguide.

You can find an example here:
http://www.sitepoint.com/using-sass-build-custom-type-scale-vertical-rhythm/
http://typecast.com/blog/4-simple-steps-to-vertical-rhythm

We will need this stuff as a baseline for **mobile-first components** in BEM.

Example:
```scss
.bem-block {
  @include base-font(); // <--- this should define the vertical rhythm in the component with relevant media queries
}
```

PR #254, opened 2015-09-07
Feature/vertical rhythm

### LINKED ISSUE #192: EM/REM for text and paddings
http://trentwalton.com/2012/06/19/fluid-type/
https://bugsnag.com/blog/responsive-typography-with-rems

See the example: http://jsfiddle.net/25gCt/4/embedded/result/

### LINKED ISSUE #230: Normalize font-family usage
See article: http://www.clockwork.net/blog/2012/11/29/616/fixing_the_font_mess_in_your_css_font-face_and_the_font-family_font-weight_saga

Will be breaking change.

mr-mig:
**WIP: will not be merged to master**
I will proceed with the changes as soon as the feature-branch will be merged to dev-branch

Old text styles:

New styles:

- Headers

- Text & links

### Mockup

http://typecast.com/QhQQkkddGr/share/87eeaf5d967424f871d1236edad36ee3d46a85b6KrPQgjrbw

### Changes 
In this set of changes I have introduced the changes to headers + text.

It relates to #13, closes #192 and closes #230.
Please, see #192 for some important info and decisions.

These are **breaking changes** already.

These things MUST be refactored across all the projects:

```
mint-text-block -> mint-text--obscure
mint-text-description -> mint-text--standout
mint-text-description--large -> mint-headline
mint-text-description--large + bold -> mint-header-primary--small
```

These things have changed it dimensions and MAY need refactoring:

```
mint-header-primary -> mint-text-bit
mint-header-primary--small -> mint-text-bit--small
mint-text-emphasised -> mint-text--emphasised
```

#### baseline push
I've tried to *push text to the baseline*, but it does not really worked out:

1. It's a bit ugly in containers
2. The push amount is inconsistent (need manual tweaking), even when `cap-height` equation is used

Tried the things described here:

1. https://medium.com/written-in-code/aligning-type-to-baseline-the-right-way-using-sass-e258fce47a9b
2. https://gist.github.com/razwan/10662500
3. https://jakegiltsoff.co.uk/posts/sassline-v2-0

I tried `cap-height=0.57` (taken from sassline) and `cap-height=0.67` (measured manually). Neither has worked out well for both text and headers simultaneously.

The problem IMO is that those articles assume that `line-height` for EVERY text element is **exact natural multiplier** of `$baseline` (1,2,3,4 ...), which is not true in many cases.

#### Headeline + secondary headers break rhythm

To have a good-looking **wrapped** header we MUST have non-natural multipliers for `line-height` for some header sizes. 

**This is a thing, which breaks the rhythm.**

I am planning to address this issue directly in containers by introducing the `min-height` property for header-related holes.

Any other suggestions?

#### Shall we remove `mint-text--emphasised` et al.? 

The reason is that we can actually make an exception and treat text as `user-generated content`. 
So, we will have **scoped** styles inside mint-text block: `<strong>`, `<em>`, `<i>`, `<b>`...

It makes sense for me, but WDYT?

#### Non pixel-perfect links

`<a>` seems to add some pixels to line box. You can see this in links example it docs - the latter lines are offseted from baseline.

I think we can live with this issue for now, as I have not figured out the reason for this behavior and how to fix it properly.

### DIFF
--- a/src/components/badges/_badges.scss
+++ b/src/components/badges/_badges.scss
@@ -4,7 +4,6 @@ $badgeDefaultColor: $white;
 $badgeLightBackgroundColor: $white;
 $badgeLightColor: $black;
 
-
 $include-html: false !default;
 
 @if ($include-html) {
@@ -13,10 +12,10 @@ $include-html: false !default;
     @include component;
 
     border-radius: $badgeFontSize;
-  	line-height: $badgeFontSize;
-  	font-size: $badgeFontSize;
-  	font-weight: bold;
-  	padding: 3px 6px;
+    line-height: $badgeFontSize;
+    font-size: $badgeFontSize;
+    font-weight: bold;
+    padding: 3px 6px;
 
     background-color: $badgeDefaultBackgroundColor;
     color: $badgeDefaultColor;
--- a/src/components/badges/_badges.scss
+++ b/src/components/badges/_badges.scss
@@ -1,9 +1,9 @@
-$badgeFontFamily: $fontFamilySecondary;
 $badgeFontSize: 10px;
 $badgeDefaultBackgroundColor: $peachPrimary;
 $badgeDefaultColor: $white;
 $badgeLightBackgroundColor: $white;
 $badgeLightColor: $black;
+
 
 $include-html: false !default;
 
@@ -13,10 +13,10 @@ $include-html: false !default;
     @include component;
 
     border-radius: $badgeFontSize;
-    line-height: $badgeFontSize;
-    font-size: $badgeFontSize;
-    font-family: $badgeFontFamily;
-    padding: 3px 6px;
+  	line-height: $badgeFontSize;
+  	font-size: $badgeFontSize;
+  	font-weight: bold;
+  	padding: 3px 6px;
 
     background-color: $badgeDefaultBackgroundColor;
     color: $badgeDefaultColor;
@@ -43,5 +43,4 @@ $include-html: false !default;
       transform: scale(1);
     }
   }
-
 }
--- a/src/components/text/_headers.scss
+++ b/src/components/text/_headers.scss
@@ -14,7 +14,7 @@ $include-html: false !default;
     }
 
     &--light {
-      color: $white;;
+      color: $white;
     }
   }
 
@@ -33,7 +33,7 @@ $include-html: false !default;
     }
 
     &--light {
-      color: $white;;
+      color: $white;
     }
   }
 
--- a/src/components/text/_headers.scss
+++ b/src/components/text/_headers.scss
@@ -10,14 +10,14 @@ $include-html: false !default;
     font-weight: $fontWeightBlack;
 
     &--small {
-      font-size: $headerPrimarySmallFontSize;
+      @include type-variant(1.75, 2, 1, 8/16);
     }
 
     &--light {
       color: $white;;
     }
-
   }
+
   .mint-header-secondary {
     @include component;
     @include type-variant(1, 0.667, 12/16, 20/16);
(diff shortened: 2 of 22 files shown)


src/components/badges/_badges.scss
```diff
@@ -15,7 +14,7 @@ $include-html: false !default;
     border-radius: $badgeFontSize;
     line-height: $badgeFontSize;
     font-size: $badgeFontSize;
-    font-family: $badgeFontFamily;
+    font-weight: bold;
```
kdzwinel: We haven't used `bold` anywhere AFAIK and we shouldn't do that - we have a separate family (ExtraBold) for that.

src/sass/_fonts.scss
```diff
@@ -3,24 +3,24 @@ $include-html: false !default;
...
   @font-face {
-    font-family: 'ProximaNovaBlack';
+    font-family: 'ProximaNova';
     src: url($mintFontsPath + 'ProximaNova-Black.woff') format('woff');
-    font-weight: bold;
+    font-weight: $fontWeightBlack;
     font-style: normal;
   }
-
+
```
matzimowski: love it :dancer:

src/components/badges/_badges.scss
kdzwinel: Ah, got it - font-weight changes the font family :+1:

aju:
@mr-mig what are we gaining changing to "rem" units? Are we planning to play with body font-size?

src/components/text/_headers.scss
```diff
@@ -1,50 +1,39 @@
...
+    font-weight: $fontWeightBlack;
 
     &--small {
-      font-size: $headerPrimarySmallFontSize;
+      @include type-variant(1.75, 2, 1, 8/16);
     }
 
     &--light {
-      color: $headerColorLight;
+      color: $white;;
```
kdzwinel: double `;`

kdzwinel:
`mint-header-primary -> mint-text-bit` - how header became a text-bit? Text-bit's are always uppercase, right? Headers are not (AFAIK)?

vergilius:
In some places you're using font-weight explicitly, and in others you're taking it from a variable ;)

mr-mig:
@vergilius Cause there's "bold" value and "500" value. I do not want to specify `500` everywhere

mr-mig:
Maybe I can use "bold" and "bolder" though...

mr-mig:
@kdzwinel It's not about the header becomes the text bit, it's mainly about the **sizes**.
The header font size and style in the old styleguide is used only it `text-bit`, so its not a `header`. 
We do not have headers like that. So the styles have migrated to `text-bit`

mr-mig:
@aju we have this idea in mind, that we can change the font-sizes depending on viewport size.

mr-mig:
So, I've added the fix for links.

**The problem**: links are `inline` elements in a separate container, and they have space for *descenders* added. As soon as line-height is **smaller** than baseline - link boxes will be aligned with the text without descenders overlow

src/components/text/_headers.scss
```diff
@@ -1,50 +1,39 @@
...
-$headerSecondaryFontSize: 16px;
-$headerSecondarySmallFontSize: 12px;
-
 $include-html: false !default;
 
 @if ($include-html) {
 
   .mint-header-primary {
     @include component;
+    @include type-variant(3, 2, 11/16, 37/16);
```
aju: I think we have all functions camel case and we decided not to make this push trick for now.

mr-mig: Yes and yes, I'll fix it

mr-mig: Hm, can I fix those things in the next PR actually?
Otherwise I will have to cherry-pick a couple of commits in between branches, which is a pain in the ass TBH.

matzimowski: it's fine for me. let's merge it :+1:

vergilius: :+1: yep let's go with it

vergilius:
good job :+1:

PR #256, opened 2015-09-07
Mustard badge added. Fixes #253

### LINKED ISSUE #253: Add mint-badge mustard modificator
screen shot 2015-09-04 at 17 15 58

kdzwinel:
`mint-badge--alt`:
screen shot 2015-09-07 at 16 47 45

and in action:
screen shot 2015-09-07 at 16 47 36

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "4.3.0",
+  "version": "4.4.0",
   "description": "Brainly Front-End Style Guide",
   "author": "Brainly",
   "private": true,
--- a/src/components/badges/_badges.scss
+++ b/src/components/badges/_badges.scss
@@ -1,11 +1,11 @@
 $badgeFontFamily: $fontFamilySecondary;
 $badgeFontSize: 10px;
-$badgeDefaultBackgroundColor: $peachPrimary;
-$badgeDefaultColor: $white;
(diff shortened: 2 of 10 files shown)


mr-mig:
I do not like `alt`. It seems non-semantic for me

kdzwinel:
We were discussing it with Michal and couldn't come up with anything better. We could call it 'mustard' reusing color name, but that's also imperfect. Any ideas?

mr-mig:
alert?

mr-mig:
Let's steal from masters: http://v4-alpha.getbootstrap.com/components/label/
`success`, `info`, `warning`, `error`, `alert`, etc

vergilius:
I have feeling that in future we might need badge-per-basic-color, and using bootstrap's names might be limiting, WDYT @mr-mig?

kdzwinel:
I'm also wondering if our default here is OK (by default badge is red/peach), maybe the default badge should be black? And then it would make some sense to use `error` for red and `warning` for yellow.

aju:
@kdzwinel +1 for the idea

kdzwinel:
names updated. Now the default badge is white and there are two variations: `--error` and `--warning`.

aju:
:ok:

vergilius:
:+1:

matzimowski:
nice branch name :older_man:

Issue #259, opened 2015-09-08
Dropdown for filters


matzimowski:
<img width="344" alt="screen shot 2015-09-08 at 14 45 06" src="https://cloud.githubusercontent.com/assets/1231144/9735063/495a6bf8-5638-11e5-8e77-036265f787c9.png">
<img width="167" alt="screen shot 2015-09-08 at 14 45 22" src="https://cloud.githubusercontent.com/assets/1231144/9735062/4959755e-5638-11e5-8b4e-294b65790525.png">

PR #265, opened 2015-09-09
Feature/dropdowns

matzimowski:
closes https://github.com/brainly/style-guide/issues/259

updated again:
screen shot 2015-09-10 at 13 02 29

### DIFF
--- a/src/components/dropdowns/_dropdowns.scss
+++ b/src/components/dropdowns/_dropdowns.scss
@@ -32,7 +32,6 @@ $include-html: false !default;
       position: absolute;
       top: $dropdownHeight/4 - 2;
       right: $dropdownHeight / 2 - $iconSize / 2;
-      pointer-events: none;
       color: $grayAlt;
       z-index: 1;
       transition: transform 0.3s;
--- a/src/components/dropdowns/_dropdowns.scss
+++ b/src/components/dropdowns/_dropdowns.scss
@@ -32,17 +32,16 @@ $include-html: false !default;
       font-size: $iconSize;
       font-weight: bold;
       position: absolute;
-      top: $dropdownHeight/2 - 1;
+      top: $dropdownHeight/4 - 2;
       right: $dropdownHeight / 2 - $iconSize / 2;
       pointer-events: none;
       z-index: 1;
-      transform: translateY(-50%);
     }
 
     &__label {
       display: flex;
       align-items: center;
-      height: $dropdownHeight;
+      height: 100%;
       padding: 0 $dropdownHeight 0 $dropdownHeight/2;
     }
 
@@ -81,11 +80,12 @@ $include-html: false !default;
       height: auto;
 
       &:before {
-        transform: translateY(-50%) rotate(180deg);
+        transform: rotate(180deg);
       }
 
       .mint-dropdown__label {
         color: $black;
+        height: $dropdownHeight - 2;
       }
 
     }
--- a/src/components/dropdowns/_dropdowns.scss
+++ b/src/components/dropdowns/_dropdowns.scss
@@ -21,8 +21,11 @@ $include-html: false !default;
     height: $dropdownHeight;
     text-transform: uppercase;
     font-family: $fontFamilySecondary;
-    box-shadow: 0 1px 0 $graySecondary;
     cursor: pointer;
+
+    &--full {
+      width: 100%;
+    }
 
     &:before {
       $iconSize: $dropdownFontSize + $iconBoostValue;
--- a/src/components/dropdowns/_dropdowns.scss
+++ b/src/components/dropdowns/_dropdowns.scss
@@ -1,0 +1,27 @@
+// your components variables
+
+$include-html: false !default;
+
+@if ($include-html-classes) {
+  .mint-dropdown {
+    background-color: $grayAltLight;
+    box-sizing: border-box;
+    padding: 10px $layoutDefaultPadding/3*2;
+
+    &__label {
+
+    }
+
+    &__list {
+
+    }
+
+    &__element {
+
+      &--active {
+
+      }
+
+    }
+  }
+}
--- a/src/components/dropdowns/_dropdowns.scss
+++ b/src/components/dropdowns/_dropdowns.scss
@@ -48,10 +48,12 @@ $include-html: false !default;
     }
 
     &__element {
+      color: $grayAlt;
       display: flex;
       align-items: center;
       height: 36px;
       padding: 0 16px;
+      text-decoration: none;
 
       &:hover,
       &:active {
--- a/src/components/dropdowns/_dropdowns.scss
+++ b/src/components/dropdowns/_dropdowns.scss
@@ -1,7 +1,6 @@
 // your components variables
 $dropdownFontSize: 12px;
 $dropdownHeight: 32px;
-$dropdownElementHeight: 36px;
 $dropdownBorderSize: 2px;
 
 $include-html: false !default;
@@ -15,7 +14,7 @@ $include-html: false !default;
     background-color: $grayAltLight;
     border: $dropdownBorderSize solid $grayAltLight;
     border-radius: 16px;
-    min-height: $dropdownHeight;;
+    min-height: $dropdownHeight;
     height: $dropdownHeight;
     cursor: pointer;
     @include hole;
--- a/src/components/dropdowns/_dropdowns.scss
+++ b/src/components/dropdowns/_dropdowns.scss
@@ -1,25 +1,76 @@
 // your components variables
+$dropdownFontSize: 12px;
+$dropdownHeight: 32px;
 
 $include-html: false !default;
 
-@if ($include-html-classes) {
+@if ($include-html) {
+
   .mint-dropdown {
+    @include component;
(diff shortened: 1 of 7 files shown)


matzimowski:
one sec, i will and --full modifier

src/components/dropdowns/_dropdowns.scss
```diff
@@ -0,0 +1,90 @@
...
+    color: $grayAlt;
+    background-color: $grayAltLight;
+    border: 2px solid $grayAltLight;
+    border-radius: 16px;
+    font-size: $dropdownFontSize;
+    line-height: 1;
+    height: $dropdownHeight;
+    text-transform: uppercase;
+    font-family: $fontFamilySecondary;
+    box-shadow: 0 1px 0 $graySecondary;
```
aju: Check select elements and make it the same

matzimowski: @aju, discussed with @JanuszSzewczyk , we gonna get rid of box-shadox here.

mr-mig:
:+1:

src/components/dropdowns/dropdowns.html
```diff
@@ -0,0 +1,61 @@
+<section class="docs-block">
+    <aside class="docs-block__info">
+        <h3 class="docs-block__header">Dropdowns</h3>
+    </aside>
+    <div class="docs-block__content">
+        <div class="mint-dropdown">
+            <div class="mint-dropdown__label">All options</div>
```
mr-mig: Can it be just a `mint-label`?

src/components/dropdowns/_dropdowns.scss
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    &:before {
+      $iconSize: $dropdownFontSize + $iconBoostValue;
+      font-size: $iconSize;
+      font-weight: bold;
+      position: absolute;
+      top: $dropdownHeight/2 - 1;
+      right: $dropdownHeight / 2 - $iconSize / 2;
+      pointer-events: none;
```
mr-mig: What is it for?

src/components/dropdowns/_dropdowns.scss
```diff
@@ -0,0 +1,93 @@
...
+      z-index: 1;
+      transform: translateY(-50%);
```
mr-mig: Why do you need both top and translate here?

src/components/dropdowns/dropdowns.html
matzimowski: @mr-mig,
i could use `mint-label` here and i'd have to add modifier with another color, it's a lil bit different. is it okay?

src/components/dropdowns/dropdowns.html
```diff
@@ -0,0 +1,84 @@
...
+            <ul class="mint-dropdown__list">
+                <li>
+                    <a class="mint-dropdown__element" href="#">Option 1</a>
```
mr-mig: Is there any reason for element to be link and not just span/div?

src/components/dropdowns/_dropdowns.scss
matzimowski: it's for arrow, it's almost the copy of arrow from select,
maybe i could prepare one mixin for both?

src/components/dropdowns/dropdowns.html
mr-mig: Yep, I think it is ok. What do you think about `inactive` or `disabled`?

src/components/dropdowns/_dropdowns.scss
matzimowski: you're right, i will update it

src/components/dropdowns/dropdowns.html
matzimowski: they do not have to be link but main usage for these dropdowns are filters, and filters have to be links, i.e.  `/subject/maths` and so on.

src/components/dropdowns/dropdowns.html
matzimowski: Hm, but this label are not inactive. there's gonna be default label or chosen by user.

matzimowski: @mr-mig i just checked it, there is different font-family for mint-label and for label in dropdown (here is the same as in selects), i am not sure anymore if i should use `mint-label` here

vergilius: @mr-mig @matzimowski I just checked labels, I would suggest to update `mint-label` font-family here to extrabold instead black? Its more readable, and doesn't have big impact ie in search page.
d1a948ee96f6e12b7adb9f034e43b01e

matzimowski: @vergilius maybe you are right..

matzimowski: @JanuszSzewczyk agreed, we should update label's font-family, then i could create new modifier:)
i created an issue: https://github.com/brainly/style-guide/issues/266

mr-mig: You should talk to Janusz/Joanna regarding this change

src/components/dropdowns/dropdowns.html
mr-mig: I am wondering whether we can make this dropdown a container in this case. 
You will be able to use `mint-link` as element. And for a standard dropdown you can use just text block/label for items

mr-mig: As far as I can understand, the item with the subject icon + text is a **label** inside a link...

aju:
@matzimowski I would make whole thing a container (I mean options)

src/components/dropdowns/dropdowns.html
vergilius: @mr-mig you still can put just anything into `<li />` element, how is this different from container?

matzimowski:
@mr-mig @aju @vergilius 
we did refactor with @kdzwinel .
check screenshot and review it again, from the beginning, please.
:older_man:  => :baby: :baby_bottle:

mr-mig:
Text centering is different between closed and open dropdown

mr-mig:
Also, dropdown in the interactive example usage without subject icon looks really weird

src/components/dropdowns/_dropdowns.scss
```diff
@@ -0,0 +1,105 @@
...
+      visibility: hidden;
+      padding: 0;
+      margin: 0 0 $layoutDefaultPadding/2;
+    }
+
+    &__item-hole {
+      display: flex;
+      align-items: center;
+      justify-content: flex-start;
+      height: 100%;
```
mr-mig: Try adding min-height here instead of 100% to prevent the strange look for options without subject icons

src/components/dropdowns/_dropdowns.scss
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+
+    &__items {
+      visibility: hidden;
+      padding: 0;
+      margin: 0 0 $layoutDefaultPadding/2;
+    }
+
+    &__item-hole {
```
mr-mig: JFYI:
All `holes` confuse me when I work with VR.
I think we may reconsider how they are used sometime.

matzimowski:
@mr-mig i did update of screenshot, it was a lil bit broken.
anyway, this one:
screen shot 2015-09-10 at 13 01 04
and this one:
screen shot 2015-09-10 at 13 01 08

are the same.

screen shot 2015-09-10 at 12 59 00

these elements have to be larger, because there are clickable (also on touch devices), it's was designed in this way.

src/components/dropdowns/_dropdowns.scss
matzimowski: check it now, please, previous screenshot (do not know why - was broken), i just updated it.

mr-mig:
@matzimowski 
What's the difference between the wide and narrow dropdown?

On the latest screenshot small dropdown text in not centered relatively to the toggle.


I's suggest to replace it with label. Label has font-centering issue solved.
There's no sense of having simple text as a label for dropdown anyway.

mr-mig:
@matzimowski after screenshot update the options look good!

src/components/dropdowns/_dropdowns.scss
mr-mig: Yep, it's ok

kdzwinel:
LGTM

matzimowski:
updated screenshot:

screen shot 2015-09-10 at 15 14 36

mr-mig:
:+1:

vergilius:
:+1:

kdzwinel:
For the record - I think that having labels is wrong - we should have minimal examples for all components. But if everyone else disagrees, you can marge it.

aju:
In this case I think we can keep label as this will be default use case for dropdown

kdzwinel:
Go to that branch and `push origin --force` it

matzimowski:
@kdzwinel it's already merged

kdzwinel:
yes, I know but if you pushed the latest version of this branch that you have locally to github it would close automatically + it would close issue #259 automatically. Without that we have to do it manually.

matzimowski:
@kdzwinel thanks for info, i didn't know that.

Issue #244, opened 2015-09-03
What benefits do we get by not referencing variables directly in components?


mr-mig:
Why do we do `myComponentFontFamily: $fontFamilyPrimary`?

aju:
Font family might be bad example as I wrote in other ticket we should make it global.
Other than that referencing for example colours and is really cool and make it easy to refactor components (you can see all important setting for component in one place)

mr-mig:
For me, the mental burden for finding `color: $white` is much less than adding a variable for each changeable parameter

vergilius:
This works fine (referencing variables directly from config), until you have to put it in 2 or more places.
Imho we should stick to having additional variables as it allows to easilly see what config variables, colors are used in component.

kdzwinel:
I agree with Michal, these variables give a quick overview of the color palette used by the component + they clearly state that "the same color should be used in these two (or more) places".

aju:
@mr-mig can we keep then this convention and close this issue?:) (We can see in some time if we still agree on this in the future)

mr-mig:
So, summing up:

1. Gives the overview for all **changeable parameters** for a given components
1. Allows easily **reuse** the variable
1. Allows **referencing** parameters of other components
1. Makes refactoring easier

aju:
@mr-mig maybe its a good idea to put this in documents to explain why we do it that way?

mr-mig:
@aju its in Contributing PR #273

aju:
thx!

PR #283, opened 2015-09-11
Create mint-popup-menu. Fixes #246

### LINKED ISSUE #246: User panel
Consider adding an user panel from symfony to the style-guide.
imho, only one thing that is worth to add is this blue container with triangle at the top.

screen shot 2015-09-03 at 13 33 25
screen shot 2015-09-03 at 13 50 35

kdzwinel:
screen shot 2015-09-11 at 17 28 45
screen shot 2015-09-11 at 17 28 06

### DIFF
--- a/src/components/popup-menu/popup-menu.scss
+++ b/src/components/popup-menu/popup-menu.scss
@@ -1,0 +1,38 @@
+$userPanelBackground: $blueSecondary;
+$userPanelPadding: $layoutDefaultPadding;
+
+$include-html: false !default;
+
+@if ($include-html) {
+  .mint-popup-menu {
+    position: relative;
+    width: 100%;
+    padding: $userPanelPadding/3 $userPanelPadding/2;
+    background-color: $userPanelBackground;
+    display: flex;
+    justify-content: space-around;
+    align-items: center;
+
+    &:before {
+      content: "";
+      border: $userPanelPadding/2 solid transparent;
+      border-bottom-color: $userPanelBackground;
+      position: absolute;
+      top: -$userPanelPadding;
+      right: 18px;
+      width: 0;
+      margin: 0 auto;
+    }
+
+    @include hole();
+
+    &__hole {
+      flex-grow: 1;
+      margin-right: $userPanelPadding/2;
+
+      &:last-child {
+        margin-right: 0;
+      }
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


src/components/popup-menu/popup-menu.scss
```diff
@@ -0,0 +1,38 @@
+$userPanelBackground: $blueSecondary;
+$userPanelPadding: $layoutDefaultPadding;
+
+$include-html: false !default;
+
+@if ($include-html) {
+  .mint-popup-menu {
+    position: relative;
+    width: 100%;
+    padding: $userPanelPadding/3 $userPanelPadding/2;
```
mr-mig: I am still confused by how do we chose what parameters go into variables?

Why `width:100%` is not `width: $userPanelWidth`, but `padding:$userPanelPadding/3` and not `padding: $layoutDefaultPadding/3`?

mr-mig:
Should there be any space between avatar & nickname & points?

src/components/popup-menu/popup-menu.scss
vergilius: Hm, for simplicity I would use variables only when they might be needed outside, or used in more than one place in file.

vergilius:
imho they are more like holes that can take anything inside, the margins should be done with something more generic

kdzwinel:
@mr-mig yeah, Michal is correct, avatar/nickname/points is an example content inside a hole. `mint-popup-menu` does not control it. This needs to be fixed with row element / grid, but we don't have it right now.

src/components/popup-menu/popup-menu.scss
mr-mig: We have it discussed here (and made a decision): https://github.com/brainly/style-guide/issues/244

Is the decision there incorrect or not full? WDYT @vergilius @kdzwinel

mr-mig:
@kdzwinel I see, make sense. :+1:

src/components/popup-menu/popup-menu.scss
kdzwinel: > Why width:100% is not width: $userPanelWidth

hm, good point - I treat `width:100%` as an equivalent of saying "take all horizontal space". I don't expect anyone to change value "100%", either this line exists or not. That's why I haven't created a `$userPanelWidth ` variable.

> but padding:$userPanelPadding/3 and not padding: $layoutDefaultPadding/3

`$userPanelPadding` exists because I'm using it in multiple places in that component, if someone decides to change its value to `$userPanelPadding: layoutDefaultPadding / 2` it will change all dependent paddings proportionally.

kdzwinel: @mr-mig ping!

Issue #268, opened 2015-09-10
Implement panel containers


vergilius:
We need containers with header,list,footer for panels like notifications, messages etc

![image](https://cloud.githubusercontent.com/assets/1142999/9814561/6013c99e-588f-11e5-9499-d09d63405959.png)

aju:
We solved it using existing container.

PR #288, opened 2015-09-14
Content list

vergilius:
screen shot 2015-09-14 at 14 48 50
screen shot 2015-09-14 at 14 48 19

- not sure about gray header,
- not sure about list item

Ref #268


kdzwinel:
I have couple of issues with that:

# 1
IMO that element should be composed out of column+row+content-box like so:

elements

We don't have row and column yet though.

# 2 
`mint-content-list-item` is a separate element, it should be in a separate section (not together with `mint-content-list `). However, since it's very specific to `mint-content-list` (even the name reveals that), IMO it should be a `mint-content-list__item` instead of `mint-content-list-item`.

# 3

`mint-content-list__header` centers the text (`text-align: center;`) - we decided that we should avoid influencing inner styles of the "holes".

vergilius:
@kdzwinel cool, I'll rebuild this according to suggestions

vergilius:
closed, this should be done via list-item (refactor)

PR #296, opened 2015-09-16
Make box-sizing:border-box global. Fixes #260

### LINKED ISSUE #260: Add box-sizing:border-box to basics
Add box-sizing:border-box into basic styles for * elements.
wdyt @mr-mig @aju @vergilius @kdzwinel

kdzwinel:
Added border-box globally, removed it from components.

Only thing that broke are iframes in the styleguide, but I replaced default iframe border with a shadow:

screen shot 2015-09-16 at 09 46 51

screen shot 2015-09-16 at 12 10 50

### DIFF
--- a/src/components/footer/_footer.scss
+++ b/src/components/footer/_footer.scss
@@ -9,7 +9,6 @@ $include-html: false !default;
 
     &__container {
       position: relative;
-      box-sizing: border-box;
       padding: $layoutDefaultPadding $layoutDefaultPadding;
 
       @media (min-width: $breakpointTablet) {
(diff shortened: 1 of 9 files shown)


kdzwinel:
Question - is that a breaking change? It doesn't break styleguide but it may break a website where it's included.

mr-mig:
If you use the thing in `html`, than it is a breaing change.
If you put it in `component` mixin - then it may be not breaking change.

LGTM anyway

aju:
Its breaking change but lets introduce it. @mr-mig it was before in component mixin :)

PR #308, opened 2015-09-17
Add mintBreakpoint mixin

### LINKED ISSUE #294: Add mint-breakpoint mixin
Should work like:
`@include mint-breakpoint(tablet-only)` === `@media (min-width: $breakpointMedium) and (max-width: $breakpointLarge - 1px)`

aju:
closes #294

### DIFF
--- a/src/components/content-box/_content-box.scss
+++ b/src/components/content-box/_content-box.scss
@@ -34,7 +34,7 @@ $include-html: false !default;
 
     &--spaced {
       .mint-content-box__content, .mint-content-box__footer {
-        @media (min-width: $breakpointMedium) {
+        @include mintBreakpoint('medium-up') {
           padding-left: 2 * $layoutDefaultPadding;
         }
       }
--- a/src/components/content-box/_content-box.scss
+++ b/src/components/content-box/_content-box.scss
@@ -34,7 +34,7 @@ $include-html: false !default;
 
     &--spaced {
       .mint-content-box__content, .mint-content-box__footer {
-        @include mintBreakpoint('medium-up') {
+        @include mintBreakpoint(medium-up) {
           padding-left: 2 * $layoutDefaultPadding;
         }
       }
--- a/src/sass/_mixins.scss
+++ b/src/sass/_mixins.scss
@@ -87,3 +87,14 @@
 @function rhythm($baselines) {
   @return $baselines * $baseline;
 }
+
+@mixin mintBreakpoint($name) {
+  $media: map-get($breakpointsMap, $name);
+  @if($media) {
+    @media #{$media} {
+      @content;
+    }
+  } @else {
+    @content;
+  }
+}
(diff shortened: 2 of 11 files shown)


vergilius:
:sunny: :sparkles: :+1:

src/components/content-box/_content-box.scss
```diff
@@ -34,7 +34,7 @@ $include-html: false !default;
 
     &--spaced {
       .mint-content-box__content, .mint-content-box__footer {
-        @media (min-width: $breakpointMedium) {
+        @include mintBreakpoint('medium-up') {
```
mr-mig: Do you think single quotes + `-` are better than no quotes and `_` (`medium_up`) or even `mediumUp`

src/sass/_mixins.scss
```diff
@@ -87,3 +87,14 @@
...
 }
+
+@mixin mintBreakpoint($name) {
+  $media: map-get($breakpointsMap, $name);
+  @if($media) {
+    @media #{$media} {
+      @content;
+    }
+  } @else {
+    @content;
```
mr-mig: It should better throw some error. 
You do not want to let some typo ruin the behavior

src/components/content-box/_content-box.scss
aju: I will check if we need to pass it with quotes at all.

src/sass/_mixins.scss
aju: Yeah good idea

aju:
@mr-mig updated

src/components/content-box/_content-box.scss
```diff
@@ -34,7 +34,7 @@ $include-html: false !default;
 
     &--spaced {
       .mint-content-box__content, .mint-content-box__footer {
-        @media (min-width: $breakpointMedium) {
+        @include mintBreakpoint(medium-up) {
```
mr-mig: Should it be `camelCase` actually?

aju: This is not variable this is breakpoint name and to follow naming convention we have I think its better to keep it like this.
eg: `.mint-hide-for-small-only`
WDYT ?

mr-mig: But we do not have a convention for that. 
I do not see a good reason to make this thing and exception from naming convention.

aju: so you think in here it would be better to have it in camel case and just in classes use it like hyphen ? (like small-only) ?

mr-mig: Yep, as with all other classes.
Classes are not keys in the map.

mr-mig: Ideally, I would call variables with dashes, to have it consistent with class names, but this is a different story.

mr-mig: Anyway, its up to you. You can merge :+1:

aju: https://github.com/brainly/style-guide/issues/310 Lets move it here

PR #309, opened 2015-09-17
Refactor scripts

aju:
I have updated "server" and "watch" scripts to set proper container names for docker. I also added removing of container for "watch" script after hitting CTRL+C (it ). 
I also refactored 'ide-run-scss-lint-from-docker.sh' to get rid of duplicated code.

### DIFF
--- a/scripts/ide-run-scss-lint-from-docker.sh
+++ b/scripts/ide-run-scss-lint-from-docker.sh
@@ -1,9 +1,3 @@
 #!/usr/bin/env bash
-
-# This script will run scss-lint
-
-PROJECT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd)
-
-docker run --rm \
-  -v $PROJECT_DIR:/style-guide \
-  brainly/style-guide bash -c "scss-lint /style-guide/$1 -f JSON | sed 's/\/style-guide\///g'"
+SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+source ${SCRIPT_DIR}/run-scss-lint.sh $1 -f JSON
--- a/scripts/run-scss-lint.sh
+++ b/scripts/run-scss-lint.sh
@@ -6,4 +6,4 @@ PROJECT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd)
 
 docker run --rm \
   -v $PROJECT_DIR:/style-guide/ \
-  brainly/style-guide scss-lint $1
+  brainly/style-guide scss-lint "$@"
--- a/scripts/server.sh
+++ b/scripts/server.sh
@@ -18,7 +18,7 @@ echo "You can now open http://your_docker_ip:8181/dev/docs in web browser"
 
 docker run -t --rm \
   -p 8181:8000 \
-  --name brainly-style-guide \
+  --name brainly-style-guide-server \
   -v $PROJECT_DIR/dist:/style-guide/dist \
   brainly/style-guide http-server ./dist -p 8000
 
@@ -26,5 +26,5 @@ docker run -t --rm \
 # this will be executed when user hit CTRL+C
 echo
 echo "Please, bear with me while I am stopping and removing container..."
-docker stop brainly-style-guide
-docker rm brainly-style-guide
+docker stop brainly-style-guide-server
+docker rm brainly-style-guide-server
--- a/scripts/watch.sh
+++ b/scripts/watch.sh
@@ -5,8 +5,14 @@
 
 PROJECT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd)
 
-docker run -t --rm \
+docker run -t --rm --name brainly-style-guide-watch \
   -v $PROJECT_DIR/src:/style-guide/src \
   -v $PROJECT_DIR/dist:/style-guide/dist \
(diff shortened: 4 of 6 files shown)


scripts/server.sh
```diff
@@ -26,5 +26,5 @@ docker run -t --rm \
 # this will be executed when user hit CTRL+C
 echo
 echo "Please, bear with me while I am stopping and removing container..."
-docker stop brainly-style-guide
-docker rm brainly-style-guide
+docker stop brainly-style-guide-server
```
mr-mig: Do we have a non-server container?>

vergilius:
:ok_hand:

mr-mig:
Can you please add some desciption so that the reviewer do not read in the whole sources trying to figure out the context? 
It would be helpful from historical point of view.

scripts/ide-run-scss-lint-from-docker.sh
```diff
@@ -1,9 +1,3 @@
...
-
-# This script will run scss-lint
-
-PROJECT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd)
-
-docker run --rm \
-  -v $PROJECT_DIR:/style-guide \
-  brainly/style-guide bash -c "scss-lint /style-guide/$1 -f JSON | sed 's/\/style-guide\///g'"
+SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+source ${SCRIPT_DIR}/run-scss-lint.sh $1 -f JSON
```
aju: Remove code duplication use already existing script

scripts/run-scss-lint.sh
```diff
@@ -6,4 +6,4 @@ PROJECT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd)
 
 docker run --rm \
   -v $PROJECT_DIR:/style-guide/ \
-  brainly/style-guide scss-lint $1
+  brainly/style-guide scss-lint "$@"
```
aju: proxy any arguments to scss-lint

scripts/server.sh
```diff
@@ -18,7 +18,7 @@ echo "You can now open http://your_docker_ip:8181/dev/docs in web browser"
 
 docker run -t --rm \
   -p 8181:8000 \
-  --name brainly-style-guide \
+  --name brainly-style-guide-server \
```
aju: We need to distinct between watch and server

scripts/watch.sh
```diff
@@ -5,8 +5,14 @@
...
+docker run -t --rm --name brainly-style-guide-watch \
   -v $PROJECT_DIR/src:/style-guide/src \
   -v $PROJECT_DIR/dist:/style-guide/dist \
   -v $PROJECT_DIR/gulpfile.js:/style-guide/gulpfile.js \
   brainly/style-guide node_modules/.bin/gulp watch
+
+# this will be executed when user hit CTRL+C
+echo
+echo "Please, bear with me while I am stopping and removing container..."
+docker stop brainly-style-guide-watch
```
aju: Removing container after hitting CTRL+C

aju:
@mr-mig added

mr-mig:
I meant the description for PR, not the code bits :)

Like, what changed and why, like in this guideline: https://github.com/brainly/documents/blob/master/rfc/readme-best-practices.md

aju:
It was refactor so it should not change anything ;) But yes I should have added some detailed description.

mr-mig:
At least now we can easily run 2 containers in parallel :)

scripts/ide-run-scss-lint-from-docker.sh
marcinb: I can't understand why you split it, also naming doesn't make it easier to get as docker-level instructions are actually in `/run-scss-lint.sh` while this file abstracts that away and doesn't even mention word `docker` at all but is still called `ide-run-scss-lint-from-docker` ? :)

aju: This file is used as scss-lint exec file for WebStorm (and all others form the "family"). I can't pass in config of the plugin options to exec so I created proxy file to do it.  Maybe better name would be `ide-scss-lint-exec.sh` ?

mr-mig: @marcinb For me it is straightforward - the script runs scss lint from docker (a call to another script is an implementation detail).

mr-mig:
Update the description for PR and merge it.
:+1:

scripts/ide-run-scss-lint-from-docker.sh
marcinb: Ah, I see.

I like the name `ide-scss-lint-exec.sh`, in this case a small comment explaining that `This is an exec script for JetBrains-family IDEs` would be nice to have :+1:

marcinb: @mr-mig after reading @aju comment I understand the purpose, I don't really use any of these IDEs so it is hard to understand the purpose of this script it is coment-less

marcinb:
Except for the minor thing I mentioned above, :+1:

scripts/ide-run-scss-lint-from-docker.sh
aju: Updated name, added inline comment about it as well. (It is explained in documentation as well) :)

aju:
Thx for review! :) I fixed those things

PR #323, opened 2015-09-21
Refactored gutter mixin

mr-mig:
`$layoutDefaultPadding` -> `gutter(1)`


mr-mig:
THE QUESTION: do we really like `gutter(1 / 4) ` and `rhythm(1 / 2)`? 
Maybe `gutter(0.25)` and `rhythm(0.5)` is better?

aju:
`gutter(0.25)`
:+1:

kdzwinel:
I'm OK with both, lets just be consistent. 
If layoutDefaultPadding is not meant to be used directly anymore can we somehow make it 'private'?

mr-mig:
AFAIK there is no way to make it private

mr-mig:
I will update gutter and rhythm to decimals in a separate PR

kdzwinel:
If you put it inside the mixin it won't be global, not sure if that's the only place where it's being used though.

mr-mig:
If I put it in mixin - it won't be a part of the global config (and it should)

kdzwinel:
:+1:

PR #336, opened 2015-09-22
Dropdown rem to pixel

matzimowski:
there was an issue on some viewports

before:
screen shot 2015-09-22 at 12 25 30

after:
screen shot 2015-09-22 at 12 25 46

### DIFF
--- a/src/sass/_mixins.scss
+++ b/src/sass/_mixins.scss
@@ -120,6 +120,10 @@
   @return $size * $layoutDefaultPadding;
 }
 
+@function strip-units($number) {
+  @return $number / ($number * 0 + 1);
+}
+
 @function toPx($rems){
-  @return $rems / rhythm(1) * $baseFont;
+  @return strip-units($rems) * strip-units($baseFont) * 1px;
 }
--- a/src/sass/_mixins.scss
+++ b/src/sass/_mixins.scss
@@ -124,7 +124,8 @@
 // Remove the unit of a length
 // @param {Number} $number - Number to remove unit from
 // @return {Number} - Unitless number
-@function strip-unit($number) {
+@function stripUnit($number) {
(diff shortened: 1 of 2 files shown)


src/sass/_mixins.scss
```diff
@@ -119,3 +119,11 @@
 @function gutter($size){
   @return $size * $layoutDefaultPadding;
 }
+
+@function strip-units($number) {
+  @return $number / ($number * 0 + 1);
```
vergilius: `($number * 0 + 1) == 1` ?

matzimowski: https://github.com/sass/sass/issues/533

vergilius: :+1: add please comment in the code

kdzwinel: https://css-tricks.com/snippets/sass/strip-unit-function/ - don't you need these ifs there?

kdzwinel:
Why exactly are we doing this? Value calculated with rems wasn't precise enough? Don't we have some other places where this should be done?

matzimowski:
There is a problem connected with borders, border-size has to be an integer.
so this strange situation happens when border was 1.992px (but transformed to 2px) and left was like 1.992px and remained like this.
I found this issue only here so far.

src/sass/_mixins.scss
matzimowski: we don't care, we do not have to  generalise it yet :older_man:

kdzwinel:
:+1:

src/sass/_mixins.scss
matzimowski: C'mon, we will update the function according to ur comment.

aju:
:ok:

PR #341, opened 2015-09-22
Breaking changes for content-box for large components

matzimowski:
- Renamed mint-content-box__footer to mint-content-box__actions
  - Added --large for __title and __actions
  - Added examples

screen shot 2015-09-22 at 16 12 42

### DIFF
--- a/src/components/content-box/_content-box.scss
+++ b/src/components/content-box/_content-box.scss
@@ -8,6 +8,20 @@ $includeHtml: false !default;
       @include componentsContainer;
       min-height: rhythm(2);
 
+      &--large {
+        display: block;
+        margin-bottom: rhythm(1);
+
+        ~ .mint-content-box__actions--large {
+          margin-top: - rhythm(1);
+        }
+
+        ~ .mint-content-box__header {
+          margin-top: - rhythm(1.5);
+          margin-bottom: rhythm(0.5);
+        }
+      }
+
       & ~ & {
         margin-top: - rhythm(0.5);
         margin-bottom: - rhythm(0.5);
@@ -16,10 +30,6 @@ $includeHtml: false !default;
       ~ .mint-content-box__header {
         margin-top: - rhythm(0.5);
         margin-bottom: rhythm(0.5);
-      }
-
-      &--large ~ .mint-content-box__actions--large {
-        margin-top: - rhythm(1);
       }
     }
 
--- a/src/components/content-box/_content-box.scss
+++ b/src/components/content-box/_content-box.scss
@@ -17,6 +17,10 @@ $includeHtml: false !default;
         margin-top: - rhythm(0.5);
         margin-bottom: rhythm(0.5);
       }
+
+      &--large ~ .mint-content-box__actions--large {
+        margin-top: - rhythm(1);
+      }
     }
 
     &__header {
@@ -28,14 +32,18 @@ $includeHtml: false !default;
       word-wrap: break-word;
     }
 
-    &__footer {
+    &__actions {
       @include componentsContainer;
       min-height: rhythm(2);
+
+      &--large {
+        min-height: rhythm(3);
+      }
     }
 
     &--spaced {
       .mint-content-box__content,
-      .mint-content-box__footer {
+      .mint-content-box__actions {
         @include mintBreakpoint(medium-up) {
           padding-left: gutter(2);
         }
--- a/src/components/toplayer/_toplayer.scss
+++ b/src/components/toplayer/_toplayer.scss
@@ -89,7 +89,7 @@ $includeHtml: false !default;
       font-size: 32px;
       height: 20px;
       font-weight: bold;
-      padding: 0 gutter(1 / 4);
+      padding: 0 gutter(0.25);
       cursor: pointer;
 
       &:hover {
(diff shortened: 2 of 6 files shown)


src/components/toplayer/_toplayer.scss
```diff
@@ -89,7 +89,7 @@ $includeHtml: false !default;
       font-size: 32px;
       height: 20px;
       font-weight: bold;
-      padding: 0 gutter (1 / 4);
+      padding: 0 gutter(1 / 4);
```
aju: 0.25 ?:)

src/components/content-box/_content-box.scss
```diff
@@ -17,6 +17,10 @@ $includeHtml: false !default;
         margin-top: - rhythm(0.5);
         margin-bottom: rhythm(0.5);
       }
+
+      &--large ~ .mint-content-box__actions--large {
```
aju: I see we are getting more and more this cases is this a way to go?

aju:
:+1:

src/components/toplayer/_toplayer.scss
matzimowski: updated

src/components/content-box/_content-box.scss
mr-mig: Both yes and no.
Yes - as we need to squash some space for good looks. This space is caused by the fact that header-secondary falls out of rhythm if it is wrapped (which  is a real case for us).

No - because it's application should be limited to as few places as possible. 
This is the abstraction leak and creates implicit context. But we have to make this tradeoff.

aju: Shouldn't we make modifier then --no-spacing and use it instead ?

vergilius:
:+1:

matzimowski:
:point_right: :point_right:    :lollipop:

src/components/content-box/_content-box.scss
mr-mig: Not sure whether it can be a simple modifier. It depends on context.
But I like this idea more.

kdzwinel:
@matzimowski @mr-mig 

OK, I don't get this PR at all.

1. First, why  do we keep bloating style guide with examples that don't make sense? 
screen shot 2015-09-24 at 09 41 02
Where do we have something like this? And if we have why aren't these two separate components (header+button and text+actions)? They don't look like a single component.

2. Why do we have a `mint-content-box__title--large`? It doesn't make any sense to me. You should be able to put anything into `mint-content-box__title` and it should just work. Besides, it looks like this class does nothing because margin collapses:

screen shot 2015-09-24 at 09 39 29
screen shot 2015-09-24 at 09 39 38
screen shot 2015-09-24 at 09 43 55

Also, if you really need that margin, shouldn't it be called `--tall` as we [discussed](https://github.com/brainly/style-guide/blob/master/CONTRIBUTING.md#modifier-dictionary)?

mr-mig:
> Where do we have something like this? And if we have why aren't these two separate components (header+button and text+actions)? They don't look like a single component.

Main page. 

It may be a good idea to use 2 content-boxes instead of 1. We have just generalized this use case in this way.

> You should be able to put anything into mint-content-box__title and it should just work. 

It is impossible to make it work for both primary and *secondary header as secondary header break the rhythm when wrapped*. The main problem is the difference in padding you have for primary and secondary headers:
- primary need to have `rhythm(1)`
- secondary need to have `always < rhythm(1)` because of wrapping issue

mr-mig:
> Besides, it looks like this class does nothing because margin collapses:

It does nothing only for the primary header, as header has margin-bottom itself. It do have the margin for text-bit, though.

> Also, if you really need that margin, shouldn't it be called --tall as we discussed?

This is correct.

mr-mig:
> examples that don't make sense? 

I think examples should show how this thing is supposed to be used. Otherwise people will put all kind of shit inside `title`, and put titles in `content`.

We have done this ourselves already in several places as we did not have the semantic names for holes yet.

If you have a better suggestion how to deal with that, lets discuss it!

kdzwinel:
# Examples

> Main page.

If you are thinking about this:
screen shot 2015-09-25 at 17 34 39

Then sure, it's a nice usage of a content-box. This, on the other hand, is a mess:


> It does nothing only for the primary header, as header has margin-bottom itself. It do have the margin for text-bit, though.

So this example doesn't make sense because here this class does nothing. It also encourages wrong usage.

----------

My point is - lets keep the examples real. As you said:

> I think examples should show how this thing is supposed to be used.

I totally agree and I believe that good examples should:

- explain why a modifier/element was created and how we are using it,
- be useful snippets of code,
- look nice.

We should also limit ourselves to keep the number of examples low.

# VR

> It is impossible to make it work for both primary and secondary header as secondary header break the rhythm when wrapped

I guess I'll need more explanation here, ATM I feel like we are breaking and bending everything around just to make VR work.

> Otherwise people will put all kind of shit inside title, and put titles in content.

Agreed, and that's why I don't like https://github.com/brainly/symfony/pull/1913 . The usage of content-box there doesn't make any sense to me.

kdzwinel:
I created https://github.com/brainly/style-guide/issues/361 so that we remember about chaning `large` to `tall`.

PR #342, opened 2015-09-22
Change browser support to the last two versions + opera mini.

kdzwinel:
It was bothering me for a while - why we had ie8/9 support there? I'm not 100% about these new settings, just wanted to open a quick discussion.


matzimowski:
:+1:  :older_man:  == :older_woman:

aju:
I think I added it when we started this project and we were still thinking about supporting those browsers (for backwards compatibility)
:+1: for this change.

PR #344, opened 2015-09-23
Reseted default styles for tags

matzimowski:
- link decoration reseted
 - margin and padding of headings and paragraph

closes https://github.com/brainly/style-guide/issues/343


kdzwinel:
Lets move this changes to basics. normalize is a 3rd party library, we want to be able to update it in the future.

aju:
I agree with Konrad - we should not touch it (I know we did but there was no other way)

matzimowski:
If its a 3rd party, it should come from node_modules

matzimowski:
We will move it :crying_cat_face:

aju:
https://github.com/brainly/style-guide/issues/345

matzimowski:
@aju @kdzwinel done

aju:
:ok:

kdzwinel:
Ack

PR #351, opened 2015-09-24
Extract breadcrumbs to a separate component

kdzwinel:
Breadcrumbs are too much different from default list, I didn't make sense to bundle these two together.

screen shot 2015-09-24 at 10 22 22


aju:
:ok:

mr-mig:
Makes sense :+1:

PR #373, opened 2015-09-29
Refactor font-weight

### LINKED ISSUE #304: omit `$fontWeightBlack` variable usage
We can use `bolder` keyword for black font, instead of `500` value.
It will allow ditch out the `$fontWeightBlack`

aju:
closes #304 
As we can't use bolder all font-weight are now set using variables to be consistent and to easy refactor/change it.


vergilius:
:+1:

matzimowski:
:point_down: 
:point_down: 
:point_down: 



:+1:

vergilius:
have a :seat:

PR #376, opened 2015-09-29
Importing icons via <object>

### LINKED ISSUE #360: Avoid referencing icon file in each <use>
I came up with a technique that will allow us to refer to the icon file only once per page (instead of referring it in every icon):

https://dl.dropboxusercontent.com/u/8989748/svgtest/test.html

It also helps us avoid the svg4everybody polyfil.

It works on all browsers that we support.

Cons: icons won't show when JS is disabled.

kdzwinel:
Closes #360 

From now on, to make icons work you have to include them like so:

    <object type="image/svg+xml" data="images/icons.svg" height="0"></object>
    <object type="image/svg+xml" data="images/subjects-icons.svg" height="0"</object>

`height="0"` is there to make this element invisible but still load the icons. `display: none` will not trigger icon download.

and after that, you can use them in this convenient way:

    <svg class="mint-icon mint-icon--fluid mint-icon--gray-secondary">
        <use xlink:href="#icon-profile"></use>
    </svg>


gulpfile.js
```diff
@@ -27,6 +27,8 @@ var DOCS = path.join(SRC, 'docs');
 var SASS = path.join(SRC, 'sass');
 var COMPONENTS = path.join(SRC, 'components');
 
+var svgPolyfill = fs.readFileSync(path.join(SRC, 'svg-polyfill.js'), "utf8");
```
aju: I think we should move it inside the task not to block gulp start (so this would probably require joining those 2 tasks for icons/subject generation)

aju: I will do it as part of gulp cleanup

aju:
:ok: :fire: :fireworks: :+1:

matzimowski:
:+1: :lollipop:

PR #385, opened 2015-09-30
Fix issue in safari rating/badges/labels

### LINKED ISSUE #354: Badges broken on safari
screen shot 2015-09-24 at 13 50 01

### LINKED ISSUE #383: Broken small buttons in Safari
screen shot 2015-09-30 at 15 58 51

aju:
closes #383
closes #354
screen shot 2015-09-30 at 15 58 22
screen shot 2015-09-30 at 15 58 29
screen shot 2015-09-30 at 15 58 42
screen shot 2015-09-30 at 15 58 51

Small labels, small rating box and badge can't have min-height of vertical rhythm as they not fit buttons.

For now min-height is disabled for those small elements.

We can keep it that way and add some modifier to use those elements as standalone components (with margin compensation for it). We can make this min-height: 0 as some modifier and apply it when used in buttons etc. Maybe other ideas?

@mr-mig @kdzwinel @vergilius @matzimowski

### DIFF
--- a/src/components/badges/_badges.scss
+++ b/src/components/badges/_badges.scss
@@ -15,7 +15,7 @@ $includeHtml: false !default;
     @include fixText($badgeFontSize);
     @include uppercaseText;
     overflow: visible; // allow badges align as text
-    min-height: auto;
+    min-height: 0;
     height: auto;
 
     border-radius: $badgeFontSize;
--- a/src/components/labels/_labels.scss
+++ b/src/components/labels/_labels.scss
@@ -88,6 +88,8 @@ $includeHtml: false !default;
     }
 
     &--small {
+      min-height: 0;
+
       .mint-label__text,
       .mint-label__number {
         @include fixText($labelFontSizeSecondary);
--- a/src/components/rating/_rating.scss
+++ b/src/components/rating/_rating.scss
@@ -62,7 +62,7 @@ $includeHtml: false !default;
 
     &--small {
       height: rhythm(0.667);
-      min-height: rhythm(0.667);
+      min-height: 0;
 
       .mint-rate-box__star {
         height: $rateSmallIconSize;
@@ -71,7 +71,6 @@ $includeHtml: false !default;
 
       .mint-rate-box__counter {
         @include fixText($rateSmallFontSize, 0.125rem);
-        margin-left: gutter(0.25);
       }
     }
   }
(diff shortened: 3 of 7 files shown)


src/components/badges/_badges.scss
```diff
@@ -15,7 +15,7 @@ $includeHtml: false !default;
     @include fixText($badgeFontSize);
     @include uppercaseText;
     overflow: visible; // allow badges align as text
-    min-height: auto;
```
aju: auto is not correct value! :)

src/components/labels/_labels.scss
```diff
@@ -88,6 +88,8 @@ $includeHtml: false !default;
     }
 
     &--small {
+      min-height: 0;
```
aju: Small label does not have min heigh

src/components/rating/_rating.scss
```diff
@@ -71,7 +70,7 @@ $includeHtml: false !default;
       }
 
       .mint-rate-box__counter {
-        @include fixText($rateSmallFontSize);
+        @include fixText($rateSmallFontSize, 0.125rem);
```
aju: For better text alignment with stars on small elements.

src/components/badges/_badges.scss
mr-mig: According to MDN it is: https://developer.mozilla.org/en-US/docs/Web/CSS/min-height

But not supported by IE and Safari.
Seems fair for me to call it incorrect :)

src/components/rating/_rating.scss
```diff
@@ -62,8 +61,8 @@ $includeHtml: false !default;
     }
 
     &--small {
-      @include fixText($rateSmallFontSize);
```
mr-mig: Why is it removed? 
It should put the text in the middle of the available area in all browsers

mr-mig:
There should be no need for `min-height: rhythm(1)` for small components. They all should have min-height reseted.
The `auto` value is my mistake.

src/components/badges/_badges.scss
aju: `An earlier revision of the spec added the auto keyword and used it as the initial value. The CSSWG subsequently resolved to revert this change, however. As of March 29, 2013, the latest Editor's Draft doesn't modify the min-width property anymore (i.e. it no longer introduces the auto value).` so it will not get any support

src/components/rating/_rating.scss
aju: Its already added in line 74.

It was centring element in chrome but in a way that if button was 20px and rate-box was 24px it was sticking out 2px from top and 2px from bottom. On Safari it was sticking 4px from bottom - thats why button were broken.

aju:
So the way to go is `min-height: 0` for small elements?

aju:
I added `min-height: 0` for small elements

vergilius:
:+1:

matzimowski:
:+1:

PR #469, opened 2015-10-23
Feature/fix mint logo and home button

aju:
Those are fixes for mint-logo and mint-home-button that works in Opera Mini

### DIFF
--- a/src/components/header/header.html
+++ b/src/components/header/header.html
@@ -20,7 +20,10 @@
         <div class="mint-header">
             <div class="mint-header__container">
                 <div class="mint-header__left">
-                    <a href="#" class="mint-home-button"></a>
+                    <a href="#" class="mint-home-button">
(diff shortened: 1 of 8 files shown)


src/components/header/header.html
```diff
@@ -20,7 +20,10 @@ <h3 class="docs-block__header">Example usage</h3>
         <div class="mint-header">
             <div class="mint-header__container">
                 <div class="mint-header__left">
-                    <a href="#" class="mint-home-button"></a>
+                    <a href="#" class="mint-home-button">
```
aju: We (me and @kdzwinel ) have decided that for now this is the best solution. (even though it will download both images - as both images will be used on mobile only drawback will be desktop downloading mobile image)

mr-mig: :+1:

mr-mig:
:ship:

kdzwinel:
:ok:

PR #516, opened 2015-11-06
Added --spaced modifier for content box title/header

### LINKED ISSUE #330: Add --spaced modifier to mint-content-box__title and __header

mr-mig:
Fixed #330 

image


aju:
I think it has wrong name it should be something like `padded-left` or something like that - what do you think ?

mr-mig:
It is a wrong design. There should be a column from the left side, as in any media object.

Regarding the name - yes, its is kinda weird. Changing the name will be a braking change though. I'd make it a separate issue TBH

aju:
ok

matzimowski:
:+1:

PR #532, opened 2015-11-09
Added --full modifier for content-box__content

### LINKED ISSUE #428: Add 'mint-content-box__content--full'
```
.mint-content-box__content--full {
  width: 100%;
  height: 100%;
}
```

mr-mig:
Fixes #428 

I did not come up with a good example for that.
Any ideas?


aju:
OK!
I don't think it will be easy to do proper example - I used it in panels for messages and notifications to stretch to parent dimensions.(and that was needed to cover whole viewport)

Issue #322, opened 2015-09-21
mint-text with user content


mr-mig:
Remove special text modifiers like `mint-text--emphasised` and use font style tags (`<b>` `<strong>` `<em>` ...) inside `mint-text` component

vergilius:
We agreed, that this should be a part of application that uses style-guide. It's usefull, but not shouldn't be a part of style-guide (you can still create it with sg mixins and variables).

PR #573, opened 2015-11-12
Feature/update dropdowns to match selects

aju:
This is an idea, and I need your input on it

### DIFF
--- a/src/components/dropdowns/dropdowns.html
+++ b/src/components/dropdowns/dropdowns.html
@@ -160,17 +160,13 @@
                 </svg>
             </div>
             <div class="mint-dropdown__hole mint-dropdown__hole--inactive">
-                <div class="mint-label mint-label--emphasised mint-label--inactive mint-label--full-width">
-                    <div class="mint-label__text">
-                        Default label
-                    </div>
+                <div class="mint-text mint-text--gray-alt mint-text--obscure mint-text--xemphasised ">
+                    Default label
                 </div>
             </div>
             <div class="mint-dropdown__hole  mint-dropdown__hole--active">
-                <div class="mint-label mint-label--emphasised mint-label--full-width">
-                    <div class="mint-label__text">
-                        Choose option
-                    </div>
+                <div class="mint-text mint-text--obscure mint-text--xemphasised">
+                    Choose option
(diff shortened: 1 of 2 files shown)


vergilius:
AFAIK we decided to use labels there just because they look like labels ;D

aju:
References https://github.com/brainly/symfony/issues/2206
screen shot 2015-11-12 at 15 46 39
On the left current version on right updated

mr-mig:
+1 for Michal's comment

mr-mig:
If the styles are different - it may be a good idea to update a label actually :)

vergilius:
the problem is that we have texts that look the same way, but semantically are different => its bad 

on the other hand, I could agree with you if you build it on mixin or something similar

aju:
I talked with Janusz and we should use "black" font family only on text bits
So I created issue here: https://github.com/brainly/style-guide/issues/574
Solving it should make it work with labels

mr-mig:
The issue is not only semantics, but the behavior: you can center labels as if they are components. You cannot do this with a basic text.

And one more argument - `mint-text` should be a text container, not a text line. 
See https://github.com/brainly/style-guide/issues/322

mr-mig:
:+1: for solving "black color" issue

src/components/dropdowns/dropdowns.html
```diff
@@ -160,17 +160,13 @@ <h3 class="docs-block__header">Example default usage (interactive)</h3>
                 </svg>
             </div>
             <div class="mint-dropdown__hole mint-dropdown__hole--inactive">
-                <div class="mint-label mint-label--emphasised mint-label--inactive mint-label--full-width">
-                    <div class="mint-label__text">
-                        Default label
-                    </div>
+                <div class="mint-text mint-text--gray-alt mint-text--obscure mint-text--xemphasised ">
```
matzimowski: i would suggest using new practise here:
`<div style="display: none !improtant;">`
instead of 
`<div class="mint-text mint-text--gray-alt mint-text--obscure mint-text--xemphasised ">`

aju:
Ok I will fix usage of "black" font family and leave labels here

PR #615, opened 2015-12-10
Changing graySecondary, fixing select.

kdzwinel:
Mateusz Tatara asked us to change graySecondary. After change I had to update the "select" component and change color of one of the buttons. I've also sorted gray colors in the styleguide.

screen shot 2015-12-10 at 16 29 35

BEFORE:
screen shot 2015-12-10 at 15 21 20

AFTER:
screen shot 2015-12-10 at 15 21 55


aju:
OK - but I liked previous order of colors

matzimowski:
OK - but I liked previous order of colors

matzimowski:
and i think now text on select is less visible.

Issue #609, opened 2015-12-04
Add .mint-text--blue


aju:
```
.mint-text--blue {
  color: $bluePrimary;
}
```

aju:
We will not use blue text

PR #635, opened 2016-01-08
Add mint text blue option

aju:
closes https://github.com/brainly/style-guide/issues/609
screen shot 2016-01-08 at 09 55 25


vergilius:
:+1:

kdzwinel:
mmmm... this looks like a link, IMO we should not add that

aju:
@kdzwinel this was used here:
http://zadane.pl/pages/points

PR #637, opened 2016-01-11
Update textarea placeholder styles

aju:
screen shot 2016-01-11 at 14 05 35
Placeholder styles should match those of input text


vergilius:
:ok:

matzimowski:
:ok:

vergilius:
:+1:

src/components/form-elements/_textarea.scss
```diff
@@ -36,12 +36,9 @@ $includeHtml: false !default;
     }
 
     &::placeholder {
-      @include uppercaseText;
-      position: relative;
-      top: rhythm(0.125);
```
kdzwinel: why are we removing this?

aju: as we increased font size it was no longer needed

Issue #650, opened 2016-01-17
Convert fonts to WOFF2


kdzwinel:
<img width="485" alt="screen shot 2016-01-17 at 23 54 58" src="https://cloud.githubusercontent.com/assets/985504/12380554/c7285abe-bd75-11e5-9878-75548867da36.png">

Reading: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization?hl=en

kdzwinel:
http://caniuse.com/#feat=woff2 :hankey:

PR #652, opened 2016-01-18
Update fonts to use woff2

aju:
screen shot 2016-01-18 at 13 10 31


kdzwinel:
LGTM

closes #650

matzimowski:
clever, babe
:+1:

PR #654, opened 2016-01-21
Enable overriding breakpoints

aju:


vergilius:
:ok

kdzwinel:
Normally, we should not allow that, but I understand that this is some kind of a hack for WC? If so, let's add a task that removes that code.

aju:
Why we should not allow it? This allows customisation and I think its more a feature then something that we should remove after.

PR #680, opened 2016-03-04
Redirect from root folder to the latest styleguide version docs 

kdzwinel:
using simple, yet effective, html meta redirect

http://styleguide.brainly.com/ will finally redirect to the latest version of the doc automagically ✨

✴it won't work locally as local root (/dist/version/docs) is different from the production root (/version/docs)

### DIFF
--- a/scripts/tasks/root-redirect-page.js
+++ b/scripts/tasks/root-redirect-page.js
@@ -1,0 +1,13 @@
+module.exports = function (gulp, plugins, consts) {
+  return function () {
+    var fs = require('fs');
+
+    var outputPath = plugins.path.join(consts.DIST, 'index.html');
+    var redirectPageContent = fs.readFileSync(plugins.path.join(consts.SRC, 'root-redirect.html'), "utf8");
+
+    redirectPageContent = redirectPageContent.replace(/#LATEST_VERSION#/g, consts.VERSION);
+
+    fs.writeFileSync(outputPath, redirectPageContent);
+
+  };
+};
--- a/src/root-redirect.html
+++ b/src/root-redirect.html
@@ -1,0 +1,14 @@
+<!DOCTYPE HTML>
+<html lang="en-US">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="refresh" content="1;url=/#LATEST_VERSION#/docs/">
+  <script type="text/javascript">
+    window.location.href = "/#LATEST_VERSION#/docs/"
+  </script>
+  <title>Page Redirection</title>
+</head>
+<body>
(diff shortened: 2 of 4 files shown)


src/root-redirect.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE HTML>
+<html lang="en-US">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="refresh" content="1;url=/#LATEST_VERSION#/docs/">
+  <script type="text/javascript">
+    window.location.href = "/#LATEST_VERSION#/docs/"
```
kdzwinel: Just to be sure

src/root-redirect.html
```diff
@@ -0,0 +1,14 @@
...
+  </script>
+  <title>Page Redirection</title>
+</head>
+<body>
+If you are not redirected automatically, follow the <a href='/#LATEST_VERSION#/docs/'>link</a>
```
kdzwinel: Just to be super-sure

aju:
Did you change s3 setting manually?

kdzwinel:
@aju which setting? Pawel told me that it should work. It generates index.html that will be put in the root /. No need for any settings AFAIK.

kdzwinel:
@pmastalerz PTAL

@aju @vergilius please check, merge and deploy (togheter with @pmastalerz). I'm out next week.

scripts/tasks/root-redirect-page.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function (gulp, plugins, consts) {
+  return function () {
+    var fs = require('fs');
+
+    var outputPath = plugins.path.join(consts.DIST, 'index.html');
```
aju: this should be "basics.html" not "index.html"

kdzwinel: Why basics? This is a root redirect file that will redirect from / to /version/docs . IMO it should be called index.html

src/root-redirect.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE HTML>
+<html lang="en-US">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="refresh" content="1;url=/#LATEST_VERSION#/docs/">
```
kdzwinel: @aju maybe you thought about this one? That I should redirect to /version/docs/basics.html instead of /version/docs? /version/docs, for some reason, opens basics.html file by default

aju: We have set basics.html in S3 AS root document

kdzwinel: ah, now I get it. Maybe lets change it back to the default (index.html) and add index.html to /docs folder? We were talking about having some 'intro' for docs anyway.

PR #685, opened 2016-03-16
Introduce style guide intro/index page.

kdzwinel:
screen shot 2016-03-16 at 14 33 28


src/docs/_includes/head.html
```diff
@@ -0,0 +1,19 @@
+<head>
```
kdzwinel: Since now we have two layouts - one for component pages and one for index page - I moved the header to a separate template (because it's used by both layouts).

src/docs/_includes/navigation.html
```diff
@@ -1,9 +1,12 @@
 <nav class="main-menu">
+    <a href="./" class="main-menu__home">
+        <img src="images/logos/brainly-mobile.svg" />
+    </a>
```
kdzwinel: Logo at the beginning of the navigation is now clickable and points to the intro page.

src/docs/_includes/page.html
```diff
@@ -1,13 +0,0 @@
-<h1 class="main-header">Brainly style guide - {{ pageConfig.name }}</h1>
```
kdzwinel: this was moved to the `items` layout

src/docs/_layouts/article.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
kdzwinel: new layout for simple pages with text only (like the intro page)

src/docs/index.html
```diff
@@ -0,0 +1,37 @@
...
+<p class="mint-text">In order to use style guide components, you have to include CSS file in the <code>&lt;head&gt;</code> of your website:</p>
+
+<pre>
+  &lt;link href="http://styleguide.brainly.com/{{ site.data.package.version }}/style-guide.css" rel="stylesheet"/&gt;
+</pre>
+
+<p class="mint-text">In order to use basic set of <a href="./basics.html#icons" class="mint-link">icons</a> include this script before the closing &lt;/body&gt; tag:</p>
+
+<pre>
+  &lt;script src="<script>document.write(document.querySelector('script[src*=subjects-icons-]').src)</script>"&gt;&lt;/script&gt;
```
kdzwinel: I never thought I'll use `document.write`, but here we are. Why? Note that icon files and image files are fingerprinted (e.g. `subjects-icons-e9967e3fb1.js`) and I need to expose a vaild URL in the doc, so... I'm taking this URL from the `<head>` and putting it here - tada :fireworks:

src/docs/_includes/head.html
```diff
@@ -0,0 +1,19 @@
...
+
+  <meta name="theme-color" content="#6ED6A0">
+
+  <link rel="stylesheet" href="../style-guide.css">
+  <link rel="stylesheet" href="{{ "/css/main.css" | prepend: site.baseurl }}">
+
+  <script src="images/icons.js"></script>
+  <script src="images/subjects-icons.js"></script>
+
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.4/jquery.js"></script>
```
vergilius: why do we need jquery in header and at all?

vergilius:
looks fabulous :+1:

src/docs/_includes/head.html
matzimowski: it's used in our components. wanna do refactor and get ride of this?

matzimowski:
LOVE IT :+1:

aju:
OK

PR #695, opened 2016-03-25
Make all separator lines solid and make their color lighter

kdzwinel:
screen shot 2016-03-25 at 15 08 32
screen shot 2016-03-25 at 15 07 05


aju:
OK

matzimowski:
:+1:

Issue #449, opened 2015-10-19
Add option `--spaced` for `mint-avatar`


aju:
This should compensate `mint-avatar` to follow VR
```
.mint-avatar--spaced {
  margin: 0 4px 4px 0;
}
```

PR #711, opened 2016-04-14
Add avatar--spaced modifier

matzimowski:
closes https://github.com/brainly/style-guide/issues/449

🐔 
screen shot 2016-04-14 at 13 46 00


src/components/avatar/_avatar.scss
```diff
@@ -81,6 +81,10 @@ $includeHtml: false !default;
         top: 10px;
       }
     }
+
+    &--spaced {
+      margin: 0 4px 4px 0;
```
kdzwinel: variable

matzimowski: you are so variable

matzimowski: @kdzwinel up

aju:
let it  be!
:ok:

Issue #653, opened 2016-01-21
Add mint-link--dark


vergilius:
add black version of link

vergilius:
used in taskline, cc @JanuszSzewczyk

kdzwinel:
I don't see that being used in WC/Symfony

aju:
@kdzwinel `sg-link--dark` used in symfony task line:   `var TIMELINE_LINK_CLASS_NAME = 'sg-link sg-link--dark';`

burakukula:
I removed this in symfony today.

burakukula:
<img width="804" alt="screen shot 2016-12-16 at 16 35 52" src="https://cloud.githubusercontent.com/assets/4272331/21268130/c1622a4e-c3ad-11e6-8f98-f97a537122b1.png">

kdzwinel:
very good comment by me. Close it.

PR #729, opened 2016-04-27
Add dark modifier for links

matzimowski:
closes https://github.com/brainly/style-guide/issues/653

screen shot 2016-04-27 at 08 55 34


aju:
OK

kdzwinel:
OMG do we really want to have a link that doesn't look like a link?

PR #757, opened 2016-06-20
Add favicon 200x200 - facebook og:image

matzimowski:
connected with bug: https://trello.com/c/03rCIF54/262-lat-add-a-default-picture-to-show-when-www-brainly-lat-is-shared-on-social-media

currently used pic is too small, it has to be at least 200x200.
we have added extra transparent border - it should look even better when shared on fb.

we have tested it on our datacontest app.
screen shot 2016-06-20 at 10 25 52


kdzwinel:
good PS skills Mati.

👍

vergilius:
you're 🅰️ Ⓜ️ 🅰️ zing 👍

aju:
OK

PR #768, opened 2016-06-30
Add gray

matzimowski:
gray secondary light is used as a page background color

screen shot 2016-06-30 at 11 19 10
screen shot 2016-06-30 at 11 20 55


aju:
OK

PR #779, opened 2016-07-05
Layout update

matzimowski:


src/components/layout/_layout.scss
```diff
@@ -8,7 +8,7 @@ $layoutBorderRadius: 8px;
   &__container {
     display: flex;
     flex-direction: column;
-    align-items: flex-start;
+    align-items: stretch;
```
kdzwinel: that's an initial value Mati, you should know that, you are a grown man!

kdzwinel: and please reverse that push you just did

kdzwinel: yes, do it

kdzwinel:
good job!

PR #780, opened 2016-07-06
Add align-items: flex-start on layout container on large screen

matzimowski:
before
screen shot 2016-07-06 at 08 48 27

after
screen shot 2016-07-06 at 08 48 27


aju:
OK

PR #790, opened 2016-07-07
Text obscure update

aju:


src/components/text/_text.scss
```diff
@@ -13,7 +13,7 @@ $sgTextDarkSemitransparentColor: rgba(0, 0, 0, 0.4);
     }
 
     &--obscure {
-      @include typeVariant(obscure, 1);
+      @include typeVariant(obscure, 0.5);
```
kdzwinel: not 0.75?

aju: it 0.5 of rhythm so 0.75.rem so 12 px :D

vergilius:
OK

kdzwinel:
ok

PR #793, opened 2016-07-08
Make colors copy'able (?) and change the design a bit.

kdzwinel:
I had to simplify the design a bit to keep things nice & easy:

screen shot 2016-07-08 at 14 01 26


aju:
OK

vergilius:
looks better 👍

PR #816, opened 2016-07-22
Update sg-avatar to be display block

aju:
closes https://github.com/brainly/style-guide/issues/815
fixes issue with weird spacing around avatars


matzimowski:
OK

vergilius:
maybe it'll also fix issue with `perspective: 1px` 👍

PR #818, opened 2016-07-22
Add break words option for sg-text

aju:


matzimowski:
BREAKING WORD
OK

vergilius:
dont break given word 👍 give up

aju:
@matzimowski @vergilius this is not standard value but ill help in chrome/safari!

matzimowski:
@aju good enough

PR #824, opened 2016-08-30
Add set of new icons

matzimowski:
screen shot 2016-08-30 at 15 58 21


aju:
Ok

kdzwinel:
OK

src/sass/_config.scss
```diff
@@ -1,5 +1,4 @@
-/*
- DO NOT USE THESE COLORS DIRECTLY!
+/* DO NOT USE THESE COLORS DIRECTLY!
```
vergilius: this might be removed, as we do not follow this rule anymore

vergilius:
👍

src/sass/_config.scss
matzimowski: clever one!

aju:
Ok

PR #838, opened 2016-09-26
Update links

aju:
resolves https://github.com/brainly/style-guide/issues/831
screen shot 2016-09-26 at 14 21 32


matzimowski:
OK

PR #857, opened 2016-10-24
Remove useless padding from avatar icon

matzimowski:
before:
screen shot 2016-10-24 at 14 16 05

after:
screen shot 2016-10-24 at 14 16 10

### DIFF
--- a/src/components/avatar/_avatar.scss
+++ b/src/components/avatar/_avatar.scss
@@ -32,7 +32,6 @@ $includeHtml: false !default;
       width: 100%;
 
       &--icon {
-        padding-top: 3px;
         background-color: $avatarDefaultBackgroundColor;
       }
     }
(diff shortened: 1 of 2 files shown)


src/components/avatar/_avatar.scss
```diff
@@ -32,27 +32,18 @@ $includeHtml: false !default;
...
-        padding-top: 3px;
         background-color: $avatarDefaultBackgroundColor;
       }
     }
 
     &--small {
       width: $avatarSmallSize;
       height: $avatarSmallSize;
-
-      .sg-avatar__image--icon {
```
kdzwinel: do we use `sg-avatar__image--icon` for anything else? If not, we should get rid of it from the HTML IMO

kdzwinel: ah, we still use it for `background-color`, my bad.

matzimowski: yep, ack please :*

PR #876, opened 2016-12-08
Icon as button update - adding transparent version to action buttons.

burakukula:
icons-update

### DIFF
--- a/src/components/buttons/_icon-as-button.scss
+++ b/src/components/buttons/_icon-as-button.scss
@@ -151,6 +151,7 @@ $includeHtml: false !default;
       border-radius: 50%;
 
       &:hover {
+        fill: $iconAsButtonLightColor;
         background: $iconAsButtonTransparentColor;
       }
     }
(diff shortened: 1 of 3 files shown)


src/components/buttons/_icon-as-button.scss
```diff
@@ -151,6 +151,7 @@ $includeHtml: false !default;
       border-radius: 50%;
 
       &:hover {
+        fill: $iconAsButtonLightColor;
```
aju: why we need it?

aju:
@burakukula I would present it as a group (look at `Action` and `Action + square`) @kdzwinel WDYT ?

src/components/buttons/_icon-as-button.scss
burakukula: @aju `--light` overwrites this fill and we need it white on hover.

kdzwinel:
if we treat it as a color it should be everywhere, if we treat it as a type it should be in a separate group. TBH IDK how to think about this one

src/components/buttons/_icon-as-button.scss
aju: Then we need to do it like `&--action {` where blue is our main color

PR #881, opened 2016-12-12
Layout updates

kdzwinel:
## 1

ATM `document.body` has 💩 dimensions because of our header always being `fixed` and content containers having weird margins.

body-before

This PR fixes that.

## 2

ATM document size can be smaller than the viewport which looks 💩

before-large
before-medium
before-small

This PR fixes that also:

after-large
after-medium
after-small

Tested on Chrome/FF/Safari.

### DIFF
--- a/src/components/layout/_layout.scss
+++ b/src/components/layout/_layout.scss
@@ -5,9 +5,14 @@ $layoutContainerMarginTop: $headerHeight;
 $layoutGapWidth: gutter(5 / 6);
 
 .sg-layout {
+  min-height: 100vh;
+  display: flex;
+  flex-direction: column;
+
   &__container {
     display: flex;
     flex-direction: column;
+    flex-grow: 1;
     max-width: $layoutMaxWidth;
     margin: $layoutContainerMarginTop auto;
 
@@ -73,7 +78,6 @@ $layoutGapWidth: gutter(5 / 6);
 
   &__content {
     min-width: 0;
-    flex-grow: 1;
     flex-shrink: 0;
     padding: 0;
     width: 100%;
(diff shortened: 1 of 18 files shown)


src/components/layout/_layout.scss
```diff
@@ -5,9 +5,14 @@ $layoutContainerMarginTop: $headerHeight;
...
 
 .sg-layout {
+  min-height: 100vh;
+  display: flex;
+  flex-direction: column;
+
   &__container {
     display: flex;
     flex-direction: column;
+    flex-grow: 1;
```
kdzwinel: container, that keeps all page content, grows pushing footer to the bottom

src/components/layout/_layout.scss
```diff
@@ -73,14 +78,14 @@ $layoutGapWidth: gutter(5 / 6);
...
     min-width: 0;
-    flex-grow: 1;
     flex-shrink: 0;
     padding: 0;
     width: 100%;
 
     @include sgBreakpoint(large-only) {
       max-width: calc(100% - #{$layoutAsideWidth + $layoutGapWidth});
       flex-basis: 1px;
+      flex-grow: 1;
```
kdzwinel: main content should grow only on desktop, pushing "aside content" to the right. On mobile it shouldn't grow because it will create a gap between main content and "aside content" which, on mobile, is below

src/components/layout/layout_large.html
```diff
@@ -30,6 +32,14 @@
 
       </div>
     </div>
+
```
kdzwinel: I've added 'Footer' to our layout demos

PR #891, opened 2016-12-13
Reverse avatar colors.

burakukula:
Before
screen shot 2016-12-13 at 17 12 46

After
screen shot 2016-12-13 at 17 12 17

### DIFF
--- a/src/components/avatar/avatar.html
+++ b/src/components/avatar/avatar.html
@@ -5,42 +5,42 @@
   <div class="docs-block__content docs-block__content--to-bottom">
     <div class="sg-avatar sg-avatar--small">
       <div class="sg-avatar__image sg-avatar__image--icon">
-        <svg class="sg-icon sg-icon--x24 sg-icon--gray-secondary">
+        <svg class="sg-icon sg-icon--x24 sg-icon--gray">
           <use xlink:href="#icon-profile"></use>
         </svg>
       </div>
     </div>
     <div class="sg-avatar">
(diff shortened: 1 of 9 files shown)


src/components/avatar/avatar.html
```diff
@@ -55,42 +55,42 @@ <h3 class="docs-block__header">Default avatars with border!</h3>
...
-          <svg class="sg-icon sg-icon--x94 sg-icon--gray-secondary">
+          <svg class="sg-icon sg-icon--x94 sg-icon--gray">
             <use xlink:href="#icon-profile"></use>
           </svg>
         </div>
       </div>
       <div class="sg-avatar sg-avatar--xxxlarge sg-avatar--with-border">
         <div class="sg-avatar__image sg-avatar__image--icon">
-          <svg class="sg-icon sg-icon--x118 sg-icon--gray-secondary">
+          <svg class="sg-icon sg-icon--x118 sg-icon--gray">
```
kdzwinel: I hope you haven't cheated using `cmd+r` and changed all of these by hand !!

aju:
@burakukula did you agree with Greg that we can change it everywhere like that?

burakukula:
@aju yes, Greg did agree to this change. and @kdzwinel yes, I did it by hand. I was pure pleasure.

src/components/avatar/avatar.html
aju: There is more places that needs this change! :)

burakukula: @aju please have a look

Issue #913, opened 2017-01-10
popup-menu - min-height instead of height


matzimowski:
```
.sg-popup-menu {
  height: auto;
  min-height: 2.75rem;
}
```

PR #914, opened 2017-01-11
Update popup-menu

matzimowski:
no visual changes. it allows to put inside some larger elements

closes https://github.com/brainly/style-guide/issues/913


src/components/popup-menu/popup-menu.scss
```diff
@@ -31,6 +31,7 @@ $includeHtml: false !default;
       right: gutter(1);
       width: 0;
       margin: 0 auto;
+      pointer-events: none;
```
matzimowski: it solves issue with empty clicks on i.e. `hamburger` button

PR #915, opened 2017-01-17
Colors update

burakukula:
- Update of existing colors
 - Adding new special colors to the palet.

Before:

screen shot 2017-01-17 at 15 46 22
screen shot 2017-01-17 at 15 46 32
screen shot 2017-01-17 at 15 46 39

After:

screen shot 2017-01-17 at 15 47 09
screen shot 2017-01-17 at 15 47 16
screen shot 2017-01-17 at 15 47 22

### DIFF
--- a/src/sass/_config.scss
+++ b/src/sass/_config.scss
@@ -1,29 +1,38 @@
 // Primary colors
-$black: #000000;
-$mintPrimary: #6ed6a0;
-$mintSecondary: #a7e6c6;
-$mintSecondaryLight: #e1f6eb;
-$bluePrimary: #5bb8ff;
-$blueSecondary: #abe3ff;
-$blueSecondaryLight: #ddf3ff;
+$black: #020a1b;
+$mintPrimary: #56d9ab;
+$mintSecondary: #aff3cf;
+$mintSecondaryLight: #d5fae4;
+$bluePrimary: #49a2ff;
+$blueSecondary: #abe1ff;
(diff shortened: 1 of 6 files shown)


kdzwinel:
IMO we should not expose these colors to users. People should not use them, they are created for a specific usecase. I'm not even sure if we really want tournament/challenges icons in the SG.

burakukula:
@kdzwinel I just leave them in config so we can use them in games. They are not visible in styleguide.

kdzwinel:
Keeping only colors as variables should be :+1:

burakukula:
@kdzwinel approved then? :)

src/sass/_config.scss
```diff
@@ -1,29 +1,37 @@
...
 // Grey colors
 $white: #ffffff;
-$grayPrimary: #70747d;
-$graySecondary: #a1a6b1;
-$grayLight: #dee2e5;
-$graySecondaryLight: #f0f3f5;
+$grayPrimary: #434e66;
+$graySecondary: #9fa6b5;
+$grayLight: #e1e8ed;
+$graySecondaryLightest: #f0f3f5;
```
kdzwinel: graySecondaryLightest? why not graySecondaryLight?

burakukula: @kdzwinel this is the first step to change the namespace:

grayLight -> graySecondaryLight
graySecondaryLight -> graySecondaryLightest

PR #917, opened 2017-01-18
Update buttons

matzimowski:
screen shot 2017-01-20 at 16 04 01
screen shot 2017-01-20 at 16 04 07

### DIFF
--- a/src/components/buttons/_buttons-mixins.scss
+++ b/src/components/buttons/_buttons-mixins.scss
@@ -90,7 +90,7 @@ $buttonDisabledFontColor: $graySecondaryLight;
   }
 
   &--disabled {
-    opacity: 0.7;
+    opacity: 0.45;
     cursor: default;
     box-shadow: none;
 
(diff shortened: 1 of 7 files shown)


src/components/labels/_labels.scss
```diff
@@ -108,6 +101,13 @@ $includeHtml: false !default;
       }
 
     }
+
+    &--unstyled {
```
aju: didn't we call it usually differently?

aju: ah its moved ok

matzimowski: i just put it a lit lower, because it worked only with sg-label--small and didn't work with sg-label--secondary

src/components/buttons/_buttons-mixins.scss
```diff
@@ -1,38 +1,100 @@
...
+    color: $buttonDisabledFontColor;
+    fill: $buttonDisabledFontColor;
+    cursor: default;
+
+    &:focus {
+      box-shadow: none;
+    }
+
+    &:active {
+      transform: scale(1);
```
kdzwinel: `none`

PR #921, opened 2017-01-23
Fix for too long text in buttons

matzimowski:
before:

screen shot 2017-01-23 at 16 22 50
screen shot 2017-01-23 at 16 23 40

after:

screen shot 2017-01-23 at 16 22 40
screen shot 2017-01-23 at 16 24 06


aju:
Ellipsis?

matzimowski:
@aju button use flexbox (inline-flex) with centering, it won't work

PR #925, opened 2017-01-27
Action list vol.1

burakukula:
Updated again:

screen shot 2017-01-31 at 17 36 49

### DIFF
--- a/src/components/action-list/action-list.html
+++ b/src/components/action-list/action-list.html
@@ -1,26 +1,10 @@
 <section class="docs-block">
   <aside class="docs-block__info">
-    <h3 class="docs-block__header">Action list simple</h3>
-
-  </aside>
-  <div class="docs-block__content">
-    <div class="sg-actions-list sg-actions-list--space-between">
-      <div class="sg-actions-list__hole">
-        <button class="sg-button-secondary sg-button-secondary--small sg-button-secondary--inverse">
-          Answer
-        </button>
-      </div>
-    </div>
-  </div>
-</section>
-
-<section class="docs-block">
-  <aside class="docs-block__info">
     <h3 class="docs-block__header">Action list space between</h3>
 
   </aside>
   <div class="docs-block__content">
-    <div class="sg-actions-list sg-actions-list--no-wrap sg-actions-list--space-between-alt">
+    <div class="sg-actions-list sg-actions-list--no-wrap sg-actions-list--space-between">
 
       <div class="sg-actions-list__hole sg-actions-list__hole--no-shrink">
 
@@ -70,7 +54,30 @@
 
 <section class="docs-block">
   <aside class="docs-block__info">
-    <h3 class="docs-block__header">Action list navigation</h3>
+    <h3 class="docs-block__header">Action list to right</h3>
+  </aside>
+
+  <div class="docs-block__content">
+    <div class="docs-block__contrast-box docs-block__contrast-box--full-width">
+      <div class="sg-actions-list sg-actions-list--to-right">
+        <div class="sg-actions-list__hole">
+          <button class="sg-button-secondary sg-button-secondary--alt sg-button-secondary--small">
+            accept
+          </button>
+        </div>
+        <div class="sg-actions-list__hole">
+          <button class="sg-button-secondary sg-button-secondary--alt-inverse sg-button-secondary--small">
+            later
+          </button>
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+
+<section class="docs-block">
+  <aside class="docs-block__info">
+    <h3 class="docs-block__header">Action list no spacing</h3>
   </aside>
 
   <div class="docs-block__content">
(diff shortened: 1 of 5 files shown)


src/components/action-list/action-list.html
```diff
@@ -0,0 +1,160 @@
...
+
+<section class="docs-block">
+  <aside class="docs-block__info">
+    <h3 class="docs-block__header">Action list navigation</h3>
+  </aside>
+
+  <div class="docs-block__content">
+    <div class="docs-block__contrast-box">
+      <div class="sg-actions-list">
+        <div class="sg-actions-list__hole sg-actions-list__hole--no-spacing sg-actions-list__hole--container">
```
kdzwinel: so we have `space-between` `space-between-alt` and `no-spacing`? So what's the default?

src/components/action-list/action-list.html
```diff
@@ -0,0 +1,160 @@
...
+  </div>
+</section>
+
+<section class="docs-block">
+  <aside class="docs-block__info">
+    <h3 class="docs-block__header">Action list space between</h3>
+
+  </aside>
+  <div class="docs-block__content">
+    <div class="sg-actions-list sg-actions-list--no-wrap sg-actions-list--space-between-alt">
```
kdzwinel: `space-between-alt` doesn't say anything about it - what's different between this and the regular `space-between`?

src/components/action-list/action-list.html
kdzwinel: `container` doesn't reveal much - name should probably suggest how it will behave IMO

src/components/action-list/action-list.html
```diff
@@ -0,0 +1,160 @@
+<section class="docs-block">
+  <aside class="docs-block__info">
+    <h3 class="docs-block__header">Action list simple</h3>
+
+  </aside>
+  <div class="docs-block__content">
+    <div class="sg-actions-list sg-actions-list--space-between">
```
aju: Remove `sg-actions-list--space-between` as simple should be plain form

src/components/action-list/action-list.html
```diff
@@ -0,0 +1,160 @@
...
+
+<section class="docs-block">
+  <aside class="docs-block__info">
+    <h3 class="docs-block__header">Action centered full width elements</h3>
+  </aside>
+
+  <div class="docs-block__content">
+    <div class="docs-block__contrast-box">
+      <div class="sg-actions-list">
+        <div class="sg-actions-list__hole--grow sg-actions-list__hole--full-width-centered">
```
kdzwinel: since there is `grow`, why do we need `full-width`? None of these classes say that this type forces column layout and row is the default.

src/components/action-list/action-list.scss
```diff
@@ -0,0 +1,87 @@
...
+
+    &--no-shrink {
+      flex-shrink: 0;
+    }
+
+    &--grow {
+      flex-grow: 1;
+    }
+
+    &--full-width-centered {
```
aju: this is not needed

src/components/action-list/action-list.scss
```diff
@@ -0,0 +1,87 @@
...
+    > .sg-actions-list__hole {
+      margin-right: 0;
+
+      &:last-child {
+        margin-left: auto;
+      }
+    }
+  }
+
+  &--space-between-alt {
```
aju: Clean this up

burakukula:
@aju @kdzwinel please have a look now. I updated examples. As you know, refactor is done in WC and SF.

src/components/action-list/action-list.html
```diff
@@ -0,0 +1,166 @@
...
+          </button>
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+
+<section class="docs-block">
+  <aside class="docs-block__info">
+    <h3 class="docs-block__header">Action list no spacing</h3>
```
aju: I would call it just action list as its default behaviour and put it first

src/components/action-list/action-list.html
```diff
@@ -0,0 +1,166 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</section>
+<section class="docs-block">
+  <aside class="docs-block__info">
+    <h3 class="docs-block__header">Action list no wrap and flex start</h3>
```
aju: ` and flex start` is not needed

src/components/action-list/action-list.html
```diff
@@ -0,0 +1,166 @@
...
+  </aside>
+
+  <div class="docs-block__content">
+    <div class="docs-block__contrast-box" style="width: 300px;">
+      <div class="sg-actions-list sg-actions-list--no-wrap sg-actions-list--to-top">
```
aju: remove `--to-top` or put it in next example so ppl can see difference

src/components/action-list/action-list.scss
```diff
@@ -0,0 +1,65 @@
...
+    > .sg-actions-list__hole {
+      margin: 0 0 0 gutter(0.5);
+    }
+  }
+
+  &--space-between {
+    justify-content: space-between;
+
+    > .sg-actions-list__hole {
+      margin-right: 0;
```
aju: I think we a re missing here 
```
&:first-child {
    margin-left: 0;
}
```

burakukula: @aju this was added to --space-between, but I think we don't use it.

burakukula:
@aju please have a look.

aju:
@burakukula the

PR #929, opened 2017-02-13
Update question icon

matzimowski:
before:
screen shot 2017-02-13 at 19 09 31

after:
screen shot 2017-02-13 at 19 09 35


vergilius:
was it optimized? looks huge compared to old one

matzimowski:
@vergilius you are right, i will try to optimise it first

kdzwinel:
Yep, looking at both icons the new one shouldn't be much more complicated than the previous one.

matzimowski:
@vergilius @kdzwinel optimized using https://jakearchibald.github.io/svgomg/

PR #932, opened 2017-02-15
Disabled thank you button

burakukula:
Propo:

screen shot 2017-02-15 at 16 06 14


aju:
maybe it should be in file `_buttons-mixins.scss` ?

burakukula:
@aju maybe like this. I moved styles from mixin disabled modifier to this new class. WDYT?

Issue #936, opened 2017-03-24
.sg-icon-as-button--action-active.sg-icon-as-button--light


matzimowski:
```
.sg-icon-as-button--action-active.sg-icon-as-button--light {
  background-color: $white;
}
```

PR #937, opened 2017-03-24
Add xxsmall and light modifier to icon-as-button

matzimowski:
closes https://github.com/brainly/style-guide/issues/935
closes https://github.com/brainly/style-guide/issues/936

screen shot 2017-03-24 at 16 10 26

screen shot 2017-03-24 at 16 22 23

### DIFF
--- a/src/components/buttons/_icon-as-button.scss
+++ b/src/components/buttons/_icon-as-button.scss
@@ -66,6 +66,13 @@ $includeHtml: false !default;
       }
     }
 
+    &--xxsmall {
+      .sg-icon-as-button__hole {
+        width: 16px;
+        height: 16px;
+      }
+    }
+
     &--with-border {
       .sg-icon-as-button__hole {
         border: solid $iconAsButtonColor 1px;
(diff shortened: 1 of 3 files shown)


src/components/buttons/_icon-as-button.scss
```diff
@@ -66,6 +66,13 @@ $includeHtml: false !default;
...
 
+    &--xxsmall {
+      .sg-icon-as-button__hole {
+        width: 16px;
+        height: 16px;
+      }
+    }
+
     &--with-border {
       .sg-icon-as-button__hole {
```
aju: do we still have this hole?

src/components/buttons/_icon-as-button.scss
```diff
@@ -66,6 +66,13 @@ $includeHtml: false !default;
       }
     }
 
+    &--xxsmall {
+      .sg-icon-as-button__hole {
+        width: 16px;
```
aju: I know we have it like that above but it should have been gutter

PR #956, opened 2017-05-10
New set of subjects icons

burakukula:
The '3 dots' icons are used just to let market's know that they need to stop using icons that were in that places until now.
icons-sg


aju:
Did you check whats the size difference (old vs new) ?

vergilius:
I kinda dislike defs (whitespaces should be removed or indented properly)

burakukula:
@aju gzip version of a new set of icons will have 14, the cuurent set have 19.7

aju:
@burakukula <3

PR #985, opened 2017-06-09
Color pallet update

burakukula:
screen shot 2017-06-09 at 16 51 11


src/docs/_data/colors.yml
```diff
@@ -1,16 +1,16 @@
 primary:
   - name: Black
     variable: $black
-    hex: "020a1a"
+    hex: "020a1b"
```
kdzwinel: black is the new black

PR #989, opened 2017-06-12
Selects update

burakukula:
screen shot 2017-06-12 at 16 36 00
screen shot 2017-06-12 at 16 35 39

### DIFF
--- a/src/components/dropdowns/dropdowns.html
+++ b/src/components/dropdowns/dropdowns.html
@@ -5,9 +5,9 @@
     <div class="docs-block__content">
         <div class="sg-dropdown">
             <div class="sg-dropdown__icon">
-                <svg class="sg-icon sg-icon--x8 sg-icon--gray">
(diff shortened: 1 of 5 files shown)


src/components/dropdowns/dropdowns.html
```diff
@@ -5,9 +5,9 @@ <h3 class="docs-block__header">Closed</h3>
     <div class="docs-block__content">
         <div class="sg-dropdown">
             <div class="sg-dropdown__icon">
-                <svg class="sg-icon sg-icon--x8 sg-icon--gray">
-                    <use xlink:href="#icon-arrow_down"></use>
-                </svg>
+              <svg class="sg-icon sg-icon--x8 sg-icon--gray">
+                <path fill="#9fa6b5" d="M0 1l4 6 4-6"/>
```
aju: I think we need to add it to style guide :) we can't copy it everywhere

PR #1001, opened 2017-06-20
Fix styles for header

burakukula:
Remove hardcoded height for topbar.


burakukula:
@aju I don't understand it was hardcoded in the first place?

PR #1006, opened 2017-06-20
Remove/rename classes that no longer exist.

kdzwinel:


src/components/header/_header.scss
```diff
@@ -28,14 +36,6 @@ $includeHtml: false !default;
       }
     }
 
-    &--fixed {
-      position: fixed;
-      top: 0;
-      left: 0;
-      right: 0;
-      z-index: $sgHeaderZIndex;
-    }
```
kdzwinel: IMO modifiers should be before elements, but linter doesn't seem to enforce that

vergilius: lets add it ;) @aju

Issue #1018, opened 2017-06-27
icon-as-button missing styles


matikrk:
Not all styles are defined in scss. or It's done baddly
Cases:
* Border + adaptive = bad border color
* action + adaptive + active = background is primary color not adaptive
* transparent + active + colors(adaptive, gray, gray-secondary, gray-light, dark ) = not defined
* icon-as-button have less amount of colors than icons

kdzwinel:
@matikrk you were rebuilding it in react, any of these issues were fixed?

matikrk:
there was issues on styles side. I only created all combinations in React and saw problems. Nothing was fixed

PR #1019, opened 2017-06-27
Icon as button react

matikrk:
* component
* examples
* tests
* fix icon spec


src/components/icon-as-button/IconAsButton.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import Icon, {types, colors as iconColors} from '../icons/Icon';
+
+const colors = {
```
aju: Can we have some const with all of it (external to components) ?
@vergilius WDYT ? we just need to verify how many components uses all available possible colors

matikrk: icon-as-button have less amount of colors than normal icons. And it has not perfectly  written scss. I created issue on github https://github.com/brainly/style-guide/issues/1018

PR #1024, opened 2017-06-29
Avatar component

matikrk:
* create component + tests + examples
* DocsBlock.jsx add toBottom property


src/components/avatar/avatar.jsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import Icon, {types as iconTypes, colors as iconColors} from '../icons/Icon';
```
aju: can we try to keep global imports (no ..)

PR #1028, opened 2017-07-03
Media react component

matikrk:
* rename avatar component from avatar.jsx to Avatar.jsx
* fix avatar ico size
* media component examples
* delete unused properties of media component
* create media component
* tests


src/components/avatar/Avatar.jsx
```diff
@@ -16,7 +16,7 @@ const sizes = {
 const iconSizes = {
   xsmall: 14,
```
vergilius: I still think, that these values should be camelized

matikrk: for now it's same as style names
maybe later we standardise it in all components, same plan is for colors

src/components/media/_media.scss
```diff
@@ -60,11 +60,6 @@ $mediaItemSecondaryColor: $grayPrimary;
     min-width: 0;
     position: relative;
     flex: 1 1 0;
-
-    &--centered {
-      display: flex;
-      align-items: center;
-    }
```
aju: not used?

matikrk: not used
https://github.com/search?utf8=%E2%9C%93&q=org%3Abrainly+sg-media__wrapper--centered&type=Code
and also I tried search it manually

PR #1070, opened 2017-07-04
Add actionList react component

### LINKED ISSUE #1035: JSX: action-list

matzimowski:
Closes #1035


src/components/action-list/ActionList.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+const DIRECTION = {
+  toRight: 'to-right',
```
aju: TO_RIGHT

src/components/action-list/ActionList.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const ActionList = ({holes = [], direction, noWrap}) => {
+  const actionListClass = classNames('sg-actions-list', {
+    [`sg-actions-list--${direction}`]: direction,
+    'sg-actions-list--no-wrap': noWrap
+  });
+
+  return (
+    <div className={actionListClass}>
+      {holes.map((hole, i) => <div key={i} className="sg-actions-list__hole">{hole}</div>)}
```
aju: I don't think ita a good idea it should be separate component

src/docs/ContrastBox.jsx
```diff
@@ -3,15 +3,15 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
 const contrastBlockCssClass = 'docs-block__contrast-box';
-const ContrastBox = ({toBottom, smallPadding, light, fullWidth, children}) => {
+const ContrastBox = ({toBottom, smallPadding, light, fullWidth, children, style}) => {
```
aju: style looks like a hack, can we fix it as well?

src/components/action-list/ActionList.jsx
matikrk: One simple element IMHO it's not bad thing

aju: Holes should be separate. Why?
When we will add modifiers for holes with this setup it will be impossible to modify them

matzimowski: there are no modifiers for now.
i would stick to current solution and when we decide to add modifiers we can create separated component

aju: Mati its same component and it will be hard to split @vergilius WDYT?

matzimowski: OK, i will create separated component

src/docs/ContrastBox.jsx
matzimowski: it's used only in docs, i would stick to this solution, it allows us to adjust this ContrastBox if needed.

aju: Nooo lets not produce garbage and it looks like garbage

matzimowski: i think it's not garbage , what is your proposition? adding `width` prop here has no point imho

matzimowski: @aju i got rid of it

src/components/action-list/pages/action-list.jsx
```diff
@@ -0,0 +1,119 @@
...
+        </div>,
+        <ButtonSecondary type={buttonSecondaryTypes.alt_inverse} small={true}>
+          start
+        </ButtonSecondary>
+      ]}>
+      </ActionList>
+    </DocsBlock>
+
+    <DocsBlock info='No wrap' multiContent={[
+      <div style={{width: '200px'}}>
```
matzimowski: @aju i did it like this, wdyt?

matzimowski: it looks the same

matikrk: for me cool

aju: add it as some option for DockBlock like narrow or something

aju: and proper modifier to scss

matzimowski:
@aju updated as you suggested

src/components/action-list/ActionList.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+const ActionList = ({holes = [], direction, noWrap}) => {
+  const actionListClass = classNames('sg-actions-list', {
+    [`sg-actions-list--${direction}`]: direction,
+    'sg-actions-list--no-wrap': noWrap
+  });
+
+  return (
+    <div className={actionListClass}>
+      {holes.map((hole, i) => <ActionListHole key={i}>{hole}</ActionListHole>)}
```
aju: Maaati this does not change a thing
just put children here and push this list from outside
Or check if childrea are type of <ActionListHole> if not the map like above

PR #1075, opened 2017-07-06
Rating star react

### LINKED ISSUE #1055: JSX: rating

matikrk:
Closes #1055


src/components/rating/Rating.jsx
```diff
@@ -0,0 +1,66 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import Star from './Star';
+
+const ICO_SIZE = {
+  SMALL: 14,
+  NORMAL: 16
+};
+
+const RateCounter = ({counter}) => {
```
aju: seprate file please

src/components/rating/Rating.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+RateCounter.propTypes = {
+  counter: PropTypes.number
+};
+
+class Rating extends Component {
+  onClick(index) {
+    const {onChange = () => undefined, rate, active} = this.props;
+
+    if (!active) {
```
aju: this.props.active ? I think its more readable ? @vergilius

src/components/rating/Star.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import Icon, {types, colors} from '../icons/Icon';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+
+const Star = ({size, checked, onClick = () => undefined}) => {
```
aju: I don't like this empty handlers, I don't think is efficient to add such handlers to DOM elements

matikrk: If you just showing rating you don't need specify function. IMHO having default dummy function is better than making checking if there is function inside

aju: @vergilius @kdzwinel WDYT?

matikrk: I can create kind of helper with return empty function and we will use just reference to this.

src/components/rating/Rating.jsx
matikrk: One time destruction shows you all needed properties.

vergilius: I prefer keeping as it is

src/components/rating/Star.jsx
vergilius: I prefer checking if callback is passed instead of having dummy one by default

src/components/rating/RateCounter.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const RateCounter = ({counter}) => {
+  if (counter === undefined) {
```
vergilius: `undefined` can be overridden, afaik its anti pattern

vergilius: oh, wrong, doest't look like it's possible

vergilius: at least in chrome

aju: this is actually OK right now

src/components/rating/Star.jsx
aju: 👍

src/components/rating/Rating.jsx
aju: OK

src/components/rating/Star.jsx
matikrk: This one looks better for you? 
```
let handleClick;
if(onClick){
handleClick=onClick;
}else{
handleClick=()=>undefined
}
```

aju: no no no :) you misses my point if there is no onClick function we don't add it as property

matikrk: ok :) 
But in Star we need it. I could delete dump function and make it required and change tests.
Star is integral part of Rating

PR #1076, opened 2017-07-06
List - react

### LINKED ISSUE #1049: JSX: list

anna-wro:
screen shot 2017-07-06 at 13 57 27
screen shot 2017-07-06 at 13 57 34

Closes #1049


src/components/list/List.jsx
```diff
@@ -0,0 +1,24 @@
...
+import classNames from 'classnames';
+import ListItem from '../list/ListItem';
+
+const List = ({items, spaced, small}) => {
+  const listClass = classNames('sg-list', {
+    'sg-list--spaced-elements': spaced
+  });
+
+  return <ul className={listClass}>
+    {items.map(({text}, index) =>
```
matikrk: as I understand do we pass in items array of objects with 1 property text?
`items=[{text:'abc'},{text:'xyz'}]`
If yes, we should pass only directly text property, not opaqued in object

src/components/list/List.jsx
```diff
@@ -0,0 +1,24 @@
...
+      <ListItem key={index} text={text} small={small}/>
+    )}
+  </ul>;
+};
+
+List.propTypes = {
+  spaced: PropTypes.bool,
+  small: PropTypes.bool,
+  items: PropTypes.array.isRequired
```
matikrk: `PropTypes.arrayOf(PropTypes.node)`
smth like this will be better

PR #1078, opened 2017-07-06
Menu list - react

anna-wro:
screen shot 2017-07-06 at 15 26 23


src/components/list/MenuList.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+const SIZE = {
+  SMALL: 'small',
+  NORMAL: 'normal',
+  LARGE: 'large'
+};
+
+const MenuList = ({items, size}) => {
+  const listClass = classNames('sg-menu-list', {
+    [`sg-menu-list--${size}`]: size !== SIZE.NORMAL
```
aju: what happens when size is undefined ?

anna-wro: nothing breaks, the list has a default size (there are only additional classes for small and large size)

aju: doesn't it have --undefined class ?:)

anna-wro: Whoops! I've just fixed it :)

aju:
@anna-wro please bum version :)

anna-wro:
@aju I did it :)

PR #1081, opened 2017-07-07
Flash messages react

anna-wro:
screen shot 2017-07-07 at 09 53 21

Fixed issue #1039


src/components/flash-messages/FlashMessage.jsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+const FlashMessage = ({text, type = TYPE.DEFAULT}) => {
+  const messageClass = classNames('sg-flash__message', {
+    [`sg-flash__message--${type}`]: type !== TYPE.DEFAULT
+  });
+
+  return <div className="sg-flash">
+    <div className={messageClass}>
+      <div className="sg-text sg-text--emphasised sg-text--small sg-text--light">
```
kdzwinel: Shouldn't we do `<Text>` before doing this? Or are you planning to come back to these components and fix it when `<Text>` is ready? Or ,for some reason, we are skipping `<Text>`?

anna-wro: I can do <Text> after I'm done with <ContentBox>, then I'll come back and fix it everywhere

kdzwinel: 👍

PR #1082, opened 2017-07-07
Avatar refactor

matikrk:
* sizes -> SIZE
* iconTypes -> ICON_TYPE
* iconColors -> ICON_COLOR
* 1 component - 1 file
* avatar examples - profile icon


src/components/avatar/AvatarContent.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon, {types as ICON_TYPE, colors as ICON_COLOR} from '../icons/Icon';
+import {SIZE} from './avatarConstants';
+
+const AvatarContent = ({imgSrc, iconType, iconColor, iconSize}) => {
+  if (imgSrc) {
+    return <img className="sg-avatar__image" src={imgSrc}/>;
```
aju: I don't know if its good idea to split it that much, its same component @kdzwinel @vergilius

src/components/avatar/AvatarContent.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon, {types as ICON_TYPE, colors as ICON_COLOR} from '../icons/Icon';
```
vergilius: it's kind of weird / tricky to pass this around.
I mean, it's a part of icon module, but we're exporting it in two places (icon/avatar-content) with different identifiers. @aju @kdzwinel wdyt?

vergilius: we're exporting it even in avatar module, again as ICON_TYPE

vergilius: I mean, there should be single source of this, right?

aju: I think it should be not reexported no point in that

src/components/avatar/AvatarContent.jsx
vergilius: If we don't use it separately, I'd suggest to merge it

matikrk: IMHO splitting everywhere is no needed, but here it's gave more readability.

aju: If we can't use something in separtion I think we should not split it

src/components/avatar/AvatarContent.jsx
matikrk: Will be more messy if you need import icon type from Icon module and then use it in avatar. Now you use avatar with dependencies exported from same file.

matikrk: All imports in constants file

PR #1088, opened 2017-07-10
Primary round button color blue -> green

matikrk:
Primary round button color blue -> green

screen shot 2017-07-10 at 14 17 01

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "86.0.0",
+  "version": "86.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "86.0.0",
+  "version": "86.1.0",
```
aju: its a fix :P but ok

matikrk: ye i know, i should give patch

PR #1095, opened 2017-07-11
Badges react

### LINKED ISSUE #1034: JSX: badges

anna-wro:
screen shot 2017-07-11 at 15 59 04

Closes #1034


src/components/badges/badge.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+test('size', () => {
+  const badge = shallow(
+    <Badge size={SIZE.SMALL}>Test</Badge>
+  );
+
+  expect(badge.hasClass('sg-badge--small')).toEqual(true);
+});
+
+test('default font size', () => {
+  const badge = mount(
```
matikrk: You can test if using Text component in render method. And then check if you use correct props, not directly checking text css class (it's part of Text.spec)

src/components/badges/badge.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+    <Badge>Test</Badge>
+  );
+  const text = badge.find(Text);
+
+  expect(text.hasClass('sg-text--xsmall')).toEqual(true);
+});
+
+test('larger font size', () => {
```
matikrk: what behaviour have small font?

src/components/badges/badge.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+    <Badge size={SIZE.LARGE}>Test</Badge>
+  );
+  const text = badge.find(Text);
+
+  expect(text.hasClass('sg-text--xsmall')).toEqual(false);
+});
+
+test('color', () => {
+  const badge = mount(
+    <Badge color={COLOR.MINT_SECONDARY}>Test</Badge>
```
matikrk: shallow copy and check if passing props correctly - single responsibility

src/components/badges/badge.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+test('rounded', () => {
+  const badge = shallow(
+    <Badge rounded={true}>Test</Badge>
+  );
+
+  expect(badge.hasClass('sg-badge--rounded')).toEqual(true);
+});
+
+test('error when no child', () => {
```
matikrk: pls add fail test when passing some div instead of string to show badge behaviour

anna-wro: Done! Thanks for your suggestions:)

src/components/badges/Badge.jsx
```diff
@@ -0,0 +1,47 @@
...
+import Text, {SIZE as TEXT_SIZE, WEIGHT as TEXT_WEIGHT} from '../text/Text';
+
+const SIZE = {
+  NORMAL: 'normal',
+  SMALL: 'small',
+  LARGE: 'large'
+};
+
+const COLOR = {
+  NORMAL: {BADGE: 'light', TEXT: 'default'},
```
kdzwinel: text colors are defined in Text component, you should not use strings here but the constants. Do sth like this instead:

```javascript
import Text, {COLOR as TEXT_COLOR} from '../text/Text';
//...
NORMAL: {BADGE: 'light', TEXT: TEXT_COLOR.DEFAULT},
```

matikrk: good point 👍

anna-wro: You're right, I've changed it 👍

PR #1097, opened 2017-07-12
Text Input component

matikrk:


src/components/form-elements/TextInput.jsx
```diff
@@ -0,0 +1,60 @@
...
+const TextInput = props => {
+  const {
+    type = 'text',
+    spacedBottom = true,
+    valid = VALIDATION.UNDEFINED,
+    size = SIZE.NORMAL,
+    color = COLOR.NORMAL,
+    fullWidth,
+    value,
+    ...additionalProps
```
aju: 👍

src/components/form-elements/TextInput.jsx
```diff
@@ -0,0 +1,60 @@
...
+  } = props;
+
+  const inputClass = classnames('sg-input', {
+    'sg-input--spaced-bottom': spacedBottom,
```
aju: do we still use this option ? :(

matikrk: 3 places:
brainly/openstudy-search-landing-page  landing-toplayer-email.hbs
brainly/srv-mirko-generator – index.html
brainly/symfony – splash_screen.html.twig

aju: I think we could fix brainly/symfony – splash_screen.html.twig and don't care about the rest :D @kdzwinel @vergilius WDYT?

matikrk: I can delete it from component and create issue in styleguide to delete it from css

kdzwinel: no need to create an issue in symfony. just bump the major version and leave info in tag notes

matikrk: 👍

PR #1100, opened 2017-07-12
Labels react

anna-wro:
screen shot 2017-07-12 at 12 53 57


src/components/labels/Label.jsx
```diff
@@ -0,0 +1,55 @@
...
+      <div className="sg-label__number">{number}</div>
+    </div>;
+  }
+
+  return <div className={labelClass}>
+    <div className="sg-label__icon">
+      <Icon type={iconType} color={iconColor} size={size.ICON_SIZE}/>
+    </div>
+    <div className="sg-label__text">{text}</div>
+  </div>;
```
kdzwinel: this code is almost the same as above - maybe we can do sth like {name ? <div>bla</div> : null}. @matikrk @aju WDYT?

src/components/labels/Label.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const labelClass = classNames('sg-label', {
+    [`sg-label--${size.CLASS_NAME}`]: size !== SIZE.NORMAL,
+    'sg-label--secondary': secondary,
+    'sg-label--unstyled': unstyled,
+    'sg-label--emphasised': emphasised,
+    'sg-label--elements-to-the-top': elementsToTop
+  });
+
+  if (number) {
```
matikrk: better use variable 
```
let numberElement;
if(number){
numberElement=     <div className="sg-label__number">{number}</div>;
}
```

anna-wro:
@kdzwinel, @matikrk - I changed it

src/components/labels/Label.jsx
anna-wro: Thanks!

src/components/labels/Label.jsx
anna-wro: I did what @matikrk suggested

PR #1101, opened 2017-07-12
Form elements react components - Select

matikrk:
* select component, tests, examples
* textarea - delete type property
* fix warnings in pages


src/components/form-elements/Select.jsx
```diff
@@ -0,0 +1,58 @@
...
+const VALIDATION = {
+  VALID: true,
+  INVALID: false,
+  UNDEFINED: undefined
+};
+
+
+const Select = props => {
+  const {
+    valid = VALIDATION.UNDEFINED,
```
aju: maybe state ?

src/components/form-elements/Select.jsx
```diff
@@ -0,0 +1,58 @@
...
+    fullWidth,
+    value,
+    options = [],
+    ...additionalProps
+  } = props;
+
+  const selectClass = classnames('sg-select', {
+    'sg-select--valid': valid === VALIDATION.VALID,
```
aju: It should not be possible to have valid and invalid on same element

src/components/form-elements/pages/text-input.jsx
```diff
@@ -8,10 +8,10 @@ const textInputs = () => <div>
     <TextInput placeholder="placeholder"/>
   </DocsBlock>
   <DocsBlock info="Valid">
-    <TextInput placeholder="placeholder" valid={true} value="This is valid example"/>
+    <TextInput placeholder="placeholder" valid={true} value="This is valid example" onChange={() => undefined}/>
```
aju: why onChange empty?

src/components/form-elements/Select.jsx
matikrk: So it isn't :)

src/components/form-elements/pages/text-input.jsx
matikrk: to get off warnings when creating our examples

aju: then make this function optional

src/components/form-elements/Select.jsx
aju: lets leave it as it is

PR #1102, opened 2017-07-12
Using <Label> component

anna-wro:


src/components/labels/Label.jsx
```diff
@@ -31,13 +35,15 @@ const Label = ({text, number, iconType, iconColor,
...
     </div>
-    <div className="sg-label__text">{text}</div>
+    {textElement}
     {numberElement}
+    {children}
   </div>;
 };
 
 Label.propTypes = {
-  text: PropTypes.string.isRequired,
```
matikrk: do we need both text and children?

anna-wro: In action list there's an example with <Text> component, not with `sg-label__text` class

anna-wro: only `sg-text--obscure` etc.

PR #1119, opened 2017-07-18
Footer react

### LINKED ISSUE #1040: JSX: footer

anna-wro:
screen shot 2017-07-18 at 09 25 51

Closes #1040


src/components/text/Link.jsx
```diff
@@ -18,7 +18,8 @@ const SIZE = {
   OBSCURE: 'obscure'
 };
 
-const Link = ({children, href = '#', color, size = SIZE.NORMAL, unstyled, underlined, emphasised, disabled}) => {
+const Link = ({children, href = '#', color, size = SIZE.NORMAL, unstyled, underlined, emphasised, disabled,
```
aju: Maybe we should refactor it in the way that if link does not have href set then its disabled?

aju: @vergilius @kdzwinel

anna-wro: @kdzwinel, @matzimowski, @burakukula, @matikrk, @vergilius what do you think about it?

matzimowski:
@anna-wro imho, we decided to stay with `sg-footer` as `<footer>`, not `sg-layout__footer`. we gonna remove `sg-layout__footer` in further steps. could you just replace `<footer>` in `<div>` again in layout component (not to have `<footer><footer></footer></footer>`?

anna-wro:
@matzimowski  so `<Footer>` would be 

```
<footer className="sg-footer">
  <div className="sg-footer__container">
    {children}
  </div>
</footer>;
```

and in `<Layout>`
`footerContent = <div className="sg-layout__footer">{footer}</div>;` ? Is that correct?

matzimowski:
for now yes. It would be correct, then I will refactor layout and get rid of useless wrapper

PR #1123, opened 2017-07-18
New builder

matikrk:
* delete jekkyl
* rename gulp task
* iframe pages in different folder
* add doctype definition on pages
* add slugify lib by npm
* cleaning (something left for new PR)


src/docs/layouts/Navigation.jsx
```diff
@@ -0,0 +1,171 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import slugify from 'slugify';
+
+const someJSScript = `
```
kdzwinel: maybe lets put it in a separate file and import it?

vergilius: +1

kdzwinel: we decided to do it in the next PR

PR #1143, opened 2017-07-21
Inputs validation

### LINKED ISSUE #1141: Inputs should have 2 properties (valid, invalid) instead of one (valid)
Properites should be true/false only

anna-wro:
* true/false → valid, invalid
Closes #1141

### DIFF
--- a/src/components/form-elements/Select.jsx
+++ b/src/components/form-elements/Select.jsx
@@ -15,7 +15,7 @@ const Select = props => {
   if (valid && invalid) {
     throw {
       name: 'WrongValidation',
-      message: 'TextInput can be either valid or invalid!'
+      message: 'Select can be either valid or invalid!'
     };
   }
 
--- a/src/components/form-elements/Select.jsx
+++ b/src/components/form-elements/Select.jsx
@@ -2,16 +2,10 @@ import React from 'react';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 
-const VALIDATION = {
-  VALID: true,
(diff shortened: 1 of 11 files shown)


src/components/form-elements/Select.jsx
```diff
@@ -2,25 +2,19 @@ import React from 'react';
...
     value,
     options = [],
     ...additionalProps
   } = props;
 
   const selectClass = classnames('sg-select', {
-    'sg-select--valid': valid === VALIDATION.VALID,
-    'sg-select--invalid': valid === VALIDATION.INVALID,
+    'sg-select--valid': valid,
+    'sg-select--invalid': invalid,
```
aju: check if both are true throw error
@kdzwinel @matikrk @vergilius

matikrk: 👍 
console.error()

kdzwinel: rather `throw new Error("bla")`?

aju: @kdzwinel it can crash application
maybe console.error is better option?

kdzwinel: if both will be true then it's a bug and we should crash IMO

aju: yeah, lets go with the throw!

anna-wro: Done!

matikrk: imho  better is console error instead of throw whole application.

PR #1147, opened 2017-07-24
Interactive improvements - 💄

kdzwinel:
screen shot 2017-07-24 at 09 20 56

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "sass-lint": "^1.10.2",
     "slug": "^0.9.1",
     "through2": "^2.0.3",
-    "uuid": "^3.1.0",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -69,6 +68,7 @@
     ]
   },
   "dependencies": {
-    "classnames": "^2.2.5"
+    "classnames": "^2.2.5",
+    "node-uuid": "^1.4.8"
```
kdzwinel: `uuid` had a dependency on native nodejs module `crypto`. This won't work on front-end. `node-uuid` has no dependencies and a lot of GH stars.

matikrk: 👍 i used it on on test, so it could be like this.

PR #1154, opened 2017-07-24
Interactive buttons

kdzwinel:
Buttons section:

screen shot 2017-07-24 at 17 24 43

Show "(default)" value where prop is not required:
screen shot 2017-07-24 at 16 21 26


src/components/avatar/pages/avatar-interactive.jsx
```diff
@@ -3,12 +3,24 @@ import Avatar, {SIZE} from '../Avatar';
...
+    },
+    {
+      name: 'spaced',
+      values: Boolean
+    },
+    {
+      name: 'imgSrc',
+      values: String
+    }
+  ];
```
kdzwinel: I had to change the format to allow more options (e.g. 'required') - I wish we could read that from propTypes :(

PR #1200, opened 2017-07-25
Icon as a button interactive

kdzwinel:
screen shot 2017-07-25 at 11 55 17


src/docs/page-components.jsx
```diff
@@ -4,36 +4,24 @@ import Avatars from '../components/avatar/pages/avatar-interactive';
...
-    <Badges />
-  </article>
-  <article>
-    <h2 className="article-header" id="buttons">
-      Buttons
-      <a href="#buttons" className="permalink">#</a>
-    </h2>
-    <Buttons />
-  </article>
-</div>;
```
kdzwinel: below fix is to avoid repetition

src/docs/page-components.jsx
```diff
@@ -4,36 +4,24 @@ import Avatars from '../components/avatar/pages/avatar-interactive';
 import Bubbles from '../components/bubble/pages/bubble-interactive';
 import Badges from '../components/badges/pages/badges-interactive';
 import Buttons from '../components/buttons/pages/buttons-interactive';
+import IconsAsButtons from '../components/icon-as-button/pages/icon-as-button-interactive';
+import slugify from 'slugify';
```
matikrk: I'm not sure, but you have bad library.
You should use local file src/docs/slugify.js

kdzwinel: good catch, although I just spotted that we can't use `slugify` it's just too big. It increases bundle size 4 times. I'll find something smaller.

PR #1237, opened 2017-08-01
interactive overlay

### LINKED ISSUE #1185: Overlay

anna-wro:
screen shot 2017-08-01 at 16 26 09
Closes #1185


src/components/overlay/pages/overlay-interactive.jsx
```diff
@@ -0,0 +1,26 @@
...
+import DocsActiveBlock from 'components/DocsActiveBlock';
+import Overlay from '../Overlay';
+
+const Overlays = () => {
+  const settings = [];
+
+  return <div>
+    <DocsActiveBlock settings={settings}>
+      <div>
+        <Overlay/>
```
kdzwinel: huh, that's how we use overlay? I'd make more sense to make it wrap content it's covering IMO. I'd make it more flexible (you can cover only part of page). @anna-wro @matikrk @aju WDYT?

anna-wro: we used it this way in the style guide, I even checked with much older versions (70-80)

anna-wro: I checked in other projects and we also use it in the way you described (wrapping content).

anna-wro:
@kdzwinel I updated the example

anna-wro:
screen shot 2017-08-07 at 08 09 28

kdzwinel:
@matikrk @aju WDYT about making overlay `absolute` by default and having `fixed` as an option? This way we could use overlay to cover only some part of the page (e.g. payment iframe, everything w/o the header, etc.), not only to always cover the whole page.

aju:
@kdzwinel I'm ok with it. It would just require some refactoring

anna-wro:
@aju, @kdzwinel - I did as you suggested
screen shot 2017-08-08 at 08 50 52

anna-wro:
screen shot 2017-08-08 at 11 19 32

Changed default positioning - it's fixed now, you can change it to absolute by checking the 'partial' prop

PR #1243, opened 2017-08-07
Add pointer-events: none to icons

matzimowski:
it was causing some issues when element had an icon inside and click event was bound (no click event fired).


kdzwinel:
check if icon-as-a-button, and other places where icons SHOULD be clickable still work

PR #1253, opened 2017-08-11
JSX generator without default props

### LINKED ISSUE #1155: Do not show props with default values in JSX
If some property has a default value, do not show it in JSX.

`<Bla xyz='abc' />` -> (assuming that 'abc' is default value of 'xyz') -> `<Bla />`

anna-wro:
before:
screen shot 2017-08-10 at 15 05 19

after:
screen shot 2017-08-10 at 15 05 44

Closes #1155

### DIFF
--- a/src/components/icon-as-button/IconAsButton.jsx
+++ b/src/components/icon-as-button/IconAsButton.jsx
@@ -23,7 +23,7 @@ const SIZE = {
   XXSMALL: {className: 'xxsmall', iconSize: 10}
 };
 
-const IconAsButtonContent = ({children, overlay}) => {
+const IconAsButtonContent = ({children = null, overlay}) => {
   if (!overlay) {
     return children;
   }
--- a/src/docs/components/DocsActiveBlock.jsx
+++ b/src/docs/components/DocsActiveBlock.jsx
@@ -63,20 +63,20 @@ class DocsActiveBlock extends Component {
 
   getPropsWithDefaults(component) {
     const settings = this.props.settings;
-    let unchangedProps, propsWithDefaults;
 
-    if (Array.isArray(settings)) {
-      unchangedProps = Object.keys(component.props)
-        .filter(key => {
-          const propSettings = settings.find(setting => setting.name === key);
-
-          return !!(!propSettings || propSettings.required);
-        });
-
-      propsWithDefaults = Object.keys(component.props)
-        .filter(key => unchangedProps.indexOf(key) === -1);
+    if (!Array.isArray(settings)) {
+      return [];
     }
-    return propsWithDefaults;
+
+    const unchangedProps = Object.keys(component.props)
+      .filter(key => {
+        const propSettings = settings.find(setting => setting.name === key);
+
+        return !!(!propSettings || propSettings.required);
+      });
+
+    return Object.keys(component.props)
+      .filter(key => unchangedProps.indexOf(key) === -1);
   }
 
   render() {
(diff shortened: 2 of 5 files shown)


src/docs/components/DocsActiveBlock.jsx
```diff
@@ -61,17 +61,37 @@ class DocsActiveBlock extends Component {
...
+
   render() {
     let {contentBefore, contentAfter} = this.props;
     const {centeredItems = true, wrapper} = this.props;
     let component;
     let code;
+    let propsWithDefaults;
 
     if (this.state.renderNormally) {
       component = React.cloneElement(this.props.children, this.state.props);
```
kdzwinel: Can we refactor it into
`component = React.cloneElement(this.props.children, removePropsWithDefaultValues(this.state.props));`

and don't touch JSXGenerator?

src/components/icon-as-button/IconAsButton.jsx
```diff
@@ -23,7 +23,7 @@ const SIZE = {
   XXSMALL: {className: 'xxsmall', iconSize: 10}
 };
 
-const IconAsButtonContent = ({children, overlay}) => {
+const IconAsButtonContent = ({children = null, overlay}) => {
```
anna-wro: I added this because there were some issues with generating JSX without it

src/docs/components/DocsActiveBlock.jsx
```diff
@@ -61,17 +61,37 @@ class DocsActiveBlock extends Component {
     this.setState({renderNormally: false}, () => this.setState({renderNormally: true}));
   }
 
+  getPropsWithDefaults(component) {
+    const settings = this.props.settings;
+    let unchangedProps, propsWithDefaults;
+
+    if (Array.isArray(settings)) {
+      unchangedProps = Object.keys(component.props)
```
matikrk: unchangedProps can be const

src/docs/components/DocsActiveBlock.jsx
```diff
@@ -61,17 +61,37 @@ class DocsActiveBlock extends Component {
     this.setState({renderNormally: false}, () => this.setState({renderNormally: true}));
   }
 
+  getPropsWithDefaults(component) {
+    const settings = this.props.settings;
+    let unchangedProps, propsWithDefaults;
+
+    if (Array.isArray(settings)) {
```
matikrk: what happen  if settings is not array?  returning empty array?
We can invert if for fast return values
```
if(!Array.isArray(settings)){
return; // undefined or should be here empty array 
}
rest of code
```

src/docs/components/DocsActiveBlock.jsx
```diff
@@ -61,17 +61,37 @@ class DocsActiveBlock extends Component {
...
     let code;
+    let propsWithDefaults;
 
     if (this.state.renderNormally) {
       component = React.cloneElement(this.props.children, this.state.props);
+      propsWithDefaults = this.getPropsWithDefaults(component);
 
       if (this.state.showCode === 'jsx') {
-        const jsx = generateJSX(component);
+        const jsx = generateJSX(component, propsWithDefaults);
```
matikrk: propsWithDefaults should be const and defined here inside if, we don't used it in other place in render()

src/docs/components/JSXGenerator.js
```diff
@@ -20,16 +21,25 @@ function propToString(prop, inJS = false) {
...
     return component;
   }
 
   const type = component.type.name || component.type;
-
   let jsxProps = Object.keys(component.props)
     .filter(key => key !== 'children')
     .filter(key => component.props[key] !== undefined)
+    .filter(key => {
+      if (Array.isArray(propsWithDefaults) && propsWithDefaults.indexOf(key) > -1) {
```
matikrk: if getPropsWithDefaults return empty array we don't need checking array

anna-wro: @matikrk I just tried returning empty array, it doesn't work

anna-wro: we need this if statement

matikrk: you got there undefined? if yes you can use default values in function

src/docs/components/DocsActiveBlock.jsx
anna-wro: @kdzwinel I did it it another [pull request](https://github.com/brainly/style-guide/pull/1256/) :)

PR #1256, opened 2017-08-18
JSXGenerator refactor ⚙

anna-wro:
no need to pass propsWithDefaults to JSXGenerator

### DIFF
--- a/src/docs/components/DocsActiveBlock.jsx
+++ b/src/docs/components/DocsActiveBlock.jsx
@@ -61,22 +61,35 @@ class DocsActiveBlock extends Component {
     this.setState({renderNormally: false}, () => this.setState({renderNormally: true}));
   }
 
-  getPropsWithDefaults(component) {
+  getPropsToRemove(props) {
     const settings = this.props.settings;
 
     if (!Array.isArray(settings)) {
       return [];
     }
 
-    const unchangedProps = Object.keys(component.props)
+    const unchangedProps = Object.keys(props)
       .filter(key => {
-        const propSettings = settings.find(setting => setting.name === key);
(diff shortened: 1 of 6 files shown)


src/docs/components/DocsActiveBlock.jsx
```diff
@@ -61,22 +61,35 @@ class DocsActiveBlock extends Component {
...
       });
 
-    return Object.keys(component.props)
-      .filter(key => unchangedProps.indexOf(key) === -1);
+    return Object.keys(props)
+      .filter(key => unchangedProps.indexOf(key) > -1);
+  }
+
+  removePropsWithDefaultValues(props) {
+    const fewerProps = Object.assign(props);
```
matikrk: `Object.assign({},props); `
to make copy

src/docs/components/DocsActiveBlock.jsx
```diff
@@ -61,22 +61,35 @@ class DocsActiveBlock extends Component {
...
 
     if (!Array.isArray(settings)) {
       return [];
     }
 
-    const unchangedProps = Object.keys(component.props)
+    const unchangedProps = Object.keys(props)
       .filter(key => {
-        const propSettings = settings.find(setting => setting.name === key);
+        const propSettings = settings.find(setting => setting.name === key.value);
```
matikrk: why we have this change?

anna-wro: I changed it back to `setting.name === key`

src/docs/components/DocsActiveBlock.jsx
anna-wro: Fixed, thanks!

PR #1258, opened 2017-08-21
fix bug in rating

matikrk:


src/docs/index2.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
vergilius: aw?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "111.0.1",
+  "version": "111.0.3",
```
aju: :O

matikrk: some problems on github. now 111.0.2

aju: change it to 111.0.2 :)

PR #1263, opened 2017-08-25
Update eslint

aju:


src/components/action-list/pages/action-list.jsx
```diff
@@ -82,31 +82,34 @@ const ActionLists = () =>
...
+            start
+          </ButtonSecondary>
         </ActionListHole>
       </ActionList>
     </DocsBlock>
 
     <DocsBlock info='No wrap' multiContent={[
-      <ContrastBox narrow={true}>
-        <ActionList noWrap={true}>
+      <ContrastBox key={1} narrow>
```
matikrk: why here key?
writing keys is on DocBlack side

aju: Elements in arrays need to have a key

PR #1267, opened 2017-08-25
Fix bind issue

aju:


src/docs/components/CodeBlock.jsx
```diff
@@ -1,12 +1,17 @@
...
 import beautify from 'html_beautify';
 import hljs from 'hljs';
 import IconAsButton, {TYPE, COLOR} from 'icon-as-button/IconAsButton';
 
 class CodeBlock extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.copyCodeFunction = this.copyCodeFunction.bind(this);
```
matikrk: delete it

src/docs/components/CodeBlock.jsx
```diff
@@ -15,14 +20,17 @@ class CodeBlock extends Component {
...
-    // eslint-disable-next-line no-undef
-    new Clipboard(copyButton, {
+    new window.Clipboard(copyButton, {
       target() {
         return copyHelperCode;
       }
     });
   }
 
+  copyCodeFunction(node)  {
```
matikrk: copyCodeFunction=node => {

aju: not supported yet in SG

PR #1270, opened 2017-08-25
Cleanup jsxgenerator

kdzwinel:


src/docs/components/JSXGenerator.js
```diff
@@ -30,13 +29,6 @@ function generateJSX(component) {
   let jsxProps = Object.keys(component.props)
     .filter(key => key !== 'children')
     .filter(key => component.props[key] !== undefined)
-    .filter(key => {
-      const componentWithoutProp = React.cloneElement(component, {[`${key}`]: undefined});
-      const codeWithProp = ReactDOMServer.renderToStaticMarkup(component);
-      const codeWithoutProp = ReactDOMServer.renderToStaticMarkup(componentWithoutProp);
-
-      return codeWithProp !== codeWithoutProp;
-    })
```
kdzwinel: AFAIK this was moved to `DocsActiveBlock.removePropsWithDefaultValues`

anna-wro: not all of it - we still need to compare html to not generate jsx for default values, like normal size in Label

kdzwinel: good point, thanks for catching that!

kdzwinel:
Phew, updated. @anna-wro PTAL

anna-wro:
@kdzwinel you forgot to bump the version:)

kdzwinel:
yeah, I'm lazy :) This change doesn't influence styleguide in any way. Only the docs, so it doesn't need a version bump IMO

PR #1279, opened 2017-09-07
subject mono icon - change color

### LINKED ISSUE #1278: Way to combine icon colors with subject icons
It should be possible to use mono option from subject icons and selected color. Right now styles from subject icons are overriding icon colors:

screen shot 2017-09-07 at 11 38 46

anna-wro:
Before:
screen shot 2017-09-07 at 14 46 49

After:
screen shot 2017-09-07 at 14 47 43

Closes #1278

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "112.2.1",
+  "version": "113.0.0",
   "description": "Brainly Front-End Style Guide",
   "author": "Brainly",
   "private": true,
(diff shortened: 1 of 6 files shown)


matikrk:
IMHO we should rename mono to color.
WDYT?

kdzwinel:
Maybe we should keep `mono` as a boolean and have `color` as a separate property? One that would only work with mono icons. WDYT?

vergilius:
I have thought that it would be best to integrate sg-icon colors with subject icons, but this would cause to bind these two together, which might not be best option.

Either way, there will be some duplicated modifier classes.

PR #1289, opened 2017-09-14
Validated input size fix

### LINKED ISSUE #1287: input changes width and height when invalid
Switching input state from default to invalid causes it to change width and height (as far as I can see because of the font weight change). This looks terrible.

Same thing for the "valid" state.

input-bug mov

anna-wro:
Another way would be to use [`-webkit-text-stroke`](http://caniuse.com/#feat=text-stroke) - WDYT?

screen shot 2017-09-14 at 13 39 59
screen shot 2017-09-14 at 13 39 44
screen shot 2017-09-14 at 13 40 06

Fixes #1287


src/components/form-elements/_text-input.scss
```diff
@@ -103,12 +103,18 @@ $includeHtml: false !default;
 
     &--valid {
       color: $inputFocusValidBorderColor;
-      font-weight: $fontWeightBold;
+      text-shadow: -0.2px -0.2px 0 $inputFocusValidBorderColor,
+      0.2px -0.2px 0 $inputFocusValidBorderColor,
+      -0.2px 0.2px 0 $inputFocusValidBorderColor,
+      0.2px 0.2px 0 $inputFocusValidBorderColor;
```
kdzwinel: Yeah, let's try `text-stroke` :) Support looks OK to me.

anna-wro: changed!

src/components/form-elements/_text-input.scss
```diff
@@ -103,12 +104,12 @@ $includeHtml: false !default;
 
     &--valid {
       color: $inputFocusValidBorderColor;
-      font-weight: $fontWeightBold;
+      -webkit-text-stroke: 0.5px $inputFocusValidBorderColor;
```
kdzwinel: do we need a vendor prefix? I thought sass is adding it by itself?

anna-wro: It didn't work without it :/

kdzwinel: hm, and does it work in firefox/safari/etc.? we will probably also need `-moz` etc.?

src/components/form-elements/_text-input.scss
```diff
@@ -1,3 +1,4 @@
+// sass-lint:disable no-vendor-prefixes
```
kdzwinel: can you disable this rule only for single lines that need it, not the whole file?

anna-wro: sure

src/components/form-elements/_text-input.scss
anna-wro: I checked in both of them and it works

kdzwinel:
Thanks for fixing it 🤘

PR #1291, opened 2017-09-14
textarea upgrade

matikrk:
* custom component Type

### DIFF
--- a/src/components/form-elements/Textarea.jsx
+++ b/src/components/form-elements/Textarea.jsx
@@ -19,6 +19,7 @@ const Textarea = props => {
     autoHeight,
     value,
     className,
+    Type = 'textarea',
     ...additionalProps
   } = props;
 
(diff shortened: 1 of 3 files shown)


src/components/form-elements/Textarea.jsx
```diff
@@ -38,11 +39,11 @@ const Textarea = props => {
     'sg-textarea--auto-height': autoHeight
   }, className);
 
-  return <textarea className={textareaClass} value={value} {...additionalProps} />;
+  return <Type className={textareaClass} value={value} {...additionalProps} />;
 };
 
 Textarea.propTypes = {
-  type: PropTypes.string,
+  Type: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
```
vergilius: why capitalized and why func allowed?

src/components/form-elements/Textarea.jsx
```diff
@@ -19,6 +19,7 @@ const Textarea = props => {
     autoHeight,
     value,
     className,
+    Type = 'textarea',
```
vergilius: shouldn't this be in some kind of object, like enum or something?

src/components/form-elements/Textarea.jsx
matikrk: string = normal component
func = react class

capitalized take variable  if small create html tag

src/components/form-elements/Textarea.jsx
matikrk: textarea it's default and we basically want only textarea or React class looking like textarea. Ex. react-textarea-autosize from npm :)

PR #1294, opened 2017-09-15
fix frontend-tools-configs in SG

matikrk:
give 'v' to github tag in frontend-tools-configs


Dockerfile
```diff
@@ -10,6 +10,6 @@ RUN apk add --no-cache git ruby python ruby-bundler build-base ruby-dev libffi-d
     npm install gulp http-server -g
 
 ADD package.json package.json
```
kdzwinel: should we also ADD yarn.lock?

matikrk: done

Issue #1290, opened 2017-09-14
sg-button-secondary - disabled state looks wrong


kdzwinel:
`sg-button-secondary sg-button-secondary--alt sg-button-secondary--disabled`

*How it should be*:
![screen shot 2017-09-14 at 15 15 10](https://user-images.githubusercontent.com/985504/30431687-ad55fc8c-995f-11e7-8275-4dd008f56d38.png)

(and it was like that since ~80.0.0 - https://styleguide.brainly.com/80.0.0/docs/components.html#buttons )

*How it is ATM* (since version ~90.0.0):

![screen shot 2017-09-14 at 15 17 00](https://user-images.githubusercontent.com/985504/30431714-cb343afc-995f-11e7-9d58-0f68b68fb64d.png)

We get white background + red text for all secondary buttons when they are disabled.

PR #1295, opened 2017-09-15
fix sg-button-secondary - disabled

matikrk:
fixes: #1290


kdzwinel:
can you add a test for this case?

PR #1299, opened 2017-09-26
Icons and logos accessibility

### LINKED ISSUE #355: Accessibility - research and add titles (and descriptions?) to icons
Check what we need to do to improve icon accessibility and implement that.

anna-wro:
add svg titles for icons and logos

Closes #355


src/images/logos/eodev-mobile.svg
```diff
@@ -2,6 +2,7 @@
   <style>
     .st0{fill:#FFFFFF;} .st1{fill:#020A1B;}
   </style>
+  <title>eodev mobile logo</title>
```
kdzwinel: I think we can skip "mobile". this is for page visitors - I think that they won't care if logo is mobile or not

anna-wro: fixed!

PR #1301, opened 2017-09-28
List - no default icon

matikrk:
List - no default icon 
List - no default icon  update tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "115.3.0",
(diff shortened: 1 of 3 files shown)


matikrk:
No one using it so i put minor :)

kdzwinel:
@matikrk if something is a major change, as this PR, please mark it as a major change. You haven't checked all projects using SG.

matikrk:
There was no change in css. Just in React component and I've checked all our projects

PR #1306, opened 2017-10-05
Feature/update react

aju:


package.json
```diff
@@ -52,14 +53,17 @@
     "scss-unused-variables": "./scripts/find_scss_unused_variables.sh src/",
     "build": "gulp build",
     "start": "http-server ./dist -p 8000",
-    "prepush": "npm test",
+    "prepush": "yarn test",
```
matikrk: why you change this? There is no difference.
But if you wanna change do it also on line 60, task test

aju: why mix yarn and npm ?:D
I made changes as you requested

matikrk: everywhere yarn is ok 👍

src/components/content-box/pages/content-box.jsx
```diff
@@ -78,7 +78,7 @@ const examplePart1 =
       </OverlayedBox>
       <SeparatorVertical />
       <Avatar size={AVATAR_SIZE.SMALL} imgSrc="https://source.unsplash.com/64x64/?dog" />
-      <Avatar size={AVATAR_SIZE.SMALL} iconType={ICON_TYPE.PROFILE} iconColor={ICON_COLOR.GRAY} />
```
kdzwinel: what happened here?

PR #1309, opened 2017-10-09
Navy blue button

aju:
screen shot 2017-10-09 at 16 05 50
screen shot 2017-10-09 at 16 05 53
screen shot 2017-10-09 at 16 06 17


kdzwinel:
navy blue was an 'additional' color - we were not supposed to use it for other things than ranks as far as I remember

aju:
screen shot 2017-10-09 at 16 10 55

The same will be done for not logged

matzimowski:
does it mean we should move this color to basic ones?

kdzwinel:
Exactly. If they want to use it all over the place it's no longer 'additional' color. BTW `DARK` looks almost the same

PR #1312, opened 2017-10-20
fix action list hole styles

matikrk:
in WC fix we had good order

### DIFF
--- a/src/components/action-list/action-list.scss
+++ b/src/components/action-list/action-list.scss
@@ -48,9 +48,6 @@
     align-items: flex-start;
   }
 
-
-
-
   &--centered {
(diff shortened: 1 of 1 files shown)


src/components/action-list/action-list.scss
```diff
@@ -44,6 +48,9 @@
     align-items: flex-start;
   }
 
+
```
vergilius: newlines?

src/components/action-list/action-list.scss
```diff
@@ -64,6 +71,8 @@
     }
   }
 
+
```
vergilius: newlines?

PR #1315, opened 2017-10-24
Limit height of toplayer--splash-screen

matzimowski:
before:
screen shot 2017-10-24 at 16 08 58

before (edge case):
screen shot 2017-10-24 at 16 11 01

after:
screen shot 2017-10-24 at 16 09 09

### DIFF
--- a/src/components/toplayer/_toplayer.scss
+++ b/src/components/toplayer/_toplayer.scss
@@ -95,6 +95,9 @@ $includeHtml: false !default;
       @include sgBreakpoint(medium-up) {
         min-height: 0;
         height: 100%;
+        // sass-lint:disable no-duplicate-properties
+        height: calc(100% - #{gutter(2)});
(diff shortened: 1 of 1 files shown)


src/components/toplayer/_toplayer.scss
```diff
@@ -93,7 +93,9 @@ $includeHtml: false !default;
       // sass-lint:enable no-duplicate-properties
 
       @include sgBreakpoint(medium-up) {
-        min-height: calc(100% - #{gutter(2)});
+        min-height: 0;
+        height: 100%;
+        max-height: gutter(33);
```
matzimowski: 792px

kdzwinel:
height: calc(100% - #{gutter(2)})
    max-height: gutter(33);

matzimowski:
@kdzwinel good catch, updated

Issue #1334, opened 2017-11-03
change exporter constants naming in SG - COLOR


matikrk:
same as https://github.com/brainly/style-guide/issues/1331 but COLOR
* Same naming convention
* merge same colors:
  * Icon, subject icon, icon as button, math symbols
  * Header (prim,sec) - 1 color so maybe flag
  * Text, Link
  * Text Bit
  * Input

PR #1338, opened 2017-11-06
change exported COLOR to ICON_COLOR in Icons

matikrk:
part of issue #1334
* all Icon-* files import form Icon.jsx
* create mixin in scss for coloring icons
* all exported thing with name ICON_COLOR point to same object

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "121.0.1",
+  "version": "122.0.0",
   "description": "Brainly Front-End Style Guide",
   "author": "Brainly",
   "private": true,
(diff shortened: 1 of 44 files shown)


matikrk:
working :)

src/components/icon-as-button/_icon-as-button.scss
```diff
@@ -166,6 +170,33 @@ $includeHtml: false !default;
...
+
+      &:hover,
+      &:active {
+        fill: $iconAsButtonBlueActiveColor;
+      }
+
+      .sg-icon-as-button__hole {
+        border-color: $iconAsButtonBlueColor;
+      }
+    }
```
kdzwinel: this should be in a separate PR, right?

matikrk: i add mising color to icaon_as _button to have same everywhere

kdzwinel:
Have you checked if interactive version of docs works correctly? Have you added new colors to docs?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "121.0.1",
+  "version": "122.0.0",
```
kdzwinel: this is a breaking change - you can't import "COLOR" from "icons/Icon"

matikrk:
interactive and normal working. Colors is based on array variable

package.json
matikrk: breaking change = major - so what it's wrong?

PR #1353, opened 2017-12-11
Add `textareaRef` prop

tomasz-sodzawiczny:
Sadly `enzyme` does not play nice with refs, so no way to test that 😞

### DIFF
--- a/src/components/form-elements/Textarea.jsx
+++ b/src/components/form-elements/Textarea.jsx
@@ -19,6 +19,7 @@ const Textarea = props => {
     autoHeight,
     value,
     className,
+    textareaRef,
     Type = 'textarea',
     ...additionalProps
   } = props;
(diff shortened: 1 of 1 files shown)


src/components/form-elements/Textarea.jsx
```diff
@@ -19,6 +19,7 @@ const Textarea = props => {
     autoHeight,
     value,
     className,
+    textareaRef,
```
aju: this can be null ? right?

matikrk:
Maybe rename it more generic, like refFunction? Later we can do it in onther component if needed

aju:
`Maybe rename it more generic, like refFunction? Later we can do it in onther component if needed` I agree

src/components/form-elements/Textarea.jsx
tomasz-sodzawiczny: Yes, it's safe

tomasz-sodzawiczny:
Disagree - this is not *some random ref*, this is **ref to the textarea**

This component's body could look like this:
```js
<div>
  <div>
    <textarea />
  </div>
</div>
```
Where should I put the `refFunction` then? Why not one of the divs?

We can also imagine some other component with multiple refs looking like this:
```js
const SomeComponent = ({cancelButtonRef, confirmButtonRef}) => {
  return (
    <div>
      <Button ref={cancelButtonRef}>Cancel</Button>
      <Button ref={confirmButtonRef}>Confirm</Button>
    </div>
  )
}
```
Notice how it's also consistent with other prop naming - e.g. we have `textareaClassName`, which follows the same schema.

PR #1358, opened 2017-12-18
Change type into buttonType for buttons

### LINKED ISSUE #1355: Buttons - change prop TYPE into something else
In some cases we would need to pass type into button, like type="button" or type="submit". We need to change current TYPE into something else

burakukula:
Fixes #1355

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -270,7 +270,7 @@ import ButtonPrimary, {types as buttonTypes} from '../../buttons/ButtonPrimary';
 
 Good:
 ```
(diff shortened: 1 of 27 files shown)


matikrk:
BUTTON_TYPE from ButtonPrimary it's different from ButtonSecondary
So it's should name same.

There is 2 ways:
* name it BUTTON_PRIMARY_TYPE & BUTTON_SECONDARY_TYPE
* or make it consistent - same types in both - IMHO better

WDYT @aju @kdzwinel @matzimowski ?

burakukula:
@matikrk like that? 🙃

PR #1359, opened 2017-12-18
Add verified, expert and student icons

matzimowski:
screen shot 2017-12-18 at 16 21 11

screen shot 2017-12-18 at 16 17 08
screen shot 2017-12-18 at 16 17 15


src/components/icons/Icon.jsx
```diff
@@ -14,11 +14,13 @@ const TYPE = {
   BOLD: 'bold',
   CAMERA: 'camera',
   CHECK: 'check',
+  CHECKED: 'checked',
```
kdzwinel: we already have 'check'. Lets use 'verified' here

matzimowski: agree

PR #1364, opened 2017-12-20
planet icon

matikrk:
screen shot 2017-12-20 at 13 23 28
screen shot 2017-12-20 at 13 22 46


src/images/icons/planet.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="32" width="32" preserveAspectRatio="xMidYMid"
+     xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 32 32">
```
kdzwinel: we dont' need some of these attributes, please copy from other icons

matikrk: it's from pencil

PR #1368, opened 2018-01-05
Extend text element

aju:
closes https://github.com/brainly/style-guide/issues/1341

### DIFF
--- a/src/components/text/Text.jsx
+++ b/src/components/text/Text.jsx
@@ -40,13 +40,41 @@ const WEIGHT = {
   BOLD: 'emphasised'
 };
 
-const Text = ({children, size, weight, color, className, type = TYPE.DIV, noWrap, breakWords, ...props}) => {
+const TEXT_TRANSFORM = {
+  UPPERCASE: 'uppercase',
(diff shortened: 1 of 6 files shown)


src/components/text/Text.jsx
```diff
@@ -40,13 +40,41 @@ const WEIGHT = {
   BOLD: 'emphasised'
 };
 
-const Text = ({children, size, weight, color, className, type = TYPE.DIV, noWrap, breakWords, ...props}) => {
+const TEXT_TRANSFORM = {
```
burakukula: @aju TEXT_WEIGHT, TEXT_COLOR, TEXT_SIZE?

burakukula: @aju or it would be too big refactor for now?

aju:
@vergilius 
screen shot 2018-01-05 at 15 08 23
screen shot 2018-01-05 at 15 08 39

src/components/text/Text.jsx
aju: One step at a time:P

PR #1387, opened 2018-02-02
Card component

burakukula:
screen shot 2018-02-02 at 17 07 26
screen shot 2018-02-02 at 17 07 01

### DIFF
--- a/src/components/card/Card.jsx
+++ b/src/components/card/Card.jsx
@@ -2,9 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-export const PADDING = {
-  SMALL: 'small-padding',
-  LARGE: 'large-padding'
+export const CARD_PADDING = {
+  SMALL: 'padding-small',
+  LARGE: 'padding-large'
 };
 
 const Card = ({children, full, centered, padding, shadow, noBorder, className, ...props}) => {
@@ -29,7 +29,7 @@ Card.propTypes = {
   centered: PropTypes.bool,
   noBorder: PropTypes.bool,
   shadow: PropTypes.bool,
-  padding: PropTypes.oneOf(Object.values(PADDING)),
+  padding: PropTypes.oneOf(Object.values(CARD_PADDING)),
   className: PropTypes.string
 };
 
--- a/src/components/card/Card.jsx
+++ b/src/components/card/Card.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export const PADDING = {
+  SMALL: 'small-padding',
+  LARGE: 'large-padding'
+};
+
+const Card = ({children, full, centered, padding, shadow, noBorder, className, ...props}) => {
+  const cardClass = classNames('sg-card', {
+    'sg-card--full': full,
+    'sg-card--with-shadow': shadow,
+    'sg-card--no-border': noBorder,
+    'sg-card--centered': centered,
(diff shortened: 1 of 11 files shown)


src/components/card/Card.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export const PADDING = {
+  SMALL: 'small-padding',
```
aju: padding-small ? padding-large?

src/components/card/_card.scss
```diff
@@ -0,0 +1,92 @@
...
+      .sg-card__hole {
+        display: flex;
+        flex-wrap: wrap;
+        align-items: center;
+        height: 100%;
+        justify-content: center;
+      }
+    }
+
+    &--small-padding {
```
aju: Maybe its better to make this modifiers on holes?

src/components/card/_card.scss
```diff
@@ -0,0 +1,92 @@
...
+@if ($includeHtml) {
+
+  .sg-card {
+    @include component();
+    display: inline-flex;
+    align-items: flex-start;
+    flex-wrap: wrap;
+    border-radius: $borderRadiusDefault;
+    border: $cardBorderColor solid $cardBorderSize;
+    overflow: hidden;
```
aju: I think we could do it on holes - would make its more flexible?

src/components/card/CardHole.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+export const COLOR = {
```
matikrk: CARD_HOLE_COLOR

src/components/card/Card.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export const PADDING = {
```
matikrk: CARD_PADDING

src/components/card/_card.scss
burakukula: @aju this is necessary here, as holes have background color, and parent have border-radius?

src/components/card/Card.jsx
burakukula: 👍

src/components/card/CardHole.jsx
burakukula: 👍

src/components/card/Card.jsx
burakukula: 👍

src/components/card/_card.scss
burakukula: @aju @matikrk , we have the same solution for Box I think. @kdzwinel suggest that it would be better to keep modifiers on parent (I assume every hole should have the same padding, we should avoid different modifiers for holes for this component, I think)

src/components/card/Card.jsx
burakukula: @aju btw, we name it same way in box. should I change it there as well?

burakukula:
@aju @matikrk please have a look now?

src/components/card/Card.spec.jsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import Card, {CARD_PADDING} from './Card';
+import CardHole, {CARD_HOLE_COLOR} from './CardHole';
+import {shallow} from 'enzyme';
+
+describe('ActionList', () => {
```
matzimowski: 'Card()'

PR #1391, opened 2018-02-05
Added href to IconAsButton

rkty13:
IconAsButton will render as a `<a>` when href is passed in as props. Also added `role="button"` for accessibility.

### DIFF
--- a/src/components/icon-as-button/IconAsButton.jsx
+++ b/src/components/icon-as-button/IconAsButton.jsx
@@ -32,14 +32,20 @@ const IconAsButton = ({
     content = children;
   }
 
+  let renderType = 'button';
+
+  if (props.href) {
+    renderType = 'a';
+  }
+
   return (
-    <button {...props} className={buttonClass}>
+    <renderType {...props} className={buttonClass}>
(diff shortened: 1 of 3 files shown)


src/components/icon-as-button/IconAsButton.jsx
```diff
@@ -53,6 +59,7 @@ IconAsButton.propTypes = {
   transparent: PropTypes.bool,
   active: PropTypes.bool,
   overlay: PropTypes.node,
+  href: PropTypes.string,
```
aju: What about interactive? Do we need anything to display this property there? :)

matikrk: good point. We need additional input in interactive.

kdzwinel: Aju is right, you should add this prop here: https://github.com/brainly/style-guide/blob/43292750917df41672d562f25d360df7053a4176/src/components/icon-as-button/pages/icon-as-button-interactive.jsx#L36 so that it's visible in the interactive part of the doc - https://styleguide.brainly.com/130.4.0/docs/interactive.html

PR #1403, opened 2018-02-15
Fix textarea value PropType

tomasz-sodzawiczny:
`PropTypes.string` was only valid for the native `<textarea />`, but not neccessarily for other node `Type`s.


kdzwinel:
@tomasz-sodzawiczny ping

Issue #1404, opened 2018-02-16
Refactor actions-list prop toTop


matzimowski:
Add `toTop` prop to `ALIGNMENT` const and use with `align`

PR #1405, opened 2018-02-16
Add ALIGNMENT baseline to ActionList

matzimowski:
screen shot 2018-02-16 at 09 30 49

### DIFF
--- a/src/components/action-list/ActionList.jsx
+++ b/src/components/action-list/ActionList.jsx
@@ -8,9 +8,14 @@ export const DIRECTION = {
   SPACE_BETWEEN: 'space-between'
 };
 
-const ActionList = ({children, toTop, direction, noWrap, className, ...props}) => {
+export const ALIGNMENT = {
+  BASELINE: 'baseline'
+};
+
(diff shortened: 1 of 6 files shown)


src/components/action-list/action-list.scss
```diff
@@ -52,6 +52,10 @@
     }
   }
 
+  &--baseline {
```
aju: align-baseline ?

src/components/action-list/ActionList.jsx
```diff
@@ -8,9 +8,14 @@ export const DIRECTION = {
   SPACE_BETWEEN: 'space-between'
 };
 
-const ActionList = ({children, toTop, direction, noWrap, className, ...props}) => {
+export const ALIGNMENT = {
+  BASELINE: 'baseline'
```
matikrk: here should be also      `align-items: flex-start` and others. So the best way will be refactor component and delete `toTop` prop. Because you can't use baseline and toTop in same time

matzimowski:
already created #1404

Issue #1414, opened 2018-03-16
Badge should accept object (other components)


vergilius:
Right now `<Badge />` wants only string as children, please update propTypes to reflect that.

vergilius:
This generic badge, should not be imported and used directly. Instead, we should build basing on it other components, like `TextBadge`, `IconBadge` or others

PR #1415, opened 2018-03-16
Reorganize Badges

vergilius:
Closes https://github.com/brainly/style-guide/issues/1414

It's a breaking change (in terms structure of components) that allows more generic children components to be passed to `<Badge />`.

For backwards compatibility I've kept `<TextBadge />` - it's widely used, having it does allow to keep color sync, and no need to import `<Text />`

Doesn't change anything related to Badges look & feel:
screen shot 2018-03-16 at 12 54 00
screen shot 2018-03-16 at 12 54 12


src/components/badges/Badge.jsx
```diff
@@ -1,49 +1,54 @@
...
-  MUSTARD: {BADGE: 'mustard', TEXT: TEXT_COLOR.LIGHT},
-  MINT: {BADGE: 'mint', TEXT: TEXT_COLOR.LIGHT},
-  MINT_SECONDARY: {BADGE: 'mint-secondary', TEXT: TEXT_COLOR.LIGHT},
-  BLUE_SECONDARY: {BADGE: 'blue-secondary', TEXT: TEXT_COLOR.LIGHT},
-  BLUE: {BADGE: 'blue', TEXT: TEXT_COLOR.LIGHT},
-  GRAY_SECONDARY: {BADGE: 'gray-secondary', TEXT: TEXT_COLOR.LIGHT},
-  MINT_SECONDARY_LIGHT: {BADGE: 'mint-secondary-light', TEXT: TEXT_COLOR.MINT},
-  PEACH_SECONDARY_LIGHT: {BADGE: 'peach-secondary-light', TEXT: TEXT_COLOR.PEACH},
-  BLUE_SECONDARY_LIGHT: {BADGE: 'blue-secondary-light', TEXT: TEXT_COLOR.BLUE}
+export const BADGE_COLOR = {
```
maciekbb: 👍 I would keep the name `COLOR` though

src/components/badges/Badge.jsx
```diff
@@ -1,49 +1,54 @@
...
   );
 };
 
 Badge.propTypes = {
-  color: PropTypes.oneOf(Object.values(COLOR)),
-  size: PropTypes.oneOf(Object.values(SIZE)),
-  children: PropTypes.string.isRequired,
+  color: PropTypes.oneOf(Object.values(BADGE_COLOR)),
+  size: PropTypes.oneOf(Object.values(BADGE_SIZE)),
+  children: PropTypes.any.isRequired,
```
tomasz-sodzawiczny: `PropTypes.node.isRequired`

src/components/badges/TextBadge.jsx
```diff
@@ -0,0 +1,48 @@
...
+  const textSize = size === BADGE_SIZE.LARGE ? TEXT_SIZE.NORMAL : TEXT_SIZE.XSMALL;
+
+  return (
+    <Badge {...props} color={color.BADGE} size={size} rounded={rounded} withAnimation={withAnimation}>
+      <Text size={textSize} weight={TEXT_WEIGHT.BOLD} color={color.TEXT}>{children}</Text>
+    </Badge>
+  );
+};
+
+TextBadge.propTypes = {
```
maciekbb: We should think about switching to flow :)

src/components/badges/TextBadge.spec.jsx
```diff
@@ -0,0 +1,94 @@
...
+      <TextBadge>Test</TextBadge>
+    );
+    const text = badge.find(Text);
+
+    expect(text.props().size).toEqual(TEXT_SIZE.XSMALL);
+    expect(badge.hasClass('sg-badge--small')).toEqual(false);
+    expect(badge.hasClass('sg-badge--large')).toEqual(false);
+  });
+
+  it('small size', () => {
```
tomasz-sodzawiczny: *"it small size"* 🤣

vergilius: xD

src/components/badges/TextBadge.jsx
tomasz-sodzawiczny: Ooor at least creating a lib def :)

src/components/badges/Badge.jsx
tomasz-sodzawiczny: @maciekbb we prefix all constants - otherwise we always do 
```js
import {COLOR as BADGE_COLOR} from ...
```

vergilius: I changed it, as in most cases we have to make aliases (this allows us to have unique consts imported)

vergilius:
@kdzwinel I've moved `Badge` to `subcomponents` directory, afaik we don't have any tools to prevent direct import

PR #1421, opened 2018-03-30
Update frontend-tools-configs

matzimowski:
🏇


matikrk:
+1000 line, stats will go up 😄

PR #1423, opened 2018-04-03
WIP - Adding three columns content option

burakukula:
screen shot 2018-04-03 at 11 57 33

We need it here:

screen shot 2018-04-03 at 11 40 30

### DIFF
--- a/src/components/layout/Layout.jsx
+++ b/src/components/layout/Layout.jsx
@@ -2,8 +2,20 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-const Layout = ({children, header, footer, reversedOrder, noMaxWidth, noMarginTop, fullPage, className, ...props}) => {
-  const layoutClass = classNames('sg-layout', className);
+const Layout = ({
+  children,
+  header,
+  footer,
+  reversedOrder,
+  threeColumns,
+  noMaxWidth,
+  noMarginTop,
+  fullPage,
+  className,
+  ...props}) => {
+  const layoutClass = classNames('sg-layout', {
+    'sg-layout--three-columns': threeColumns
+  });
   const layoutContainerClass = classNames('sg-layout__container', {
     'sg-layout__container--reversed-order': reversedOrder,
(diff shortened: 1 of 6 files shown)


src/components/layout/iframe-pages/layout_three_columns.jsx
```diff
@@ -0,0 +1,61 @@
...
+      <style>{'body {background-color: #f0f3f5;}'}</style>
+    </head>
+    <body>
+      <Layout threeColumns header={header} footer={footer}>
+        <LayoutLeftContent>
+          <LayoutBox>
+            <Text>
+              Left content
+            </Text>
+          </LayoutBox>
```
tomasz-sodzawiczny: Remove the padding and background, it's not the way we'll use it, it's also confusing when you look at the example (the right column explicitaly says *"Aside content has no padding and no background"*).

burakukula: @tomasz-sodzawiczny ups. On my way. thx

src/components/layout/Layout.jsx
```diff
@@ -32,6 +44,7 @@ Layout.propTypes = {
   header: PropTypes.node,
   footer: PropTypes.node,
   reversedOrder: PropTypes.bool,
+  threeColumns: PropTypes.bool,
```
tomasz-sodzawiczny: How do we pass the third column in?

src/components/layout/Layout.spec.jsx
```diff
@@ -130,3 +141,15 @@ describe('LayoutBox', () => {
     expect(layoutBox.hasClass('sg-layout__box')).toEqual(true);
   });
 });
+
+describe('LayoutLeftContent', () => {
```
tomasz-sodzawiczny: Do we always render the contents of the left-most column? What happens on small screens?
```
L C R
```
v
```
L
C
R
```
?

burakukula: @tomasz-sodzawiczny we should display:

```
L
C
R
```

PR #1426, opened 2018-04-04
Action list hole equal width

burakukula:
screen shot 2018-04-04 at 17 15 59

And we need it here:

screen shot 2018-04-04 at 17 17 23

### DIFF
--- a/src/components/action-list/action-list.scss
+++ b/src/components/action-list/action-list.scss
@@ -51,6 +51,12 @@
       align-self: stretch;
     }
 
+    &--equal-width {
+      flex-basis: 0;
+      flex-grow: 1;
+      flex-shrink: 1;
+    }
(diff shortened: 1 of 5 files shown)


src/components/action-list/action-list.scss
```diff
@@ -51,6 +51,12 @@
       align-self: stretch;
     }
 
+    &--equal-width {
+      flex-basis: 0;
```
vergilius: afaikt this has issues in older safari and IE, can you confirm/deny?

burakukula: @vergilius thx, I will check it.

burakukula: @vergilius IE11 and older Safari seem's to be working fine.

Issue #1416, opened 2018-03-19
Add hide overflow option for actions list hole


aju:
```
.sg-actions-list > .sg-actions-list__hole--hide-overflow {
  overflow: hidden;
}
```
why:
<img width="436" alt="screen shot 2018-03-19 at 13 05 01" src="https://user-images.githubusercontent.com/316313/37594682-38ac53aa-2b76-11e8-8c7e-a6fa809d487e.png">
to make possible middle TOC section

kdzwinel:
maybe some kind of overflow helper we can wrap the text in? 🤔

matzimowski:
not valid anymore

PR #1432, opened 2018-04-11
extend action list hole

aju:
screen shot 2018-04-11 at 15 26 28


aju:
closes https://github.com/brainly/style-guide/issues/1416

Issue #1413, opened 2018-03-12
Add `--large` modifier for rating


vergilius:
See for details: https://github.com/brainly/brainly-frontend/pull/563

PR #1433, opened 2018-04-13
Upgrade rating component

vergilius:
- added `LARGE` size, change way the props are setting size,
- removed `SMALL` size,
- added more power: component allows to listen on star events - example usage:
mar-16-2018 09-24-50
- counter visibility now done in CSS,
- closes https://github.com/brainly/style-guide/issues/1413
- closes https://github.com/brainly/style-guide/issues/1435

**Screenshot**
screen shot 2018-04-17 at 09 55 12

**Styles change for visual comparison:**
screen shot 2018-04-16 at 15 55 12
screen shot 2018-04-16 at 15 55 31

### DIFF
--- a/src/components/rating/Rating.jsx
+++ b/src/components/rating/Rating.jsx
@@ -83,11 +83,11 @@ class Rating extends Component {
 
   render() {
     const {metricSize, rate, size = RATING_SIZE.NORMAL,
-      active, toBottom, className, counterText, activeText} = this.props;
+      active, className, counterText, activeText, altLabels} = this.props;
     const ratingClass = classnames('sg-rate-box', {
       'sg-rate-box--large': size === RATING_SIZE.LARGE,
       'sg-rate-box--small': size === RATING_SIZE.SMALL,
-      'sg-rate-box--to-bottom': toBottom,
+      'sg-rate-box--with-alt-labels': altLabels,
       'sg-rate-box--active': active
     }, className);
 
@@ -127,7 +127,7 @@ Rating.propTypes = {
   rate: PropTypes.number,
   metricSize: PropTypes.number,
   active: PropTypes.bool,
-  toBottom: PropTypes.bool,
+  altLabels: PropTypes.bool,
   onChange: PropTypes.func,
   onStarMouseEnter: PropTypes.func,
   onMouseLeave: PropTypes.func,
--- a/src/components/rating/Rating.jsx
+++ b/src/components/rating/Rating.jsx
@@ -5,7 +5,6 @@ import Star from './subcomponents/Star';
 import RateCounter from './subcomponents/RateCounter';
 
 export const RATING_SIZE = {
-  SMALL: 14,
   NORMAL: 16,
   LARGE: 24
 };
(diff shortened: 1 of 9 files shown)


src/components/rating/Rating.spec.jsx
```diff
@@ -95,12 +95,6 @@ describe('rating', () => {
 
     stars.at(1).simulate('click');
     expect(onChange.mock.calls).toHaveLength(2);
-
-    const lastRatedStarIndex = rate - 1;
```
vergilius: this caused error:
when current rate was === 4, then clicking on 4th star didn't call update ;(

src/components/rating/subcomponents/RateCounterItem.jsx
```diff
@@ -1,27 +1,19 @@
 import React, {Fragment} from 'react';
 import PropTypes from 'prop-types';
-import classnames from 'classnames';
 
-const RateCounterItem = ({text, hidden = false}) => {
-  const dynamicItemClassName = classnames('sg-rate-box__counter-item-dynamic', {
```
vergilius: this is now done in CSS

src/components/rating/Rating.jsx
```diff
@@ -91,30 +104,33 @@ class Rating extends Component {
...
         </div>
-        <div className="sg-rate-box__stars-container" onMouseEnter={this.onMouseEnter} onMouseLeave={this.onMouseLeave}>
+        <div className="sg-rate-box__stars-container" onMouseLeave={this.onMouseLeave}>
           <div className="sg-rate-box__filled-stars" style={{width: `${100 * rate / metricSize}%`}}>
             {starsProps.map(props => <Star key={props.key} {...props} />)}
           </div>
           <div className="sg-rate-box__background-stars">
-            {starsProps.map(props => <Star key={props.key} {...props} />)}
+            {starsProps.map(props =>
+              <Star key={props.key} onMouseEnter={this.starsMouseEnterFunctions[props.key]} {...props} />)}
```
kdzwinel: why onEnter is here and onLeave is on the container?

kdzwinel:
looks good, please update static docs and interactive page

src/components/rating/Rating.jsx
vergilius: There's no real need for listening on Star 'mouseleave' event, when you are switching from star to star it's easier to check for update from onEnter than on both leave and enter listeners.

On the other side, its important to know when you are actually leaving whole component.
Generally these two behaviours allow to cover all current use cases:
- hovering stars - entering star equals to entering component,
- hovering stars - going from star 1 to star 2 (no onLeave ==> less events triggered)
- leaving component

kdzwinel: hm, makes sense but maybe we should be more explicit? e.g. `onLeaveComponent`, `onEnterStar`? we may need `onEnterComponent` and `onLeaveStar` in the future.

BTW `onLeaveComponent` isn't the same as `<Rating onMouseLeave=` ?

src/components/rating/pages/rating-interactive.jsx
```diff
@@ -9,7 +9,11 @@ const Ratings = () => {
       values: Boolean
     },
     {
-      name: 'small',
+      name: 'size',
```
vergilius: @kdzwinel react interactive

src/components/rating/pages/rating.jsx
```diff
@@ -13,12 +12,32 @@ const ratings = () => (
       <Rating rate={3} active counter={34} />
     </DocsBlock>
 
+    <DocsBlock info="Active with counters">
```
vergilius: @kdzwinel static docs

src/components/rating/_rating.scss
```diff
@@ -18,8 +20,20 @@ $includeHtml: false !default;
     display: flex;
     align-items: center;
     height: $rateHeight;
+    overflow: visible;
```
kdzwinel: that's the default value?

vergilius: `component` mixin uses `overflow: hidden` as default

PR #1434, opened 2018-04-16
Text as container

aju:


vergilius:
I kinda wonder if this really should have container behaviour, you could wrap it with container

PR #1451, opened 2018-06-12
change textbit

ElaLukaszewicz:
TextBit changes: font-size form 24px to 25px, letter spacing to 0.
now:
image
was:
image

### DIFF
--- a/src/components/text/_text-bits.scss
+++ b/src/components/text/_text-bits.scss
@@ -46,9 +46,9 @@ $includeHtml: false !default;
 
     &--small {
       @include uppercaseText(0);
-      font-size: rhythm(25 / 24);
+      font-size: fontSize(headline);
       line-height: rhythm(1);
-
+      
       @include sgBreakpoint(medium-up) {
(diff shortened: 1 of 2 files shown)


src/components/text/_text-bits.scss
```diff
@@ -45,7 +45,8 @@ $includeHtml: false !default;
     }
 
     &--small {
-      font-size: rhythm(1);
+      @include uppercaseText(0);
+      font-size: rhythm(25 / 24);
```
vergilius: I think font sizes should be taken from: https://github.com/brainly/style-guide/blob/master/src/sass/_mixins.scss#L41

aju: 👍

ElaLukaszewicz: i;m not sure how, because there is no variable in fontSizes which has 25px - i was asked to make it exactly that big

ElaLukaszewicz: image

PR #1452, opened 2018-06-13
Vertical action list

anna-wro:
screenshot 2018-06-13 14 41 45

### DIFF
--- a/src/components/action-list/action-list.scss
+++ b/src/components/action-list/action-list.scss
@@ -110,9 +110,9 @@
   &--vertical {
     flex-direction: column;
     align-items: flex-start;
- 
+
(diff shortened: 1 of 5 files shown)


src/components/action-list/action-list.scss
```diff
@@ -106,4 +106,13 @@
   &--no-wrap {
     flex-wrap: nowrap;
   }
+
+  &--vertical {
+    flex-direction: column;
+    align-items: flex-start;
+
+    > .sg-actions-list__hole:not(:last-child) {
```
maciekbb: Now I'm wondering if we should really introduce this rule, ona can always add spacing directly to list hole.

maciekbb:
We decided to use content box for vertical positioning

PR #1453, opened 2018-06-14
consistent messages for fallback on feed 

ElaLukaszewicz:
Design for consistent messages for users in case of: disabling some feature in Brainly, "your search include questions with Verified Answers", info about patron of a subject etc.
It is actually a Box with new modifiers: 'message' and 'onClose' - function on 'X' closing a message.
'Message' excludes shadow and different padding - to stay consistent.

Design:
image

Implementation:
image

### DIFF
--- a/src/components/box/Box.jsx
+++ b/src/components/box/Box.jsx
@@ -36,7 +36,7 @@ const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeig
     'sg-box--no-min-height': noMinHeight,
     'sg-box--with-shadow': shadow && !message,
     'sg-box--message': message,
-    'sg-box--with-onclose': onClose
+    'sg-box--with-close': onClose
 
   }, className);
 
--- a/src/components/box/Box.jsx
+++ b/src/components/box/Box.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Icon, {TYPE as iconTypes, ICON_COLOR} from '../icons/Icon';
 
 export const COLOR = {
   BLUE: 'blue',
@@ -10,6 +11,7 @@ export const COLOR = {
   MINT_SECONDARY: 'mint-secondary',
   MINT_SECONDARY_LIGHT: 'mint-secondary-light',
   NAVYBLUE_SECONDARY: 'navyblue-secondary',
+  BLUE_SECONDARY: 'blue-secondary',
   BLUE_SECONDARY_LIGHT: 'blue-secondary-light',
   GRAY_SECONDARY_LIGHT: 'gray-secondary-lightest',
   PEACH: 'peach'
@@ -23,15 +25,19 @@ export const PADDING = {
   LARGE: 'large-padding'
 };
 
-const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeight, shadow, className, ...props}) => {
+const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeight, shadow,
+  message, onClose, className, ...props}) => {
   const boxClass = classNames('sg-box', {
     [`sg-box--${color}`]: color,
     'sg-box--no-border': !border,
     'sg-box--full': full,
-    [`sg-box--${padding}`]: padding,
+    [`sg-box--${padding}`]: padding && !message,
     'sg-box--image-wrapper': imgSrc,
     'sg-box--no-min-height': noMinHeight,
-    'sg-box--with-shadow': shadow
+    'sg-box--with-shadow': shadow && !message,
+    'sg-box--message': message,
+    'sg-box--with-onClose': onClose
+
   }, className);
 
   let content;
@@ -44,6 +50,11 @@ const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeig
 
   return (
     <div {...props} className={boxClass}>
+      {onClose ?
+        <div className="sg-box--message__close" onClick={onClose}>
+          <Icon type={iconTypes.X} color={ICON_COLOR.DARK} size={14} />
+        </div> : null
+      }
       {content}
     </div>
   );
@@ -58,6 +69,8 @@ Box.propTypes = {
   padding: PropTypes.oneOf(Object.values(PADDING)),
   imgSrc: PropTypes.string,
   shadow: PropTypes.bool,
+  message: PropTypes.bool,
+  onClose: PropTypes.bool,
   className: PropTypes.string
 };
 
--- a/src/components/box/Box.jsx
+++ b/src/components/box/Box.jsx
@@ -36,7 +36,7 @@ const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeig
     'sg-box--no-min-height': noMinHeight,
     'sg-box--with-shadow': shadow && !message,
     'sg-box--message': message,
-    'sg-box--with-onClose': onClose
+    'sg-box--with-onclose': onClose
 
   }, className);
 
@@ -51,7 +51,7 @@ const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeig
   return (
     <div {...props} className={boxClass}>
       {onClose ?
-        <div className="sg-box--message__close" onClick={onClose}>
+        <div className="sg-box--message__close" onClose={onClose}>
           <Icon type={iconTypes.X} color={ICON_COLOR.DARK} size={14} />
         </div> : null
       }
@@ -70,7 +70,7 @@ Box.propTypes = {
   imgSrc: PropTypes.string,
   shadow: PropTypes.bool,
   message: PropTypes.bool,
-  onClose: PropTypes.bool,
+  onClose: PropTypes.func,
   className: PropTypes.string
 };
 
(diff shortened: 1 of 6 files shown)


ElaLukaszewicz:
BTW: Font-size is actually the same as in invissionapp

src/components/box/Box.jsx
```diff
@@ -44,6 +50,11 @@ const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeig
 
   return (
     <div {...props} className={boxClass}>
+      {onClose ?
+        <div className="sg-box--message__close" onClose={onClose}>
```
aju: `sg-box--message__close` is a wrong bem selector :) I think `sg-box__close` is good :)

src/components/box/_box.scss
```diff
@@ -123,6 +127,24 @@ $includeHtml: false !default;
       align-items: center;
       justify-content: center;
     }
+
+    &--message {
+      padding: gutter(1);
+    }
+
+    &--with-onclose {
```
aju: just `--with-close`

src/components/box/Box.jsx
ElaLukaszewicz: i agree 👍

src/components/box/Box.jsx
```diff
@@ -44,6 +50,11 @@ const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeig
 
   return (
     <div {...props} className={boxClass}>
+      {onClose ?
```
maciekbb: IMO onClose is not part of the `Box`, let's talk on weekly meeting

src/components/box/_box.scss
ElaLukaszewicz: oki

ElaLukaszewicz:
image

ElaLukaszewicz:
image

src/components/message-box/_message_box.scss
```diff
@@ -0,0 +1,24 @@
...
+      position: absolute;
+      right: 0;
+      top: 0;
+      padding: gutter(1 / 2);
+      cursor: pointer;
+      z-index: 1;
+   }
+  }
+
+  .sg-box {
```
vergilius: This should be part of box styles

src/components/message-box/MessageBox.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+export {COLOR};
+
+const MessageBox = ({children, onClose, ...props}) => {
+  const boxClass = classNames({
+    'sg-box--with-close': onClose
+  });
+
+  return (
+    <div className={'sg-message-box'}>
```
maciekbb: I think `className` and `styles` attributes should we passed to this div instead of `Box`

src/docs/navigation.js
```diff
@@ -199,6 +200,10 @@ const navigation = [
         'location': 'box/box',
         component: box
       },
+      {
+        'name': 'Message Box',
+        component: messageBox
```
maciekbb: inconsistency

src/components/message-box/pages/message-box-interactive.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import MessageBox, {COLOR} from '../MessageBox';
+import DocsActiveBlock from 'components/DocsActiveBlock';
+
+const closeCallback = () => undefined;
+
+const MessageBoxes = () => {
```
vergilius: I believe we always used singular nouns as a name

src/components/message-box/_message_box.scss
ElaLukaszewicz: i thought of that but wasn't sure - changing now

src/components/message-box/_message_box.scss
```diff
@@ -0,0 +1,24 @@
+$includeHtml: false !default;
+
+@if ($includeHtml) {
+  .sg-message-box {
+    position: relative;
+    display: inline-block;
```
maciekbb: Why `inline-block`?

src/components/message-box/pages/message-box-interactive.jsx
ElaLukaszewicz: actually Box is also in pages as plural

src/components/message-box/_message_box.scss
aju: yeah 👍

src/components/message-box/_message_box.scss
aju: :+

src/components/message-box/pages/message-box-interactive.jsx
ElaLukaszewicz: image

src/components/message-box/MessageBox.jsx
```diff
@@ -0,0 +1,34 @@
...
+      {onClose ?
+        <div className="sg-message-box__close" onClose={onClose}>
+          <Icon type={iconTypes.X} color={ICON_COLOR.DARK} size={14} />
+        </div> : null
+      }
+      <Box {...props} className={boxClass} >
```
aju: shouldn't we have specific padding for Box ?

src/components/message-box/_message_box.scss
maciekbb: @vergilius but this is specific style to be used just by message box

src/components/message-box/MessageBox.jsx
ElaLukaszewicz: default one is ok according to designers

ElaLukaszewicz: i can set it only for small or large (there is no normal) so i would mess it up only

src/components/message-box/_message_box.scss
maciekbb: This is for it not to fill entire available space, this can be achieved by passing `fullWidth` prop

src/components/message-box/MessageBox.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import Box, {COLOR} from '../box/Box';
+import Icon, {TYPE as iconTypes, ICON_COLOR} from '../icons/Icon';
+
+export {COLOR};
```
matikrk: export as MESSAGE_BOX_COLOR or BOX_COLOR

src/components/message-box/MessageBox.spec.jsx
```diff
@@ -0,0 +1,63 @@
...
+    const messageBox = shallow(
+      <MessageBox color={COLOR.BLUE_SECONDARY} />
+    );
+
+    expect(messageBox.find(Box).props().color).toEqual('blue-secondary');
+  });
+
+  it('check if close button exists', () => {
+    const mockCallback = jest.fn();
+    const messageBox = mount(<MessageBox onClose={mockCallback} />);
```
matikrk: do we need mount here?

src/components/message-box/MessageBox.spec.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+    expect(messageBox.find('.sg-message-box__close')).toHaveLength(1);
+  });
+
+  it('check when no close button', () => {
+    const messageBox = mount(<MessageBox />);
```
matikrk: shallow will be enough here

ElaLukaszewicz:
image

src/components/message-box/MessageBox.spec.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+    expect(messageBox.find('.sg-message-box__close')).toHaveLength(0);
+  });
+
+  it('check if padding-right is bigger when close button exists', () => {
+    const mockCallback = jest.fn();
+    const messageBox = mount(<MessageBox onClose={mockCallback} />);
+
+    expect(messageBox.find(Box).hasClass('sg-message-box--closable')).toEqual(true);
```
matikrk: checking like this is not only unit test, coz u test 2 components.

You could use shallow render and check props instead of hasClass

src/components/message-box/MessageBox.jsx
```diff
@@ -0,0 +1,31 @@
...
+import Box, {COLOR} from '../box/Box';
+import Icon, {TYPE as iconTypes, ICON_COLOR} from '../icons/Icon';
+
+const MessageBox = ({children, onClose, ...props}) => {
+  const boxClass = classNames('sg-message-box', {
+    'sg-message-box--closable': onClose
+  });
+
+  return (
+    <Box {...props} className={boxClass} >
```
aju: Oh! this was good! we don't want to mix classes and overrides ! CC @maciekbb

src/components/message-box/_message_box.scss
```diff
@@ -0,0 +1,19 @@
...
+    &__close {
+      position: absolute;
+      right: 0;
+      top: 0;
+      padding: gutter(1 / 2);
+      cursor: pointer;
+      z-index: 1;
+    }
+
+    &--closable {
```
aju: I think this should be a property of box like padding-right-large

src/components/message-box/MessageBox.jsx
burakukula: @maciekbb maybe Aju is right, maybe we should keep specific styles for Box in Box and here just display this specific Box, with proper padding?

PR #1461, opened 2018-07-05
add gray prop

ElaLukaszewicz:
Add gray Avatar for the purpose of new placeholders for ranking, feed, etc.
image
image


burakukula:
@ElaLukaszewicz @aju let's keep it as PlaceholderRounded in the project, seems like we could use it also for icons, so it is more generic I think.

PR #1468, opened 2018-07-18
Add option to Bubble to disable box-shadow

wasilakkf:
screen shot 2018-07-18 at 15 42 09

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "136.6.0",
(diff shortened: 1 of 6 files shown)


aju:
@wasilakkf @maciekbb is this confirmed with Patrycja Rozmus?

wasilakkf:
I think it is. Karol and Łukasz said that in this case with deduplication design we want a Bubble without shadow - like on this mockup:
screen shot 2018-07-19 at 08 57 09

PR #1469, opened 2018-07-20
Reverse order layout

ElaLukaszewicz:
Before:
image

After:
image


aju:
@ElaLukaszewicz what pages will be affected? and how will they look like?

PR #1473, opened 2018-07-27
export colors as a hex values

bartoszlorek:
and now we will be able to import like that:
```javascript
import hex from 'style-guide/src/components/colors/hex';

hex.mintPrimary  // #53cf92
hex.mustardPrimary  // #fec83c
```

### DIFF
--- a/src/components/colors/hex.js
+++ b/src/components/colors/hex.js
@@ -1,0 +1,11 @@
+import colors from './colors';
+
+const hex = {};
+
+Object.keys(colors).forEach(groupName => {
+  colors[groupName].forEach(color => {
+    hex[color.variable.slice(1)] = `#${color.hex}`;
+  });
+});
+
+export default hex;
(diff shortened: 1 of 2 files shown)


src/components/colors/hex.js
```diff
@@ -0,0 +1,11 @@
+import colors from './colors';
+
+const hex = {};
+
+Object.keys(colors).forEach(groupName => {
```
vergilius: I would export is as a function

vergilius:
👍

src/components/colors/hex.js
bartoszlorek: iife?

vergilius: nah, more like colors module could export function `getHexColors()` etc

but it's random thought, it's okay this way too

vergilius: It's a bit misleading, as colors already have `hex` property

PR #1484, opened 2018-08-24
Add classname support to ListItem && ListItemIcon

matikrk:
*  ...rest props
* add tests


src/components/list/List.spec.jsx
```diff
@@ -26,4 +26,13 @@ describe('<List>', () => {
 
     expect(list.hasClass('sg-list--spaced-elements')).toEqual(false);
   });
+
+  it('renders additional classes', () => {
+    const list = shallow(
+      <List className="m4l" />
```
kdzwinel: I approve this class name 👌

PR #1485, opened 2018-09-02
Bit update

burakukula:
**WILL BE MERGED 7.09.2018**
This is the first part of the update of text, that will be made in next days.
 - new config for font size and line-height - in case of BIT they are equal, so only one cofing needed.
 - custom function to getSizes
 - I will rename all the "context" like namespaces - "alt" "warning" and change them into color indicator modifiers "--blue-secondary" and "--peach-primary"
 - I am removing BIT behavior and "not-responsive" modifier. They should have the same simple meaning as other components. In projects I will replace each usage of BIT with according 2 components.

typography-handoff-map

### DIFF
--- a/src/sass/_config.scss
+++ b/src/sass/_config.scss
@@ -60,8 +60,12 @@ $fontWeightBold: 700;
 $fontWeightBlack: 900;
 
 // Vertical rhythm
-$baseFont: 16px;
+$baseFont: 18px;
 $baseline: 1.5rem;
+
+
+$newFontSize: 18px;
+$newBaseLine: 27px;
 
 // Typo
 $fontSizes: (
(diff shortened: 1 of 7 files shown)


src/sass/_config.scss
```diff
@@ -80,6 +80,13 @@ $fontSizes: (
   header: 3rem
 );
 
+$bitSizes: (
```
aju: I would keep it in bit scss ;)

src/sass/_mixins.scss
```diff
@@ -42,6 +42,11 @@
   @return map-get($fontSizes, $fontsizeKey);
 }
 
+// $bitSizes one config for font size and line-height
```
aju: I would keep it in bit scss ;)

src/sass/_config.scss
burakukula: @aju why? I think would be better to keep config for all types of texts in one place?

src/sass/_mixins.scss
burakukula: @aju ok, I will move config and function into BIT scss. You have a point, we will keep all the styles related to the component in it's scss.

PR #1486, opened 2018-09-04
Headline component

burakukula:
**Will be merged as soon as implementation of BIT's is finished**

The second part of refactoring of texts contains:

 - new Headline component. It will replace HeaderPrimary and HeaderSecondary. After refactoring in projects and making sure we have a new component in place on production, I will remove Headers from style guide.

 - We will support extra-bold for each size of a headline.
 - `--light` modifier is replaced with `--white`
 - a new component will not have any responsive behavior. I will replace existing headers with 2xheadlines if needed - for small and large screens.

screen shot 2018-09-03 at 16 54 29

You can have a look on the implementation map here:
typography-handoff-map

### DIFF
--- a/src/components/text/_headlines.scss
+++ b/src/components/text/_headlines.scss
@@ -1,67 +1,74 @@
 $includeHtml: false !default;
 
-$headlineFontSizes: (
-  xxlarge: 53px,
-  xlarge: 39px,
-  large: 28px,
-  normal: 21px,
-  small: 18px,
-  xsmall: 14px
+$headlineSizes: (
+  xxlarge: (
+    fontSize: 53px,
+    lineHeight: 60px
+  ),
+  xlarge: (
+    fontSize: 39px,
+    lineHeight: 44px
+  ),
+  large: (
+    fontSize: 28px,
+    lineHeight: 32px
+  ),
+  normal: (
+    fontSize: 21px,
+    lineHeight: 24px
+  ),
+  small: (
+    fontSize: 18px,
+    lineHeight: 20px
+  ),
+  xsmall: (
+    fontSize: 14px,
+    lineHeight: 15px
+  )
 );
 
-$headlineLineHeight: (
-  xxlarge: 60px,
-  xlarge: 44px,
-  large: 32px,
-  normal: 24px,
-  small: 20px,
-  xsmall: 15px
-);
-
-@function getHeadlineFontSize($fontsizeKey) {
-  @return map-get($headlineFontSizes, $fontsizeKey);
+@function getSizeFromMap($map, $keys...) {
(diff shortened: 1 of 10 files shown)


src/components/text/_headlines.scss
```diff
@@ -0,0 +1,87 @@
...
+$headlineFontSizes: (
+  xxlarge: 53px,
+  xlarge: 39px,
+  large: 28px,
+  normal: 21px,
+  small: 18px,
+  xsmall: 14px
+);
+
+$headlineLineHeight: (
```
aju: can we merge it somehow with headlineFontSizes ?

src/components/text/_headlines.scss
```diff
@@ -0,0 +1,87 @@
...
+    color: $black;
+    font-size: getHeadlineFontSize(normal);
+    line-height: getHeadlineLineHeight(normal);
+    font-weight: $fontWeightBold;
+    letter-spacing: -1px;
+    max-width: 100%;
+    overflow: visible;
+
+    &--xsmall {
+      font-size: getHeadlineFontSize(xsmall);
```
aju: I think you can create a mixing to set `font-size` and `line-height` instead of repeating it every time

src/components/text/headlineConsts.js
```diff
@@ -0,0 +1,27 @@
+export const HEADLINE_TYPE = {
```
aju: I think we should keep it in one file to make import easier - or move those 2 files to separate directory `headlines` and add index.js that reexports from both of them.

burakukula: @aju yeah, sounds like a good idea. But maybe I will keep it for later? maybe we could use one config for headlines and body text somehow. 
I made a refactor of logic for headlines. If you have a moment, please have a look and tell me what do you think. 🐯

src/components/text/_headlines.scss
```diff
@@ -0,0 +1,94 @@
...
+    fontSize: 18px,
+    lineHeight: 20px
+  ),
+  xsmall: (
+    fontSize: 14px,
+    lineHeight: 15px
+  )
+);
+
+@function getSizeFromMap($map, $keys...) {
```
aju: I think it should be called getHeadLineSizeFromMap ;) -> avoid name conflicts

src/components/text/_headlines.scss
```diff
@@ -0,0 +1,94 @@
...
+  @each $key in $keys {
+    $map: map-get($map, $key);
+  }
+  @return $map;
+}
+
+@mixin typeSizeVariant($size) {
```
aju: this should have headlineTypeSizeVarinat

PR #1491, opened 2018-09-13
Change close button in box

ElaLukaszewicz:
image
Add white close button in Box.
Delete duplicated 'className'


src/components/box/Box.jsx
```diff
@@ -26,7 +26,7 @@ export const PADDING = {
 };
 
 const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeight, shadow, noBorderRadius,
-  onClose, className, ...props}) => {
+  onClose, whiteCloseIcon, className, ...props}) => {
```
vergilius: I don't think we should be that much specific, can we just use closeIconColor with allowed light & dark?

vergilius: and if not, I would suggest to use `light` instead

src/components/box/Box.jsx
```diff
@@ -45,12 +50,11 @@ const Box = ({color, padding, full, children, border = !color, imgSrc, noMinHeig
   } else {
     content = <div className="sg-box__hole">{children}</div>;
   }
-
   return (
     <div {...props} className={boxClass}>
       {onClose ?
         <div className="sg-box__close" onClick={onClose}>
-          <Icon type={iconTypes.X} color={ICON_COLOR.DARK} size={10} />
+          <Icon type={iconTypes.X} color={closeIconColor ? ICON_COLOR[closeIconColor] : ICON_COLOR.DARK} size={10} />
```
vergilius: 🥇

PR #1496, opened 2018-09-21
Flow setup

burakukula:


src/components/text/TextBit.jsx
```diff
@@ -26,7 +27,15 @@ export const TEXT_BIT_COLOR = {
   PEACH_PRIMARY: 'peach-primary'
 };
 
-const TextBit = ({children, type = TEXT_BIT_TYPE.H1, color, size, className, ...props}) => {
+type TextBitPropsType = {
+  children: Object,
```
maciekbb: children should be React.Node?

src/components/text/TextBit.jsx
```diff
@@ -26,7 +27,15 @@ export const TEXT_BIT_COLOR = {
...
+  size: string,
```
maciekbb: some enum here? same for color and type

PR #1500, opened 2018-09-24
Final fix for BIT

burakukula:
Rollback to a previous version where the default value is exported - seem's like more intuitive also in terms of usage. Normal value does not append a modifier to the class `sg-text-bit`, it just points to this default (just `sg-text-bit`)
Example of usage:
`size={screenInfo.isSmallOnly ? TEXT_BIT_SIZE.SMALL : TEXT_BIT_SIZE.NORMAL}`

### DIFF
--- a/src/components/text/TextBit.jsx
+++ b/src/components/text/TextBit.jsx
@@ -45,7 +45,7 @@ const TextBit = ({
   ...props}: TextBitPropsType) => {
   const Type = type;
   const textClass = classNames('sg-text-bit', {
-    [`sg-text-bit--${size}`]: size,
+    [`sg-text-bit--${size}`]: size && size !== TEXT_BIT_SIZE.NORMAL,
     [`sg-text-bit--${color || ''}`]: color
   }, className);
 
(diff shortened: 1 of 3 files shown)


src/components/text/TextBit.jsx
```diff
@@ -44,7 +45,7 @@ const TextBit = ({
   ...props}: TextBitPropsType) => {
   const Type = type;
   const textClass = classNames('sg-text-bit', {
-    [`sg-text-bit--${size || ''}`]: size,
```
maciekbb: What if size is undefined? Shouldn't it be `size && size !== TEXT_BIT_SIZE.NORMA`

src/components/text/TextBit.jsx
```diff
@@ -44,7 +45,7 @@ const TextBit = ({
...
+    [`sg-text-bit--${size || ''}`]: size !== TEXT_BIT_SIZE.NORMAL,
```
rkty13: Why not use:

```
const TextBit = ({ . . . , size = TEXT_BIT_SIZE.NORMAL }) => { . . . }
```

So you don't have to have the conditional `${size || ''}`. By default if no `size` prop is passed, it will be normal.

burakukula:
@maciekbb I added one that checkes if there is no class added when it is default value. I think this will do it.

PR #1501, opened 2018-09-24
Body text update

burakukula:
New body text changes includes:

 - Namespaces changes:
screen shot 2018-09-27 at 14 47 35
 ** we are not supporting font-size 10px any more
 - Link can be created in every variant of Text, it is still a separate component, but support common sizes/colors/weight/ transforms as for text.
 - We remove link unstyled and disabled option - it is not needed to support them. Disabled is btw type of hack to provide link behavior on not link elements. If there will be a need for that, we should create a helper component for that.

Presenation in SG changes a little also:
screen shot 2018-09-27 at 12 52 59

PS. I this PR I keep sg link class to not break twig, I will introduce change in twig in 2nd step
PS2. I do not reexport Link stuff from index.js as I hope to manage to move Link into separate file in the future, so it will have it's own reexport and we would be able to import like follows:
`import Link, {LINK_SIZE} from 'link'`


src/components/bubble/pages/bubble.jsx
```diff
@@ -9,7 +9,8 @@ import ButtonSecondary, {BUTTON_SECONDARY_TYPE} from 'buttons/ButtonSecondary';
 import Text from 'text/Text';
 import Avatar from 'avatar/Avatar';
 import Breadcrumb from 'breadcrumbs/Breadcrumb';
-import Link, {COLOR as LINK_COLOR} from 'text/Link';
+import Link from 'text/Link';
+import {TEXT_COLOR} from '../../text';
```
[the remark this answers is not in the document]
maciekbb: `import Link, {LINK_COLOR} from 'text/Link';`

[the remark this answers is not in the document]
maciekbb: `export {TEXT_COLOR as LINK_COLOR} from './Text'`

src/docs/components/DocsActiveBlockSettings.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Text, {SIZE as TEXT_SIZE, COLOR as TEXT_COLOR} from 'text/Text';
+import Text, {TEXT_SIZE, TEXT_COLOR} from '../../../src/components/text';
```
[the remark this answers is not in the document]
aju: 👍

src/components/text/textConsts.js
```diff
@@ -0,0 +1,53 @@
+export const TEXT_TYPE = Object.freeze({
```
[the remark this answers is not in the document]
aju: 👍

src/components/action-list/pages/action-list-interactive.jsx
```diff
@@ -3,7 +3,8 @@ import ActionList, {DIRECTION, ALIGNMENT} from '../ActionList';
 import ButtonSecondary, {BUTTON_SECONDARY_TYPE} from 'buttons/ButtonSecondary';
 import ButtonPrimary, {BUTTON_PRIMARY_TYPE} from 'buttons/ButtonPrimary';
 import Icon, {ICON_COLOR, TYPE as ICON_TYPE} from 'icons/Icon';
-import Text, {TYPE as TEXT_TYPE, COLOR as TEXT_COLOR} from 'text/Text';
+import Text, {TEXT_TYPE, TEXT_COLOR} from 'text';
```
maciekbb: does it makes sense to have `index.js` in `components/text`? Wouldn't you rather have all the imports consistent like:
`text/Text`
`text/Link`
`text/Headline`

src/components/text/Link.jsx
```diff
@@ -1,61 +1,59 @@
...
-  FINE_PRINT_LIGHT: 'for-fine-print-light'
-};
+export const LINK_COLOR = TEXT_COLOR;
+export const LINK_TYPE = TEXT_TYPE;
+export const LINK_SIZE = TEXT_SIZE;
+export const LINK_WEIGHT = TEXT_WEIGHT;
+export const LINK_TRANSFORM = TEXT_TRANSFORM;
+export const LINK_ALIGN = TEXT_ALIGN;
 
-export const SIZE = {
```
maciekbb: This is not backward-compatible

src/components/action-list/pages/action-list-interactive.jsx
burakukula: @maciekbb maybe your right, I was hoping that maybe we would move components into separate folders eventually. But maybe this would be better to do, when I actually would move them.

src/components/text/Link.jsx
burakukula: It's not, but in many places we anyway need to make import like:
`import Link, {SIZE as LINK_SIZE} from 'style-guide/text/Link`
You think is does make seanse to fix it now?

src/components/action-list/pages/action-list-interactive.jsx
rkty13: +1 to @maciekbb, I think if we want to restructure components into separate folders, we should do them all at once.

PR #1503, opened 2018-10-05
Adding backward compatibility to Text and Link

burakukula:


burakukula:
@maciekbb I think so. I checked locally from branch and it seems to work fine. Some of the old props are just mapped to new scss.

PR #1504, opened 2018-10-09
Alignment in headlines

burakukula:
We need to support alignment also in Headline. As Headlines would tend to appear in pair with body text we need to have the same options for both components.
Flow would be added later on, as soon as body text implementation is done.

### DIFF
--- a/src/components/text/Headline.jsx
+++ b/src/components/text/Headline.jsx
@@ -1,9 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import {HEADLINE_COLOR, HEADLINE_SIZE, HEADLINE_TYPE, HEADLINE_TRANSFORM} from './headlineConsts';
+import {HEADLINE_COLOR, HEADLINE_SIZE, HEADLINE_TYPE, HEADLINE_TRANSFORM, HEADLINE_ALIGN} from './headlineConsts';
 
-export {HEADLINE_TYPE, HEADLINE_SIZE, HEADLINE_COLOR, HEADLINE_TRANSFORM} from './headlineConsts';
+export {HEADLINE_TYPE, HEADLINE_SIZE, HEADLINE_COLOR, HEADLINE_TRANSFORM, HEADLINE_ALIGN} from './headlineConsts';
 
 const Headline = ({
   children,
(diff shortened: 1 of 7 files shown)


src/components/text/Headline.jsx
```diff
@@ -1,16 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import {HEADLINE_COLOR, HEADLINE_SIZE, HEADLINE_TYPE, HEADLINE_TRANSFORM} from './headlineConsts';
```
maciekbb: flow? :)

burakukula: @maciekbb on a separate branch, as soon as body text are implemented. 🐾

src/components/text/Headline.jsx
```diff
@@ -1,16 +1,17 @@
...
-export {HEADLINE_TYPE, HEADLINE_SIZE, HEADLINE_COLOR, HEADLINE_TRANSFORM} from './headlineConsts';
+export {HEADLINE_TYPE, HEADLINE_SIZE, HEADLINE_COLOR, HEADLINE_TRANSFORM, HEADLINE_ALIGN} from './headlineConsts';
 
 const Headline = ({
   children,
   type = HEADLINE_TYPE.H1,
   size = HEADLINE_SIZE.NORMAL,
   extraBold,
   transform,
+  align,
```
vergilius: @burakukula it's a verb, I think we should use a noun here: `alignment`, wdyt?

vergilius: or there's a reason to do it this way?

vergilius: @aju @maciekbb wdyt?

burakukula: @vergilius hmm, I thought more like: align-left/align-right, if we change it we should probably change it everywhere - we already use this pattern in SG.

burakukula: @vergilius acctualy, we have both patterns now. 🤯

vergilius: @burakukula okay, sounds reasonable

burakukula: @aju @maciekbb @vergilius I assume, `ALIGNMENT` in context of Content box and `TEXT_ALIGN`/`HEADLINE_ALIGN` in context of text make sense? or do you think we should have the same pattern everywhere?

PR #1508, opened 2018-10-15
Update content box

aju:
I added it so we know that this functionality is needed when designing new system

### DIFF
--- a/src/components/content-box/ContentBox.jsx
+++ b/src/components/content-box/ContentBox.jsx
@@ -9,8 +9,8 @@ export type ContentBoxPropsType = {
   spacedSmall?: boolean,
   full?: boolean,
   className?: string,
-  spacedTop?: $Values<typeof SIZE>,
-  spacedBottom?: $Values<typeof SIZE>
+  spacedTop?: ?$Values<typeof SIZE>,
+  spacedBottom?: ?$Values<typeof SIZE>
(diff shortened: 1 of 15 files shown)


src/components/content-box/ContentBox.jsx
```diff
@@ -1,11 +1,31 @@
...
+// @flow
+import * as React from 'react';
 import classNames from 'classnames';
 import {SIZE} from './ContentBoxConstants';
 
-const ContentBox = ({children, spacedTop, spacedBottom, spaced, full, className, ...props}) => {
+export type ContentBoxPropsType = {
+  children: React.Node,
+  spaced?: boolean,
+  spacedSmall?: boolean,
```
vergilius: this is new, right?

aju: yes

src/components/content-box/ContentBoxConstants.jsx
```diff
@@ -1,3 +1,4 @@
+// @flow
 const SIZE = {
```
burakukula: maybe `Object.freeze` here?

burakukula:
@aju you could try to bump SG from a branch in brainly-frontend, to check if there is a lot of changes if so to revert flow before you actually push new SG tag.

PR #1517, opened 2018-11-07
Add ref to Input

wasilakkf:


src/components/form-elements/Input.jsx
```diff
@@ -23,21 +23,35 @@ export const TYPE = {
...
+  value?: string | number,
+  size?: $Values<typeof SIZE>,
+  color?: $Values<typeof COLOR>,
+  valid?: boolean,
+  invalid?: boolean,
+  fullWidth?: boolean,
+  noBorder?: boolean,
+  withIcon?: boolean,
+  className?: string,
+  setInputRef: (ref: ?HTMLElement) => mixed
```
bartoszlorek: is it necessary to be required?

wasilakkf: Ups, mistake - good catch :+1:

wasilakkf:
@aju to fix [BBUG-323](https://jira.z-dn.net/browse/BBUG-323). ref is needed to focus on input if there was an error

src/components/form-elements/Input.jsx
```diff
@@ -23,21 +23,35 @@ export const TYPE = {
...
+  invalid?: boolean,
+  fullWidth?: boolean,
+  noBorder?: boolean,
+  withIcon?: boolean,
+  className?: string,
+  setInputRef?: (ref: ?HTMLElement) => mixed
 };
 
-const Input = props => {
+const Input = (props: PropsType) => {
```
maciekbb: If we don't pass ref to input anywhere right now maybe we could use ref forwarding https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components

PR #1524, opened 2018-12-13
Remove additional colors from palette

burakukula:
RIP navy blue 🙂


src/components/colors/colors.js
```diff
@@ -139,68 +139,6 @@ const color = {
       'variable': '$graySecondaryUltraLight',
       'hex': 'f5f8fa'
     }
-  ],
```
anna-wro: don't you need it?

matikrk: it's ok, last is open

PR #1529, opened 2018-12-17
Fix components after palette update

burakukula:
After refreshing palette we need to update some components:

 - Buttons to use a dark color on inverse version
 - option to use dark colors in texts  - e.g used for errors
 - valid and invalid selects, inputs and text area to use dark colors
 - refresh avatar to use brighter colors
 - deep blue links

### DIFF
--- a/src/components/text/pages/text.jsx
+++ b/src/components/text/pages/text.jsx
@@ -56,7 +56,7 @@ const TextExamples = () => {
           weight={LINK_WEIGHT.BOLD}
           size={LINK_SIZE.LARGE}
         >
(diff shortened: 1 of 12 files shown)


src/components/text/pages/text.jsx
```diff
@@ -56,7 +56,7 @@ const TextExamples = () => {
           weight={LINK_WEIGHT.BOLD}
           size={LINK_SIZE.LARGE}
         >
-          link / bold / standard / xlarge / standard
+          link / bold / standard / xlarge /  la la
```
matikrk: `la la` is better ?? :D

PR #1530, opened 2018-12-17
Add ultra light colors to palette

burakukula:
Additional ultra light colors in the palette.


src/components/layout/_layout.scss
```diff
@@ -59,7 +59,7 @@ $includeHtml: false !default;
         flex-wrap: wrap;
         padding: gutter(1 / 2);
         max-width: $layoutMaxWidth;
-        background-color: $graySecondaryLightest;
+        background-color: $graySecondaryUltraLight;
```
burakukula: We would need to update background color for body in brainly-frontend as well. I will do it when I bump SG.

PR #1533, opened 2019-01-30
added babel v7.x config, with compatibility with v6.x

PinkaminaDianePie:
added babel v7.x config, with compatibility with v6.x
`.babelrc` will be used by babel 6 until this repo will be upgraded to v7, while all consumers will be able to use styleguide by consuming `.babelrc.js` config

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
       {
         "modules": false
       }
-    ], 
+    ],
     "react",
     "flow"
   ],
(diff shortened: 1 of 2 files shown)


PinkaminaDianePie:
we can't drop compatibility just now, because if any of the fixes will be needed in the current styleguide - we won't be able to do them (storybook works with babel 6). we need to bump storybook with dependencies before droping config for babel 6, but since it could be time consuming - it's better to do it later

PR #1543, opened 2019-03-18
Update packages

burakukula:
Update webpack and gulp-sass. Changes solve issue that SG does not work on node 10.

I removed not a necessary folder in dist.
 In a separate PR I would like remove all the `pages` folder from dist, that contains `jsx` files used to create pages. But they are not needed in the dist at the end.
At the end it would be lovely to 🔪gulp and build everything with webpack. 

FROM:
folders-sg

TO:
Screenshot 2019-03-18 at 16 40 21

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "gulp-rename": "^1.2.2",
     "gulp-replace": "^0.6.1",
     "gulp-rev": "^7.0.0",
-    "gulp-sass": "^3.1.0",
+    "gulp-sass": "3.2.1",
     "gulp-sourcemaps": "^2.6.0",
     "gulp-svg-sprite": "^1.2.7",
     "gulp-util": "^3.0.6",
@@ -45,8 +45,8 @@
     "s3": "^4.4.0",
     "sass-lint": "^1.10.2",
     "through2": "^2.0.3",
-    "uglifyjs-webpack-plugin": "^0.4.6",
-    "webpack": "^3.6.0",
+    "uglifyjs-webpack-plugin": "^2.1.2",
+    "webpack": "4.6.0",
     "yargs": "^4.3.1"
   },
   "scripts": {
@@ -78,5 +78,8 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "node-uuid": "^1.4.8"
+  },
+  "resolutions": {
+    "natives": "1.1.3"
   }
 }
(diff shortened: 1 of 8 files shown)


scripts/tasks/build-react-interactive-pages.js
```diff
@@ -5,9 +5,9 @@ const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 const coreConfig = {
   target: 'node',
   module: {
-    loaders: [
+    rules: [
       {
-        test: /\.js|jsx?$/,
+        test: /\.js|.jsx$/,
```
aju: why ?:O

package.json
```diff
@@ -78,5 +78,8 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "node-uuid": "^1.4.8"
+  },
+  "resolutions": {
```
aju: why we need it?

scripts/tasks/build-react-interactive-pages.js
burakukula: @aju fixed.

package.json
burakukula: @aju we don't. fixed

scripts/tasks/build-react-pages.js
```diff
@@ -16,14 +16,11 @@ fs.readdirSync('node_modules')
 const coreConfig = {
   target: 'node',
   module: {
-    loaders: [
+    rules: [
       {
         test: /\.js|jsx?$/,
+        exclude: /\.json$/,
```
burakukula: We import package.json in some files to get and display version of the SG

package.json
```diff
@@ -45,8 +45,8 @@
     "s3": "^4.4.0",
     "sass-lint": "^1.10.2",
     "through2": "^2.0.3",
-    "uglifyjs-webpack-plugin": "^0.4.6",
-    "webpack": "^3.6.0",
+    "uglifyjs-webpack-plugin": "^2.1.2",
+    "webpack": "4.6.0",
```
vergilius: `^4.6.0`

package.json
```diff
@@ -30,7 +30,7 @@
     "gulp-rename": "^1.2.2",
     "gulp-replace": "^0.6.1",
     "gulp-rev": "^7.0.0",
-    "gulp-sass": "^3.1.0",
+    "gulp-sass": "3.2.1",
```
vergilius: `^3.2.1`

PR #1544, opened 2019-04-09
fixed types in text and context-box for flow autocomplete

PinkaminaDianePie:
fixed types in text and context-box for flow autocomplete

### DIFF
--- a/src/components/text/Link.jsx
+++ b/src/components/text/Link.jsx
@@ -62,7 +62,7 @@ const Link = (props: LinkPropsType) => {
   if (!href) {
     return (
       <Text
-        type={LINK_TYPE.SPAN}
+        type="span"
         {...additionalProps}
         className={linkClass}
       >
@@ -73,7 +73,7 @@ const Link = (props: LinkPropsType) => {
 
   return (
     <Text
-      type={LINK_TYPE.LINK}
+      type="a"
       {...additionalProps}
       className={linkClass}
       href={href}
--- a/src/components/text/textTypes.js
+++ b/src/components/text/textTypes.js
@@ -1,0 +1,75 @@
+// @flow strict
+export type TextTypeType =
+  |'span'
+  | 'p'
+  | 'h1'
+  | 'h2'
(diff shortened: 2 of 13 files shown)


src/components/text/textTypes.js
```diff
@@ -0,0 +1,75 @@
+// @flow strict
+export type TextTypeType =
+  |'span'
```
maciekbb: space? :D

PinkaminaDianePie: wow, i thought eslint will caught it so i didn't look carefully at indentation :D

src/components/text/textTypes.js
```diff
@@ -0,0 +1,75 @@
...
+  | 'peach-dark'
+  | 'peach'
+  | 'mustard'
+  | 'blue-dark'
+  | 'blue'
+  | 'blue-secondary'
+  | 'blue-secondary-light';
+
+export type TextWeightType =
+  |'regular'
```
vergilius: some whitespaces missing here and there 👇

PinkaminaDianePie:
flow strict in components will be time consuming to introduce, so i add it only to the new files for now

vergilius:
awesome!

src/components/text/textTypes.js
PinkaminaDianePie: fixed

src/components/text/textTypes.js
PinkaminaDianePie: fixed

src/components/text/Link.jsx
```diff
@@ -62,7 +62,7 @@ const Link = (props: LinkPropsType) => {
   if (!href) {
     return (
       <Text
-        type={LINK_TYPE.SPAN}
+        type="span"
```
aju: What do we gain by this change?

PinkaminaDianePie: in the newest Flow and VS Code new version can be auto-suggested, while the first one needs to be remembered :)

PR #1545, opened 2019-04-10
Flow to most used components

PinkaminaDianePie:


src/components/action-list/ActionList.jsx
```diff
@@ -15,10 +27,19 @@ export const ALIGNMENT = {
   STRETCH: 'stretch'
 };
 
-const ActionList = ({children, toTop, direction, align, noWrap, className, ...props}) => {
+type ActionListPropsType = {
+  children: Node,
+  toTop?: ?boolean,
```
vergilius: is the problem with optional type `?: ?` in sg itself or somewhere else?

PinkaminaDianePie: in our main project, we use different patterns, some of them can results to `null`, some of them to `undefined`, but at the end it doesn't matter because `classnames` just doesn't use `null`\`undefined`.

vergilius: I see 👍

PR #1546, opened 2019-04-11
Form elements types

burakukula:


aju:
Flow strict maybe?

src/components/form-elements/Input.jsx
```diff
@@ -38,10 +70,10 @@ export const COLOR = Object.freeze({
 });
 
 type PropsType = {
-  type?: $Values<typeof TYPE>,
+  type?: InputType,
```
PinkaminaDianePie: props with unions, such as `InputType`, `InputSizeType`, `InputColorType` is better to make nullable: `type?: ?InputType`, because in main app we use patterns like
`color={ ifSomething ? COLOR : null}` and such patterns will be broken

burakukula: @PinkaminaDianePie ah, yeah. sure, I will change that in a sec. thx for feedback. 🙂

PR #1548, opened 2019-04-15
Make textareaRef nullable

burakukula:
Bugfix to form components


PinkaminaDianePie:
maybe `HTMLTextAreaElement` instead of `HTMLElement` ?

PR #1549, opened 2019-04-16
Icon thumbs up and exclamation mark

PinkaminaDianePie:
added new icons: "thumbs up" and "exclamation mark"
image
image


src/images/icons/exclamation_mark.svg
```diff
@@ -0,0 +1 @@
+<svg width="32" height="32" xmlns="http://www.w3.org/2000/svg"><g transform="translate(13)" fill="#000" fill-rule="evenodd"><path d="M6 28.8c0 .88-.294 1.633-.881 2.26C4.53 31.687 3.825 32 3 32c-.825 0-1.531-.313-2.119-.94C.294 30.433 0 29.68 0 28.8c0-.88.294-1.633.881-2.26.588-.627 1.294-.94 2.119-.94.825 0 1.531.313 2.119.94.587.627.881 1.38.881 2.26z" fill-rule="nonzero"/><rect width="6" height="23.04" rx="1.92"/></g></svg>
```
burakukula: @PinkaminaDianePie I think you can optimize this svg. we maybe could get rid of group here? Also, in both svg's viewbox is missing.

PinkaminaDianePie: minimizer which we always use for svg (https://jakearchibald.github.io/svgomg/) produced such a result. we had viewbox in the original file, but minimizer just removed it (seems like if your dimensions same with viewbox it can be safely omitted)

PinkaminaDianePie: done

PR #1550, opened 2019-04-16
Spacing util function

burakukula:
It would be great to have a common set of spacings between design and FE teams.
The following system is already used on the new HP for not logged in user:

Screenshot 2019-04-16 at 18 03 09

I would like to introduce new helper function that would help us to slowly replace gutter with spacings based on 8px system.

Usage example:

`.your-class {
  padding: spacing(m);
}`

it will output:
`.your-class {
  padding: 24px;
}`

WDYT? If you are ok with this change, I will extend documentation.


burakukula:
@aju if anyone have time to do it, would be awesome, but it's a huge task, so I guess better to just consider gutter and rythm as deprecated and do it slowly. WDYT?

PinkaminaDianePie:
fibonacci makes more sense for me than just random multipliers on 24px :+1:

burakukula:
@PinkaminaDianePie yeah, this is the system that is already used in the new designs, so maybe slowly we can replace gutter(). And we gonna have spacing system in sync with the design team.

PR #1556, opened 2019-04-25
remove input text stroke

aju:
It looks weird specially in placeholders


aju:
Screenshot 2019-04-25 at 11 16 35
example of broken placeholder

PR #1558, opened 2019-04-29
Some more flow types

PinkaminaDianePie:
add flow types to separators, icon as button and overlayed box components

### DIFF
--- a/src/components/icon-as-button/IconAsButton.jsx
+++ b/src/components/icon-as-button/IconAsButton.jsx
@@ -1,22 +1,52 @@
+// @flow
 import React from 'react';
-import PropTypes from 'prop-types';
+import type {Node} from 'react';
 import classNames from 'classnames';
-import Icon, {TYPE, ICON_COLOR} from '../icons/Icon';
+import * as IconModule from '../icons/Icon';
 import IconAsButtonContent from './subcomponents/IconAsButtonContent';
 
+const {ICON_COLOR, default: Icon} = IconModule;
+
+type IconSizeType =
+  | 'normal'
+  | 'small'
+  | 'xsmall'
+  | 'xxsmall';
+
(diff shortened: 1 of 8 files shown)


src/components/icon-as-button/IconAsButton.jsx
```diff
@@ -1,22 +1,52 @@
+// @flow
 import React from 'react';
-import PropTypes from 'prop-types';
+import type {Node} from 'react';
 import classNames from 'classnames';
-import Icon, {TYPE, ICON_COLOR} from '../icons/Icon';
+import * as IconModule from '../icons/Icon';
```
PinkaminaDianePie: `import as *` is only way to import types in a way compatible with our vs code intellisence

aju: Why not just as Icon ?

aju: ah to spread it 👍

src/components/icon-as-button/IconAsButton.jsx
```diff
@@ -1,22 +1,52 @@
...
+  [SIZE.SMALL]: 18,
+  [SIZE.XSMALL]: 14,
+  [SIZE.XXSMALL]: 10
+};
+
+type IconAsButtonPropsType = {
+  size?: IconSizeType,
+  color?: IconModule.IconColorType,
+  type?: ?IconModule.IconTypeType,
+  children?: ?Element<any>,
```
maciekbb: hmm why not `React.Node`?

PinkaminaDianePie: We should pass exactly one element, not an array, and `Element` is just like that, while `Node` can be an array or literal. We had such test before, you can check it :)

PR #1560, opened 2019-04-29
flow-typed :)

Icokie:
added in order to use flow strict in SG


flow-typed/npm/node-uuid_v1.4.x.js
```diff
@@ -0,0 +1,34 @@
+// flow-typed signature: 1c22a8c07f5233703efd793409d6890e
```
maciekbb: @aju shouldn't we drop uuid?

aju: I think we should :D

Icokie: done :)

PR #1562, opened 2019-04-29
added documentation POC and release scripts

PinkaminaDianePie:
added **latest** folder to build, so jsdoc will be able to always point to the latest version. added jsdoc example as well


aju:
I don't like it :/ I think we should first try by fixing redirecting on s3 
https://styleguide.brainly.com/docs/containers.html#box should redirect to https://styleguide.brainly.com/XXXX/docs/containers.html#box 
Where XXXX is newest version

aju:
To make it stick to the version (link) I would prefer to do in on deployment step so we don't have to update files with each commit.

tomasz-sodzawiczny:
what for?

PinkaminaDianePie:
to be able to use such links in the docs:
`* @see https://styleguide.brainly.com/latest/docs/containers.html#box`, right now we can point only to specific version in SG, but we can't point to the latest

tomasz-sodzawiczny:
Got it, good idea.
Copying seems hacky though. I guess if we can't do it properly with a redirect we can go with it.

PinkaminaDianePie:
For now, it's the best solution we've found. The other was even hackier :)

PR #1564, opened 2019-05-02
added flow types to most of components

PinkaminaDianePie:
added flow types to most of components

### DIFF
--- a/src/components/avatar/Avatar.jsx
+++ b/src/components/avatar/Avatar.jsx
@@ -18,7 +18,7 @@ export const SIZE = {
   XXLARGE: 'xxlarge'
 };
 
-export const ICON_SIZE_PADDED = {
+export const ICON_SIZE_FOR_AVATARS_WITH_BORDER = {
   [SIZE.SMALL]: 22,
   [SIZE.NORMAL]: 30,
   [SIZE.LARGE]: 46,
@@ -58,7 +58,11 @@ const Avatar = ({size = SIZE.NORMAL, border = false, spaced, imgSrc, className, 
   } else {
     avatarContent = (
       <div className="sg-avatar__image sg-avatar__image--icon">
-        <Icon type="profile" color="gray-secondary" size={border ? ICON_SIZE_PADDED[size] : ICON_SIZE[size]} />
+        <Icon
+          type="profile"
+          color="gray-secondary"
+          size={border ? ICON_SIZE_FOR_AVATARS_WITH_BORDER[size] : ICON_SIZE[size]}
+        />
       </div>
     );
   }
--- a/src/components/avatar/Avatar.jsx
+++ b/src/components/avatar/Avatar.jsx
@@ -1,23 +1,52 @@
+// @flow
 import React from 'react';
-import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import Icon, {TYPE as ICON_TYPE, ICON_COLOR} from '../icons/Icon';
+import Icon from '../icons/Icon';
 
-const DEFAULT_ICON = ICON_TYPE.PROFILE;
-const DEFAULT_COLOR = ICON_COLOR.GRAY_SECONDARY;
-const BORDER_SIZE = 2;
+export type AvatarSizeType =
+  | 'small'
+  | 'normal'
+  | 'large'
+  | 'xlarge'
+  | 'xxlarge';
 
 export const SIZE = {
-  SMALL: {className: 'small', iconSize: 24},
-  NORMAL: {className: 'normal', iconSize: 32},
-  LARGE: {className: 'large', iconSize: 48},
-  XLARGE: {className: 'xlarge', iconSize: 64},
-  XXLARGE: {className: 'xxlarge', iconSize: 120}
+  SMALL: 'small',
+  NORMAL: 'normal',
+  LARGE: 'large',
+  XLARGE: 'xlarge',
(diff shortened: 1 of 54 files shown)


src/components/badges/subcomponents/Badge.spec.jsx
```diff
@@ -60,14 +60,4 @@ describe('<Badge />', () => {
 
     expect(badge.hasClass('sg-badge--rounded')).toEqual(true);
   });
-
-  it('logs error when no child has been passed', () => {
```
aju: why removing those checks?

src/components/badges/TextBadge.jsx
```diff
@@ -1,22 +1,35 @@
+// @flow
```
aju: why not strict?

src/components/card/CardHole.jsx
```diff
@@ -1,7 +1,27 @@
+// @flow strict
 import React from 'react';
-import PropTypes from 'prop-types';
+import type {Node} from 'react';
 import classnames from 'classnames';
 
+export type ColorType =
+  | 'blue'
```
aju: maybe we should have generic ColorType with all colors? WDYT?

src/components/header/HeaderContainer.jsx
```diff
@@ -1,8 +1,15 @@
+// @flow strict
 import React from 'react';
-import PropTypes from 'prop-types';
+import type {Node} from 'react';
```
aju: `import * as React from 'react';`

src/components/header/HeaderContent.jsx
```diff
@@ -1,8 +1,15 @@
+// @flow strict
 import React from 'react';
-import PropTypes from 'prop-types';
+import type {Node} from 'react';
```
aju: `import * as React from 'react';`

src/components/subject-icons/SubjectIcon.jsx
```diff
@@ -1,7 +1,92 @@
+// @flow
```
aju: strict?

src/components/badges/subcomponents/Badge.spec.jsx
PinkaminaDianePie: this error was thrown in runtime by `prop-types` before, right now it's a dev-time error in flow :)

aju: 👍

src/components/badges/TextBadge.jsx
PinkaminaDianePie: it depends on non-strict module, `Text`, which Yury works on. after he will introduce `flow strict` there, i will be able to cover TextBadge as well. don't worry, status of `flow strict` coverage is tracked here: https://docs.google.com/spreadsheets/d/1uacUj6OAwSeM9kn5IuIdtOfw8QGQNpV1_I3Zqk5SZE0/edit?pli=1#gid=0

src/components/subject-icons/SubjectIcon.jsx
PinkaminaDianePie: same as above

src/components/card/CardHole.jsx
PinkaminaDianePie: maybe at least a few such types - different components allows only a subset of colors we have, not all of them

src/components/header/HeaderContainer.jsx
PinkaminaDianePie: our intellisence doesn't recognize type of `Node` in that case

src/components/header/HeaderContent.jsx
PinkaminaDianePie: same as above

src/components/avatar/Avatar.jsx
```diff
@@ -1,23 +1,52 @@
...
-  XXLARGE: {className: 'xxlarge', iconSize: 120}
+  SMALL: 'small',
+  NORMAL: 'normal',
+  LARGE: 'large',
+  XLARGE: 'xlarge',
+  XXLARGE: 'xxlarge'
 };
 
-const Avatar = ({size = SIZE.NORMAL, border = false, spaced, imgSrc, className, link, title, ...props}) => {
+export const ICON_SIZE_PADDED = {
```
burakukula: @PinkaminaDianePie I am not sure about this namespace. Can we go with `ICON_SIZE_WITH_SPACE` or something similar?

burakukula: Or maybe somehow indicate that it is used for avatars with border?

src/components/header/HeaderContainer.jsx
burakukula: but maybe you could do something like this? :
```
import * as React from 'react';
something: React.Node
```

PinkaminaDianePie: yeah, i meant exactly that - jsdoc based intellisense shows `any` instead of `Node` on hover. no idea why does it work like that. direct import works fine here

src/components/avatar/Avatar.jsx
PinkaminaDianePie: done

src/components/header/HeaderContainer.jsx
burakukula: Ah, ok, I get it now. 🙃

PR #1570, opened 2019-05-07
changed last components from prop-types to flow

PinkaminaDianePie:
changed last components from prop-types to flow

### DIFF
--- a/src/components/dropdowns/DropdownContainer.jsx
+++ b/src/components/dropdowns/DropdownContainer.jsx
@@ -1,24 +1,47 @@
+// @flow strict
 import React, {Component} from 'react';
-import PropTypes from 'prop-types';
 import Dropdown from './Dropdown';
 
-class DropdownContainer extends Component {
+type ItemType = {
+  id: string,
+  text: string
+};
+
+type PropsType = {
+  label: string,
+  fullWidth?: boolean,
+  fixed?: boolean,
+  isOpened?: boolean,
+  items: Array<ItemType>,
+  className?: string,
+  currentItem?: ItemType,
+  onChange: string => mixed,
+  onToggle?: SyntheticMouseEvent<HTMLDivElement>
+};
+
+type StateType = {
+  currentItem: ItemType,
+  label: string
+};
+
+class DropdownContainer extends Component<PropsType, StateType> {
   static defaultProps = {
     items: []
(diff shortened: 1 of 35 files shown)


src/components/dropdowns/DropdownContainer.jsx
```diff
@@ -1,24 +1,47 @@
...
+  items: Array<ItemType>,
+  className?: string,
+  currentItem?: ItemType,
+  onChange: string => mixed,
+  onToggle?: SyntheticMouseEvent<HTMLDivElement>
+};
+
+type StateType = {
+  currentItem: ItemType,
+  label: string
```
aju: why?

src/components/form-elements/Input.jsx
```diff
@@ -69,7 +69,7 @@ export const COLOR = Object.freeze({
   LIGHT_ALT: 'light-alt'
 });
 
-type PropsType = {
+export type PropsType = {
```
aju: `InputPropsType` ?

src/docs/components/ComponentSettingsInput.jsx
```diff
@@ -1,9 +1,16 @@
+// @flow
 import React from 'react';
-import PropTypes from 'prop-types';
 import Input from 'form-elements/Input';
 import Checkbox from 'form-elements/Checkbox';
 
-const ComponentSettingsInput = ({values, currentValue = '', onChange}) => {
+type PropsType = {
+  // eslint-disable-next-line flowtype/no-primitive-constructor-types
+  values: Boolean | Number | String,
```
aju: ?? can it be just boolean etc?

src/components/dropdowns/DropdownContainer.jsx
PinkaminaDianePie: what "why"?

src/docs/components/ComponentSettingsInput.jsx
PinkaminaDianePie: no, because we'll need to change logic then. Right now this component accept actual constructors, not strings with their names

src/components/form-elements/Input.jsx
PinkaminaDianePie: fixed

PR #1573, opened 2019-05-09
switch from uglify to terser

PinkaminaDianePie:


burakukula:
@PinkaminaDianePie bump to patch maybe?

PR #1576, opened 2019-05-13
Introduce flexbox component

burakukula:
I know it's big, but it's kinda one thing that I introduce here and big part of it is documentation. So... Flex and it's documentation:

 - introduction of Flex component - jsdocs and documentation is included. I added also Markdown docs to have some examples of usage. I added small description in the flex page
 - documentation of the spacing was missing for a while so I added also in this PR, as spacings are used in margins setup in the Flex component

flexbox mov

### DIFF
--- a/src/components/flex/Flex.jsx
+++ b/src/components/flex/Flex.jsx
@@ -63,31 +63,31 @@ const Flex = (props: FlexPropsType) => {
   } = props;
 
   const flexClass = classNames(
-    style['sg-flex'],
-    fullWidth ? style['sg-flex--full-width'] : null,
-    fullHeight ? style['sg-flex--full-height'] : null,
-    noShrink ? style['sg-flex--no-shrink'] : null,
+    'sg-flex',
+    fullWidth ? 'sg-flex--full-width' : null,
+    fullHeight ? 'sg-flex--full-height' : null,
+    noShrink ? 'sg-flex--no-shrink' : null,
     //flex setup
-    inlineFlex ? style['sg-flex--inline'] : null,
-    alignItems ? style[`sg-flex--align-items-${alignItems}`] : null,
-    alignContent ? style[`sg-flex--align-content-${alignContent}`] : null,
-    justifyContent ? style[`sg-flex--justify-content-${justifyContent}`] : null,
-    wrap ? style['sg-flex--wrap'] : null,
-    wrapReverse ? style['sg-flex--wrap-reverse'] : null,
-    alignSelf ? style[`sg-flex--align-self-${alignSelf}`] : null,
-    direction === FLEX_DIRECTION.COLUMN ? style['sg-flex--column'] : null,
+    inlineFlex ? 'sg-flex--inline' : null,
+    alignItems ? `sg-flex--align-items-${alignItems}` : null,
+    alignContent ? `sg-flex--align-content-${alignContent}` : null,
+    justifyContent ? `sg-flex--justify-content-${justifyContent}` : null,
+    wrap ? 'sg-flex--wrap' : null,
+    wrapReverse ? 'sg-flex--wrap-reverse' : null,
+    alignSelf ? `sg-flex--align-self-${alignSelf}` : null,
+    direction === FLEX_DIRECTION.COLUMN ? 'sg-flex--column' : null,
     direction === FLEX_DIRECTION.COLUMN_REVERSE ?
-      style['sg-flex--column-reverse'] :
+      'sg-flex--column-reverse' :
       null,
-    direction === FLEX_DIRECTION.ROW ? style['sg-flex--row'] : null,
+    direction === FLEX_DIRECTION.ROW ? 'sg-flex--row' : null,
     direction === FLEX_DIRECTION.ROW_REVERSE ?
-      style['sg-flex--row-reverse'] :
+      'sg-flex--row-reverse' :
       null,
-    margin ? style[`sg-flex--margin-${margin}`] : null,
-    marginTop ? style[`sg-flex--margin-top-${marginTop}`] : null,
-    marginRight ? style[`sg-flex--margin-right-${marginRight}`] : null,
-    marginBottom ? style[`sg-flex--margin-bottom-${marginBottom}`] : null,
-    marginLeft ? style[`sg-flex--margin-left-${marginLeft}`] : null,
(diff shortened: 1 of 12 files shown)


src/components/flex/Flex.jsx
```diff
@@ -0,0 +1,265 @@
...
+
+export type FlexPropsType = {
+  /**
+   * Children to be render inside of the Box. Takes no effect if **imgSrc** prop specified.
+   * @example <Flex>
+   *            children to be render here
+   *          </Flex>
+   *
+   */
+  children: ?Node,
```
PinkaminaDianePie: is children optional? if yes, it should be `children?: Node`, otherwise just `children: Node`. `Node` type is nullable by default, so `?Node` is the same as `Node`

src/components/flex/Flex.jsx
```diff
@@ -0,0 +1,265 @@
+// @flow
```
PinkaminaDianePie: `// @flow strict` maybe?

src/components/flex/Flex.jsx
```diff
@@ -0,0 +1,265 @@
...
+export {
+  FLEX_DIRECTION,
+  FLEX_JUSTIFY_VALUES,
+  FLEX_ALIGNMENT_VALUES,
+  SPACING_SET as FLEX_MARGINS
+};
+
+export type FlexPropsType = {
+  /**
+   * Children to be render inside of the Box. Takes no effect if **imgSrc** prop specified.
```
PinkaminaDianePie: seems like copy-paste :) `imgSrc` makes no sense here :)

src/components/flex/Flex.jsx
```diff
@@ -0,0 +1,265 @@
...
+   * Specify flex direction
+   * @example <Flex direction="column">
+   *            component content
+   *          </Flex>
+   * @see direction=column https://styleguide.brainly.com/latest/docs/interactive.html?direction=column#flexbox
+   * @see direction=column-reverse https://styleguide.brainly.com/latest/docs/interactive.html?direction=column-reverse#flexbox
+   * @see direction=row https://styleguide.brainly.com/latest/docs/interactive.html?direction=row#flexbox
+   * @see direction=row-reverse https://styleguide.brainly.com/latest/docs/interactive.html?direction=row-reverse#flexbox
+   */
+  direction?: ?FlexDirectionType,
```
PinkaminaDianePie: just `direction?: FlexDirectionType,`

src/components/flex/Flex.jsx
```diff
@@ -0,0 +1,265 @@
...
+   * @see justifyContent=center https://styleguide.brainly.com/latest/docs/interactive.html?justifyContent=center#flexbox
+   * @see justifyContent=flex-start https://styleguide.brainly.com/latest/docs/interactive.html?ustifyContent=flex-start#flexbox
+   * @see justifyContent=flex-end https://styleguide.brainly.com/latest/docs/interactive.html?justifyContent=flex-end#flexbox
+   * @see justifyContent=baseline https://styleguide.brainly.com/latest/docs/interactive.html?justifyContent=baseline#flexbox
+   * @see justifyContent=space-between https://styleguide.brainly.com/latest/docs/interactive.html?justifyContent=space-between#flexbox
+   * @see justifyContent=space-around https://styleguide.brainly.com/latest/docs/interactive.html?justifyContent=space-around#flexbox
+   * @see justifyContent=space-evently https://styleguide.brainly.com/latest/docs/interactive.html?justifyContent=space-evently#flexbox
+   * @see justifyContent=stretch https://styleguide.brainly.com/latest/docs/interactive.html?justifyContent=stretch#flexbox
+   */
+  justifyContent?: ?FlexJustifyValuesType,
```
PinkaminaDianePie: same here. please make all types not-nullable if possible.

src/components/flex/Flex.jsx
```diff
@@ -0,0 +1,265 @@
...
+    marginTop,
+    marginBottom,
+    marginLeft,
+    marginRight,
+    children,
+    className,
+    ...otherProps
+  } = props;
+
+  const flexClass = classNames(
```
PinkaminaDianePie: maybe organize it as an object? like we did in other places

src/components/flex/FlexConsts.js
```diff
@@ -0,0 +1,38 @@
+// @flow
```
PinkaminaDianePie: `// @flow strict`?

src/components/flex/Flex.jsx
burakukula: ups, true

src/components/flex/Flex.jsx
```diff
@@ -0,0 +1,265 @@
...
+  | 'xl'
+  | 'xxl'
+  | 'xxxl'
+  | 'xxxxl';
+
+export {
+  FLEX_DIRECTION,
+  FLEX_JUSTIFY_VALUES,
+  FLEX_ALIGNMENT_VALUES,
+  SPACING_SET as FLEX_MARGINS
```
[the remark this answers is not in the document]
burakukula: @PinkaminaDianePie I would like to introduce `Styled` component, that we also have in a new SG, which would use the same set for paddings. I would like them to share this settigns. But maybe I can introduce it when I will introduce `Styled` compontent.

src/components/flex/Flex.jsx
burakukula: @PinkaminaDianePie I guess children should not be optional in this case. on my way. thx.

src/components/flex/FlexConsts.js
burakukula: @PinkaminaDianePie fixed.

src/components/flex/Flex.jsx
burakukula: @PinkaminaDianePie fixed

src/components/flex/Flex.jsx
burakukula: @PinkaminaDianePie done

src/components/flex/Flex.jsx
burakukula: @PinkaminaDianePie done

src/components/flex/Flex.jsx
burakukula: @PinkaminaDianePie done

PR #1577, opened 2019-05-13
remove node-uuid

matikrk:
* write function generateRandomString

### DIFF
--- a/src/js/generateRandomString.js
+++ b/src/js/generateRandomString.js
@@ -1,0 +1,5 @@
+// @flow
+
+const generateRandomString = () => Math.random().toString(32).slice(2);
+
+export default generateRandomString;
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "143.6.5",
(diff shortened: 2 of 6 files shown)


src/js/generateRandomString.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+const generateRandomString = () => Math.random().toString(32).slice(2);
```
mpacholec: maybe this one: 🙂 
```js
import {randomBytes} from 'crypto';

export const getRandomString= (length = 16, encoding='hex') => randomBytes(length).toString(encoding);
```

matikrk: We need a simple one. No need other imports.

mpacholec: `crypto` is part of nodeJS itself so why not to use it :) - but this is just a suggestion

aju: @mpacholec but we use it in browser ?:P

mpacholec: @aju https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto

PR #1579, opened 2019-05-14
Fix Zooming to Input and Select on Focus in iOS Devices

bartoszlorek:
The font size of inputs and selects should be equal or greater than `16px` for iOS to not zooming. We agreed with design to go with it for all inputs.

### DIFF
--- a/src/components/form-elements/_select.scss
+++ b/src/components/form-elements/_select.scss
@@ -1,7 +1,7 @@
 $selectHeight: rhythm(3 / 2);
 $selectBackground: $white;
 $selectTextColor: $grayPrimary;
-$selectFontSize: fontSize(small);
+$selectFontSize: 16px; // fix zooming to select on focus in iOS devices
 $selectIconSize: rhythm(1 / 3);
 $selectBorderRadius: 7px;
 $selectActiveBorderColor: $grayPrimary;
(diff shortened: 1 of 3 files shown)


src/components/form-elements/_select.scss
```diff
@@ -1,7 +1,7 @@
 $selectHeight: rhythm(3 / 2);
 $selectBackground: $white;
 $selectTextColor: $grayPrimary;
-$selectFontSize: fontSize(small);
+$selectFontSize: 16px; // fix zooming to select on focus in iOS devices
```
matzimowski: LET IT BE. USERS GONNA BE HAPPY

src/components/form-elements/_select.scss
```diff
@@ -1,7 +1,7 @@
 $selectHeight: rhythm(3 / 2);
 $selectBackground: $white;
 $selectTextColor: $grayPrimary;
-$selectFontSize: fontSize(small);
```
bartoszlorek: `fontSize(default)` can also work but I want to be more explicit for the future.

src/components/form-elements/_text-input.scss
```diff
@@ -4,7 +4,7 @@ $inputSmallHeight: rhythm(5 / 4);
 
 $inputBackground: $white;
 $inputTextColor: $black;
-$inputFontSize: fontSize(small);
+$inputFontSize: 16px; // fix zooming to input on focus in iOS devices
```
matzimowski: LET IT BE. USERS GONNA BE HAPPY

PR #1581, opened 2019-05-24
Introduce new buttons

burakukula:
Intorduction of the new buttons:

 - **Primary buttons** - they do not change at all in terms of colors, just adjustments of sizes (buttons are bigger form 36px to 40px) with no shadow, border radius and so on. Additionally, I removed all the peach buttons, as they are not used in any place in the application (neither in JSX or in twig):

BEFORE:
Screenshot 2019-05-24 at 08 32 33

AFTER:
Screenshot 2019-05-24 at 08 32 41

 - **Primary buttons inverted** - for this type of the buttons we have the biggest change. Buttons type `INVERSE` and `ALT_INVERSE` were merged into one in terms of the look, `DARK_INVERSE` changes it's styling:

BEFORE:
Screenshot 2019-05-24 at 09 34 16

AFTER:
Screenshot 2019-05-24 at 09 34 26

 - **Secondary and secondary small** - these buttons ware merged in terms of the size - they are both 32px now. Regarding styling changes, secondary buttons are exactly the same as the primary once:

BEFORE:
Screenshot 2019-05-24 at 09 38 00

AFTER:
Screenshot 2019-05-24 at 09 40 52

We have 2 additional cases where we use buttons secondary with icons: it's `THANK YOU` and `MARK AS BRAINLIEST` which I prepare as custom cases. the second one gets its own type, which makes it easier for me to replace it in the application.

BEFORE:
Screenshot 2019-05-24 at 09 48 21

AFTER:
Screenshot 2019-05-24 at 09 43 40

hover/active:
Screenshot 2019-05-24 at 09 49 40

All of the above changes are temporary. The main goal here is to adjust current styles to the new styling of the buttons without changing too much on the application side. In the next step, I will introduce a new component `Button.jsx` (to rule them all), which would have final naming and separate styles. I will replace existing buttons with the new slowly.

### DIFF
--- a/src/components/buttons/ButtonPrimary.jsx
+++ b/src/components/buttons/ButtonPrimary.jsx
@@ -18,7 +18,6 @@ export const BUTTON_PRIMARY_TYPE = {
   INVERSE: 'inverse',
   ALT_INVERSE: 'alt-inverse',
   DARK_INVERSE: 'dark-inverse',
-  PEACH: 'peach',
   FB: 'fb'
 };
 
--- a/src/components/buttons/ButtonSecondary.jsx
+++ b/src/components/buttons/ButtonSecondary.jsx
@@ -19,7 +19,7 @@ export const BUTTON_SECONDARY_TYPE = {
   ALT_INVERSE: 'alt-inverse',
   DARK_INVERSE: 'dark-inverse',
   ACTIVE_INVERSE: 'active-inverse',
-  PEACH: 'peach'
+  ACTIVE_MUSTRAD: 'active-mustard'
 };
 
 type ButtonSecondaryPropsType = {
--- a/src/components/buttons/ButtonSecondary.jsx
+++ b/src/components/buttons/ButtonSecondary.jsx
@@ -19,7 +19,7 @@ export const BUTTON_SECONDARY_TYPE = {
   ALT_INVERSE: 'alt-inverse',
   DARK_INVERSE: 'dark-inverse',
   ACTIVE_INVERSE: 'active-inverse',
-  ACTIVE_MUSTRAD: 'active-mustard'
+  ACTIVE_MUSTARD: 'active-mustard'
 };
 
 type ButtonSecondaryPropsType = {
--- a/src/components/buttons/_buttons-mixins.scss
+++ b/src/components/buttons/_buttons-mixins.scss
@@ -76,7 +76,10 @@ $buttonPrimaryFbHoverColor: #4367a9;
   &--alt {
     background-color: $buttonAltColor;
 
-    &:hover {
+    &:hover,
+    &:focus,
+    &:active,
+    &:active:focus {
       background-color: mix($bluePrimaryDark, $buttonAltColor, 80%);
       border-color: mix($bluePrimaryDark, $buttonAltColor, 80%);
     }
@@ -95,7 +98,10 @@ $buttonPrimaryFbHoverColor: #4367a9;
   &--dark {
     background-color: $buttonDarkColor;
 
-    &:hover {
+    &:hover,
+    &:focus,
+    &:active,
+    &:active:focus {
       background-color: mix($grayPrimary, $buttonDarkColor, 80%);
       border-color: mix($grayPrimary, $buttonDarkColor, 80%);
     }
@@ -119,8 +125,9 @@ $buttonPrimaryFbHoverColor: #4367a9;
     background: none;
 
     &:hover,
+    &:focus,
     &:active,
-    &:focus {
+    &:active:focus {
       background-color: mix($black, $transparentWhite, 12%);
       border-color: $black;
     }
@@ -143,8 +150,9 @@ $buttonPrimaryFbHoverColor: #4367a9;
     border: 2px solid $white;
 
     &:hover,
+    &:focus,
     &:active,
-    &:focus {
+    &:active:focus {
       background-color: mix($graySecondaryUltraLight, $white, 80%);
       border-color: mix($graySecondaryUltraLight, $white, 80%);
     }
@@ -191,8 +199,9 @@ $buttonPrimaryFbHoverColor: #4367a9;
     }
 
     &:hover,
+    &:focus,
     &:active,
-    &:focus {
+    &:active:focus {
       border: none;
     }
   }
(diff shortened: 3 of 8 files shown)


src/components/buttons/ButtonPrimary.jsx
```diff
@@ -18,7 +18,6 @@ export const BUTTON_PRIMARY_TYPE = {
   INVERSE: 'inverse',
   ALT_INVERSE: 'alt-inverse',
   DARK_INVERSE: 'dark-inverse',
-  PEACH: 'peach',
```
burakukula: This type of buttons is not used at all the application. 🔪

src/components/buttons/ButtonSecondary.jsx
```diff
@@ -19,7 +19,7 @@ export const BUTTON_SECONDARY_TYPE = {
   ALT_INVERSE: 'alt-inverse',
   DARK_INVERSE: 'dark-inverse',
   ACTIVE_INVERSE: 'active-inverse',
-  PEACH: 'peach'
```
burakukula: This type of buttons is not used at all the application. 🔪

src/components/buttons/ButtonSecondary.jsx
```diff
@@ -19,7 +19,7 @@ export const BUTTON_SECONDARY_TYPE = {
...
+  ACTIVE_MUSTRAD: 'active-mustard'
```
burakukula: I add this type to have a separate styling for `MARK AS BEST ANSWER` button.

src/components/buttons/_buttons-mixins.scss
```diff
@@ -32,85 +25,213 @@ $buttonActiveScale: scale(0.985);
...
 
-  &__container {
-    display: flex;
-    align-items: center;
-    justify-content: center;
-  }
-
-  &--full-width {
-    width: 100%;
+  &.sg-button-primary--disabled,
```
burakukula: We have now changes in color of each button on hover, so I need to make sure that for disabled buttons does not happen. Probably there is a better way to do it in SASS, but honestly, I consider it as a waste of time to figure it out now, as I will remove all this code soon. 🤷‍♀️I would prefer to put more attention into the new Button component instead. But if you have some ideas I can, of course, make adjustments here 🙃

src/components/buttons/_buttons_primary.scss
```diff
@@ -1,8 +1,7 @@
-$buttonPrimaryHeight: rhythm(3 / 2);
-$buttonPrimaryFontSize: fontSize(default);
-$buttonPrimaryBorderRadius: 9px;
+$buttonPrimaryHeight: 40px;
```
burakukula: bye bye `rythm`

src/components/buttons/_buttons_primary.scss
```diff
@@ -1,8 +1,7 @@
-$buttonPrimaryHeight: rhythm(3 / 2);
-$buttonPrimaryFontSize: fontSize(default);
-$buttonPrimaryBorderRadius: 9px;
-$buttonPrimaryPadding: gutter(1);
-$buttonPrimaryFbColor: #3b5998;
+$buttonPrimaryHeight: 40px;
+$buttonPrimaryFontSize: 15px;
+$buttonPrimaryBorderRadius: 20px;
+$buttonPrimaryPadding: spacing(m);
```
burakukula: welcome `spacing`

src/components/buttons/_buttons-mixins.scss
maciekbb: I'd not put it inside mixin maybe...?

src/components/buttons/_buttons-mixins.scss
```diff
@@ -32,85 +25,213 @@ $buttonActiveScale: scale(0.985);
   white-space: nowrap;
   font-weight: $fontWeightBold;
   text-transform: uppercase;
-  box-shadow: $componentShadow;
-  transition-property: box-shadow, transform, opacity;
+  box-shadow: none;
```
matzimowski: remove me maybe?

src/components/buttons/_buttons-mixins.scss
```diff
@@ -32,85 +25,213 @@ $buttonActiveScale: scale(0.985);
...
-    box-shadow: $componentShadow;
-    transform: $buttonActiveScale;
+  &--full-width {
+    width: 100%;
   }
 
-  &:active:focus {
-    box-shadow: $buttonActiveBoxShadow;
+  &:focus {
+    outline: none;
```
matzimowski: not for now, but actually would be good to have some outline here in the future.

vergilius:
@burakukula considering larger size of buttons, do we know if that will introduce any inconsistencies to the interface? for example here?

Screenshot 2019-05-24 at 10 31 39

vergilius:
is there a plan for getting rid of those inconsistencies?

burakukula:
@vergilius It's not an inconsistency in terms of size, as these buttons are the same type... Anyway, in the tests the buttons on the right pane, in challenges changes:

Screenshot 2019-05-24 at 10 51 53

vergilius:
you are changing size for a button, how is that not an inconsistency? ;)

Screenshot 2019-05-24 at 10 59 34

src/components/buttons/ButtonSecondary.jsx
anna-wro: you have a typo here - mustRAd

burakukula:
@vergilius I'm sorry, I don't really understand what you are talking about? Inconsistency compare to what? Sizes of the buttons change, that's correct. But we change them globaly: primary buttons will have 40px height, seconary and secondary small 32px. Please explain more what do you mean?

vergilius:
- feed footer, icons size on left: 24px, button size on right: 24px
- change buttons to larger size
- different sizes of components ;)

burakukula:
@vergilius we are overviewing these changes with the design team, so yeah, it is taken into consideration and will be adjusted on the application side. Obviously.

vergilius:
Cool, thanks!

burakukula:
@vergilius no problem, sorry, it was not clear to me what do you mean at the beginning. Good catch 👀 Thx for feedback 🙂

PR #1582, opened 2019-05-28
Introduce button component

burakukula:
New button component with finalized naming, which will replace current `ButtonPrimary` and `ButtonSecondary`. Changes:

 - New button component with a small documentation
 - From now on we will have buttons related to the actions of the user more then the size of the button, like we did was in the past. So, we can use `primary`, `secondary` or`link-button` in 3 sizes: `large`, `normal` and `small`.
 - After talking with Pencils regarding the presentation of the buttons in SG, they suggest the following layout with type and state of the button presented. This required some changes in `DocsBlock` component. Maybe it could be refactored in the future, to allow us to use `Flex` for layout purposes in SG, but I don't consider it the most important part right now. (Documentation also could be extended by an example of a button with loading in the future).
 - I don't remove old buttons yet, as they have to be replaced by new buttons slowly, but please consider them as deprecated from now on.
 - Documentation could be extended, I guess, but I would do it as soon as the documentation on the Pencils side is ready, so we could adjust use cases. For now buttons README file we can treat as a starting point to the full documentation. 
- We can consider adding `aria-pressed` or `aria-expanded` props for buttons in the future, for now, I added just `role="button"` for non button elements (if the element is `<a>`)

**Proxima issue**
Proxima font is broken. This issue is known for a while, but in case of the component like buttons, is especially visible. It is impossible to align text inside the component, because of this issue:

Screenshot 2019-07-03 at 14 19 10
Screenshot 2019-07-03 at 14 19 19

I use a hack, for now, to move the text into the bottom of the component, but as soon as we get Proxima which is fixed or we switch to the system fonts for regular, we could remove the span from `<Button>`

ISSUE:
Screenshot 2019-07-03 at 14 47 10

FIXED:
Screenshot 2019-07-03 at 14 47 22

buttons-new-1 mov


src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,158 @@
...
+  if (props.href) {
+    TypeToRender = 'a';
+  }
+
+  return (
+    <TypeToRender
+      className={btnClass}
+      {...props}
+      disabled={disabled}
+      role={props.href ? 'button' : null}
```
matzimowski: 🥇

src/components/buttons/Button.spec.jsx
```diff
@@ -0,0 +1,82 @@
...
+});
+
+test('primary don\'t have small', () => {
+  const button = shallow(
+    <Button small>Some text</Button>
+  );
+
+  expect(button.hasClass('sg-button--small')).toEqual(false);
+});
+test('wide', () => {
```
matzimowski: wide

src/components/buttons/Button.spec.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import Button, {BUTTON_TYPE} from './Button';
+import {shallow} from 'enzyme';
+
+test('render', () => {
```
burakukula: It is the same convention as we use for other the specs, it might be rewritten in the future to `describe` and `it`.

src/components/buttons/pages/buttons.jsx
```diff
@@ -1,206 +1,205 @@
...
-      <br />
-      <br />
-      <ButtonPrimary
-        icon={<Icon type={iconTypes.FB} color={ICON_COLOR.ADAPTIVE} size={16} />}
-        buttonType={BUTTON_PRIMARY_TYPE.FB}
-      >
-        Login with Facebook
-      </ButtonPrimary>
-    </DocsBlock>
+function getValues(object, addUndefined = true) {
```
burakukula: this is used in other places, I will move it to some helper in a separate PR

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,159 @@
...
+    >
+      {ico}
+      {/* As soon as we have Proxima fixed, we could remove that span */}
+      <span className="sg-button__text">{children}</span>
```
burakukula: To align text in the middle, we need to move the text down, as 'Proxima' font is broken. As soon (if ever) we get Proxima which is fixed (and there is one like that available), we could remove the span, which is kind of hack right now. Look into PR description.

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,159 @@
+// @flow
```
PinkaminaDianePie: strict?

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,159 @@
...
+  | 'medium'
+  | 'small';
+
+export {BUTTON_SIZE, BUTTON_TYPE};
+
+export type ButtonPropsType = {
+  /**
+   * Children to be rendered inside Button
+   * @example <Button
+   *           icon={<Icon type={ICON_TYPE.ANSWER} color={ICON_COLOR.LIGHT} size={24} />}
```
PinkaminaDianePie: just `  <Button icon={<Icon type="answer" color="light" size={24} />} buttonType="facebook">` here and everywhere :)

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,159 @@
...
+   *          </Button>
+   */
+  href?: string,
+  /**
+   * Optional boolean for disabled button
+   * @example <Button type="primary-mint" disabled>
+   *            button
+   *          </Button>
+   */
+  disabled?: ?boolean,
```
PinkaminaDianePie: don't use nullable values, especially for booleans

src/components/buttons/pages/buttons-interactive.jsx
```diff
@@ -9,30 +8,7 @@ const Buttons = () => {
...
-    {
-      name: 'icon',
-      values: allIcons
-    },
-    {
-      name: 'href',
-      values: String
-    }
-  ];
+    allIcons[type] = <Icon type={ICON_TYPES[type]} color={ICON_COLOR.ADAPTIVE} size={24} />);
```
mpacholec: have you considered using sth like this ?
```js
const allIcons = Object.entries(ICON_TYPES).reduce((acc, [key, type]) => ({
  ...acc, 
  [key]: <Icon type={type} color={ICON_COLOR.ADAPTIVE} size={24} />
})
, {});
```

PinkaminaDianePie:
please add reexport  to index.js as well :)

src/docs/components/DocsBlock.jsx
```diff
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react';
 import type {Node} from 'react';
+import classnames from 'classnames';
```
mpacholec: what do you think about:

```jsx
import cx from 'classnames';
```
and later 
```jsx
className={cx('docs-block', {
  'docs-block--even-columns': evenColumns,
  'docs-block--full-width': fullWidth
})}
```

src/docs/components/DocsBlock.jsx
```diff
@@ -9,17 +10,52 @@ type PropsType = {
...
+    >
+      {children}
+    </ContentBlock>
+    {multiContent.map((extraChild, i) => (
+      <ContentBlock
+        toBottom={toBottom}
+        centered={centered}
+        spacedBetween={spacedBetween}
+        justified={justified}
+        key={i}
```
mpacholec: `key` as array's index ?

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,159 @@
+// @flow
+
+import * as React from 'react';
+import type {Node} from 'react';
```
mpacholec: if you're using `React.Node` you don't need this import

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,159 @@
...
+  | 'destructive'
+  | 'warning'
+  | 'facebook';
+
+type ButtonSizeType =
+  | 'large'
+  | 'medium'
+  | 'small';
+
+export {BUTTON_SIZE, BUTTON_TYPE};
```
maciekbb: Do we need that since we relay on flow?

src/docs/components/DocsBlock.jsx
burakukula: @mpacholec good idea, on my way.

src/components/buttons/Button.jsx
burakukula: @PinkaminaDianePie sure! on it.

src/docs/components/DocsBlock.jsx
maciekbb: I guess they never change/reorder

maciekbb: also it's in docs

src/components/buttons/README.md
```diff
@@ -0,0 +1,157 @@
+# Buttons
+
+Buttons are created to help user to make actions, choices and to move around the product.
+
+### General usage examples
+
+```jsx
+import Button from 'style-guide/src/components/buttons/Button';
```
PinkaminaDianePie: import Button from 'style-guide';

burakukula:
@maciekbb do you have anything in mind? I will take care of the migration to the new button component, I can do it slowly, as we have both versions of the components now. And since the naming has changed for the new component, I am not sure if I could do it fully automatically. But I have time in this sprint for that, so maybe it will not be such a long process...

src/components/buttons/Button.jsx
burakukula: hmm, I think, I would need to have a look at html sketchapp as well, as I export all the buttons using it. not sure if this change would not require changes in export logic as well. Maybe we can have a look on it together?

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,159 @@
...
+  icon,
+  fullWidth,
+  disabled,
+  children,
+  className,
+  ...props
+}: ButtonPropsType) => {
+  const btnClass = classNames(
+    'sg-button', {
+      [`sg-button--${size || ''}`]: size,
```
mpacholec: I believe that you don't need to write `${size || ''}`
```jsx
  [`sg-button--${size}`]: size
```
if size will be falsy classNames will handle it

src/docs/components/DocsBlock.jsx
mpacholec: 👍

src/components/buttons/pages/buttons.jsx
```diff
@@ -1,206 +1,134 @@
...
-    </DocsBlock>
+  getValues(BUTTON_TYPE, false).forEach(type => {
+    let iconColor;
 
-    <DocsBlock info="Primary round button">
-      <ButtonRound label="Add question">
-        <Icon type={iconTypes.PLUS} size={16} />
-      </ButtonRound>
-    </DocsBlock>
+    if (
```
mpacholec: I would extract this logic into separate function
```jsx
const getIconColor = (type) => {
  ...
}
```

src/components/buttons/Button.jsx
mpacholec: I don't believe that adding `flow strict` is better than `flow` and it shouldn't be treated as request for changes in entire PR...

PinkaminaDianePie: You can write untyped or untested code for your own features, but common modules which will be used by everyone should be done in a proper way. I don't ask for `//@ flow strict` in PRs for daily tasks. But this thing will be used by everyone and should be typed in a good way.

burakukula: @PinkaminaDianePie I will change it to `//@ flow strict`, as you requested. I really don't remember when, or if ever, I merged code, without solving feedback or comments in the PR, so I guess we could avoid `request changes` flag. But no worries, if you prefer this way.
Also, please keep in mind, that the goal for all of us is to write high quality code, so I think there is no reason to suggest otherwise.
Anyways, I'm on my way with changes, I'll ping you when I'm done. Thx a lot for feedback.

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,158 @@
...
+
+  if (props.href) {
+    TypeToRender = 'a';
+  }
+
+  return (
+    <TypeToRender
+      className={btnClass}
+      {...props}
+      disabled={disabled}
```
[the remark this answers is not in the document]
burakukula: @Icokie For me, this syntax is less readable. What would be the benefit of this change? Is the outcome the same?

src/components/buttons/pages/buttons-interactive.jsx
burakukula: @mpacholec yeah, seems like a better solution, I think. Thx.

src/components/buttons/Button.jsx
burakukula: @PinkaminaDianePie done - changed to `//@ flow strict`, I have made some adjustments for `icon` and `href`,  pls have a look if you will have a moment. thx.

src/components/buttons/Button.jsx
burakukula: done, I hope I didn't miss anything.

src/components/buttons/Button.jsx
burakukula: solved

src/docs/components/DocsBlock.jsx
burakukula: solved here and in `Button.jsx`. I guess we could extend this pattern to other components in the future.

src/components/buttons/Button.jsx
burakukula: I have changed it to `import React from 'react';` as it was used this way in other places. hope that's fine.

src/components/buttons/Button.jsx
PinkaminaDianePie: @burakukula actually request changes was not because of `strict` but because of documentation - it's more important than `strict`. We introduced JSX with literals and without importing constants, we don't event reexport any constants from `index.js`. If we want to be consistent we shouldn't use the old way in the documentation. Otherwise, we will end up with both approaches in the same codebase forever. Please forgive me if `request changes` looked offending for you - I won't use it next time :) My intention was not to block PR from merging but to stress out that issues I mentioned are really important. But next time I will express it in words, not in some flags :)

src/components/buttons/README.md
burakukula: done

src/components/buttons/Button.jsx
burakukula: Totally, thx. Fixed.

src/components/buttons/Button.jsx
PinkaminaDianePie: only one last place in Readme, `size={BUTTON_SIZE.SMALL}`

src/components/buttons/Button.jsx
PinkaminaDianePie: unfortunately if we use `React.Node`, React should be imported as `import * as React from 'react';`, otherwise JSX autocompletion in VSCode won't work :(

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,171 @@
...
+  );
+
+  const iconClass = cx(
+    'sg-button__icon', {
+      [`sg-button__icon--${size || ''}`]: size
+    },
+    className
+  );
+
+  const ico = icon !== undefined ? <span className={iconClass}>{icon}</span> : null;
```
PinkaminaDianePie: we should check for `null` as well i believe, otherwise `span` will be rendered as well (`Node` type is nullable by default, so it's still possible to pass `null` there)

src/components/buttons/Button.jsx
```diff
@@ -0,0 +1,171 @@
...
+  const TypeToRender = href !== undefined ? 'a' : 'button';
+
+  return (
+    <TypeToRender
+      className={btnClass}
+      {...props}
+      disabled={disabled}
+      role={href !== null ? 'button' : undefined}
```
PinkaminaDianePie: `href` can't be `null`, it's a `string | undefined`

src/components/buttons/Button.jsx
burakukula: @PinkaminaDianePie 👍 no worries, all good.

src/components/buttons/Button.jsx
burakukula: @PinkaminaDianePie soooorrry, I am blind. should be fixed now. thx 🙂

src/components/buttons/Button.jsx
PinkaminaDianePie: we checked it and in the current case, it works with both styles of imports, the problem I mentioned was about something different I believe, no actions needed about it.

src/components/buttons/_buttons.scss
```diff
@@ -0,0 +1,165 @@
...
+    &--small {
+      margin-right: spacing(xs);
+    }
+
+    &--large {
+      margin-right: spacing(s);
+    }
+  }
+
+  &:focus {
```
ElaLukaszewicz: for a11y I would recommend not to add outline: none, here is why: https://developers.google.com/web/fundamentals/accessibility/how-to-review?utm_source=lighthouse&utm_medium=extension -> "...It’s common practice to disable focus styles without providing an alternative by using outline: none in CSS, but this is an anti-pattern. If a keyboard user can’t see what’s focused, they have no way of interacting with the page." wdyt @burakukula ? :)

burakukula: @ElaLukaszewicz sure thing! good catch 👍

PR #1587, opened 2019-07-01
Make basic icons sprite available as svg file

matzimowski:
in short: make svg sprite of basic icons available as a static svg file and test it.

it's not publicly available yet.

Screenshot 2019-07-02 at 08 46 22

i'd like to add it and test how it could work with our client apps.

it's related to a research "how to include sg icons differently in our apps" and it allows to do sth like:
```
<svg class="sg-icon">
    <use href="STYLEGUIDE_PATH/icons.svg#icon-search"></use>
</svg>
```
pros:
- no need to inject svg sprite in JS
- nodes of all icons are not included in DOM
- svg sprite is cached

to support IE we'd need to use svg4everybody https://www.npmjs.com/package/svg4everybody

### DIFF
--- a/src/docs/pages/index.jsx
+++ b/src/docs/pages/index.jsx
@@ -151,6 +151,11 @@ const index = () => (
       </a>.
     </p>
 
+    <input
+      type="hidden"
+      value="https://styleguide.brainly.com/%%images/icons.svg%%"
+    />
+
     <script dangerouslySetInnerHTML={{__html: 'hljs.initHighlightingOnLoad();'}} />
   </ArticlePage>
 );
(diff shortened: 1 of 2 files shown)


matzimowski:
the issue here is the lack of support of IE. It works since Edge 13, some polyfill would be needed

src/docs/pages/index.jsx
```diff
@@ -151,6 +151,11 @@ const index = () => (
       </a>.
     </p>
 
+    <input
+      type="hidden"
+      value="https://styleguide.brainly.com/%%images/icons.svg%%"
+    />
```
tomasz-sodzawiczny: What is this `input` for? 🤔

matzimowski: it could be any hidden element, i need this value to be available somehow to use it later on.

tomasz-sodzawiczny: Ah, it's just to know the url?

How are you going to test it? Does it need to be served from style-guide? If it's just local testing it sounds like it could be also locally served.

PR #1588, opened 2019-07-02
added index file

PinkaminaDianePie:
allows us to replace this:
image
by this:
image


PinkaminaDianePie:
done, from now on SG is tree-shakable :D

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Brainly Front-End Style Guide",
   "author": "Brainly",
   "private": true,
+  "main": "src/index.js",
```
tomasz-sodzawiczny: add `module` as well

src/index.js
```diff
@@ -0,0 +1,70 @@
+// @flow
+export {default as ActionList} from './components/action-list/ActionList';
```
tomasz-sodzawiczny: What about all other exports?

PinkaminaDianePie: We don't need to import constants anymore. In all new code, we just use literals instead. Since this `index.js` will be probably used in new code, it's better not to re-export them

tomasz-sodzawiczny: Are we plannig to run a codemod to remove all the old usages?

package.json
PinkaminaDianePie: replaced `main` by `module`

src/index.js
PinkaminaDianePie: i'd like to do it eventually, but not sure how time-consuming will it be

PinkaminaDianePie:
@maciekbb we can't until we create this codemod, because any possible lint will complain about old files. We can just ask people to use a new way, but in a meantime, we need to create some codemod. It's an only option i believe

PR #1591, opened 2019-07-05
marked all strict-compatible modules as strict

PinkaminaDianePie:
these modules was already `flow strict` compatible so no changes were required


PinkaminaDianePie:
sure, i'll wait till next week :)

PinkaminaDianePie:
It works vise versa. It could solve some issues in strict files in the main app, but wont affect non-strict files

PR #1596, opened 2019-07-12
cover last components by flow strict

PinkaminaDianePie:


src/components/form-elements/Checkbox.jsx
```diff
@@ -12,15 +12,15 @@ export type CheckboxPropsType = {
...
+  id: string
 };
 
 class Checkbox extends PureComponent<CheckboxPropsType, CheckboxStateType> {
   constructor(props: CheckboxPropsType) {
     super(props);
 
     this.state = {
-      id: props.id || generateRandomString()
+      id: props.id === undefined || props.id === '' ? generateRandomString() : props.id
```
maciekbb: it cannot be undefined

src/components/form-elements/Textarea.jsx
```diff
@@ -17,9 +17,10 @@ export const SIZE = {
 };
 
 export type TextareaPropsType = {
+  // $FlowFixMe any generic prop types here broke autocomplete, so let's leave it as is for now
```
maciekbb: this is an argument to switch to typescript

my vscode almost never proviedes autocomplete

src/components/form-elements/Checkbox.jsx
PinkaminaDianePie: `id?: string,`

src/components/form-elements/Textarea.jsx
PinkaminaDianePie: 1) your vscode doesn't provide autocomplete because of typescript :) it just hangs on big projects such as ours and because of poor vscode architecture it doesn't show autocomplete from anything else if TS is broken. Check their github, they have tons of such issues.
2) code should be not only typed, but typechecked. It's a main purpose of type system. Typescript can't check type usage, it can check only type definitions. Like eslint, but for types. Type system based not on graphs doesn't provide any runtime guarantees, so it doesn't work by definition.

src/components/icon-as-button/IconAsButton.jsx
```diff
@@ -31,27 +30,26 @@ type IconAsButtonPropsType = {
   size?: IconSizeType,
   color?: IconModule.IconColorType,
   type?: ?IconModule.IconTypeType,
-  children?: ?Element<any>,
+  children?: ?Element<*>,
   border?: boolean,
   action?: boolean,
   transparent?: boolean,
   active?: boolean,
-  overlay?: ?Node,
```
burakukula: 🎉

src/components/icon-as-button/IconAsButton.jsx
```diff
@@ -1,9 +1,8 @@
-// @flow
+// @flow strict
 import React from 'react';
```
burakukula: @PinkaminaDianePie I am not sure, but I think we have to ways of importing react `import * as React from 'react';`. Maybe we should unify it?

PinkaminaDianePie: Yeah, i'd like to unify it as `import * as React from 'react';`. Maybe in next PR :)

PR #1599, opened 2019-07-18
Introduce new set of icons with STD prefix

burakukula:
Introducing a new set of icons with `std-` prefix:

 - Icons can be used with `std-` prefix like following:
  JSX: `<Icon type="std-answer" size="46" color="blue" />`
  HTML: `<svg class="sg-icon sg-icon--x46 sg-icon--blue"><use xlink:href="#icon-std-answer"></use></svg>`
 - Icons have  `viewBox="0 0 512 512"` as decided with the team.
 - Icons have "safe space" inside, so visually they are slightly smaller than the current once, **keep that in mind during implementation** - I would recommend replacing icons by feature, but in the way, that we don't have a situation that old and new icons are next to each other, e.g - change whole topbar icons or icons in the editor (I will be working on this last one). 
 - I will introduce this set for logged-in user only, for now. On QPage for not logged, I recommend change **all the icons at once to not attach both sets to this page**.

Both sets are presented for now in SG, with the comment that the first one is deprecated:

icons-set mov


src/components/icons/Icon.jsx
```diff
@@ -87,9 +145,71 @@ export type IconSizeType =
...
+  STD_TWITTER: 'std-twitter', // new
+  STD_UNDERLINED: 'std-underlined',
+  STD_VERIFIED: 'std-verified',
+  STD_YOUTUBE: 'std-youtube'
+};
+
 export const TYPE = {
   ANSWER: 'answer',
-  ANSWERED: 'answered',
+  ANSWERED: 'answered', // not needed in new set
```
burakukula: types with comments would be removed separately.

aju:
@burakukula how big is new set (kb) ?

burakukula:
@aju 
Screenshot 2019-07-19 at 09 54 48

We don't want to have both sets on the page for a long time, it's important to change it quickly, but the whole team need's to get involved to do it. I hope we will manage fast.

src/components/icons/pages/icons.jsx
```diff
@@ -1,12 +1,14 @@
...
 import ContrastBox from 'components/ContrastBox';
-import Icon, {TYPE, ICON_COLOR, SIZE} from '../Icon';
+import Icon, {TYPE, STD_TYPE, ICON_COLOR, SIZE} from '../Icon';
+import Text from '../../text/Text';
 
 const icons = () => (
   <div>
     <DocsBlock>
       <ContrastBox>
+        <Text size="small">This set of the icons is deprecated</Text>
```
matzimowski: hm, maybe we could make it a bit more visible, wdyt?

matzimowski: maybe just a white color since background is dark?

burakukula: Done, I move it above the `ContrastBox` and change the color of the text, now it's peach and on the white background.

PR #1601, opened 2019-07-20
Remove margin override

### LINKED ISSUE #1600: Action list - to right, missing spacing
Hi!

Action-list component has `--to-right` modifier but it does not contains/respects the `__hole`'s  `--no-spacing` or `--spaced-xsmall` or `--spaced-small` or `--spaced-large` modifiers. It should have more control over spacing modifiers.

https://github.com/brainly/style-guide/blob/87a724e84a7784a9799d907f75d44346e8ea6f32/src/components/action-list/action-list.scss#L89-L100

seahindeniz:
This change solves the issue #1600 but it's not respects the way of angle of adding spaces.

Before:
image

After:
image

---
In order to respect `--to-right` modifier and adding spaces with `margin-left` instead of `margin-right` would may require some additional lines from L14-L24 to L91-L++ (like copy paste with a small change of moving the gutter fn to param 4th of the margin prop).
https://github.com/brainly/style-guide/blob/87a724e84a7784a9799d907f75d44346e8ea6f32/src/components/action-list/action-list.scss#L14-L24


seahindeniz:
Bumped and I'll try the `Flex` c. Thanks for the suggestion @burakukula 🍻

PR #1602, opened 2019-07-23
basic webpack build for interactive pages

PinkaminaDianePie:
it currently builds only interactive pages. will add others in separate PR. it should be compatible with current builds and work separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "sass-lint": "^1.10.2",
     "sass-loader": "^7.1.0",
     "style-loader": "^0.23.1",
+    "svg-sprite-loader": "^4.1.6",
     "terser-webpack-plugin": "^1.2.3",
     "through2": "^2.0.3",
     "webpack": "^4.6.0",
@@ -88,8 +89,7 @@
     }
   },
   "dependencies": {
-    "classnames": "^2.2.5",
-    "svg-sprite-loader": "^4.1.6"
+    "classnames": "^2.2.5"
   },
   "sideEffects": false
 }
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "watch": "gulp build && gulp watch & yarn start",
     "test": "yarn run scss-lint && yarn run scss-unused-variables && yarn flow && yarn run js-lint && jest",
     "eslint": "./node_modules/frontend-tools-configs/node_modules/.bin/eslint",
-    "beta:watch": "webpack-dev-server --hot --inline --open"
+    "beta:watch": "webpack-dev-server --hot --inline --open --watch --mode=development"
   },
   "jest": {
     "setupFiles": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "146.0.0",
+  "version": "146.0.1",
   "description": "Brainly Front-End Style Guide",
   "author": "Brainly",
   "private": true,
--- a/package.json
+++ b/package.json
@@ -12,12 +12,15 @@
     "@babel/plugin-transform-flow-strip-types": "^7.2.3",
     "@babel/preset-env": "^7.3.1",
     "@babel/preset-react": "^7.0.0",
+    "@hot-loader/react-dom": "^16.8.6",
     "babel-eslint": "^7.2.3",
     "babel-jest": "^20.0.3",
     "babel-loader": "^8.0.5",
+    "css-loader": "^3.1.0",
     "del": "^2.2.0",
     "enzyme": "^3.1.0",
     "enzyme-adapter-react-16": "^1.0.1",
+    "file-loader": "^4.1.0",
     "flow-bin": "^0.102.0",
     "frontend-tools-configs": "git+https://github.com/brainly/frontend-tools-configs.git#v16.2.7",
     "gulp": "^3.9.0",
@@ -35,19 +38,26 @@
     "gulp-sourcemaps": "^2.6.0",
     "gulp-svg-sprite": "^1.2.7",
     "gulp-util": "^3.0.6",
+    "html-webpack-plugin": "^3.2.0",
     "http-server": "^0.9.0",
     "husky": "^0.11.3",
     "jest": "^24.0.0",
+    "postcss-loader": "^3.0.0",
     "query-string": "^6.5.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
+    "react-hot-loader": "^4.12.8",
     "react-test-renderer": "^16.2.0",
     "run-sequence": "^1.1.1",
     "s3": "^4.4.0",
     "sass-lint": "^1.10.2",
+    "sass-loader": "^7.1.0",
+    "style-loader": "^0.23.1",
     "terser-webpack-plugin": "^1.2.3",
     "through2": "^2.0.3",
     "webpack": "^4.6.0",
+    "webpack-cli": "^3.3.6",
+    "webpack-dev-server": "^3.7.2",
     "yargs": "^4.3.1"
   },
   "scripts": {
@@ -61,7 +71,8 @@
     "postversion": "git push && git push --tags",
     "watch": "gulp build && gulp watch & yarn start",
     "test": "yarn run scss-lint && yarn run scss-unused-variables && yarn flow && yarn run js-lint && jest",
-    "eslint": "./node_modules/frontend-tools-configs/node_modules/.bin/eslint"
+    "eslint": "./node_modules/frontend-tools-configs/node_modules/.bin/eslint",
+    "beta:watch": "webpack-dev-server --hot --inline --open"
   },
   "jest": {
     "setupFiles": [
@@ -77,7 +88,8 @@
     }
   },
   "dependencies": {
-    "classnames": "^2.2.5"
+    "classnames": "^2.2.5",
+    "svg-sprite-loader": "^4.1.6"
   },
   "sideEffects": false
 }
--- a/src/docs/components/CodeBlock.jsx
+++ b/src/docs/components/CodeBlock.jsx
@@ -42,7 +42,7 @@ class CodeBlock extends Component<PropsType> {
       throw new Error('Passed child is not a string.');
     }
 
-    let markup = html_beautify(children, { // eslint-disable-line no-undef
+    let markup = html_beautify(children, {
       indent_size: 2,
       unformatted: [],
       wrap_line_length: 0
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -77,7 +88,8 @@
     }
   },
   "dependencies": {
-    "classnames": "^2.2.5"
+    "classnames": "^2.2.5",
+    "svg-sprite-loader": "^4.1.6"
```
maciekbb: in dev deps?

src/docs/components/CodeBlock.jsx
```diff
@@ -45,16 +41,17 @@ class CodeBlock extends Component<PropsType> {
     if (typeof children !== 'string') {
       throw new Error('Passed child is not a string.');
     }
-
-    let markup = beautify(children, {
+    // $FlowFixMe
+    let markup = html_beautify(children, { // eslint-disable-line no-undef
```
maciekbb: mark it as global instead?

src/docs/components/CodeBlock.jsx
```diff
@@ -45,16 +41,17 @@ class CodeBlock extends Component<PropsType> {
...
       indent_size: 2,
       unformatted: [],
       wrap_line_length: 0
     });
 
     if (type === 'jsx') {
       //HACK <i> was added to force highlightJS to highlight first tag
-      markup = hljs.highlight('jsx', `<i>${markup}</i>`).value;
+      // $FlowFixMe
+      markup = hljs.highlight('jsx', `<i>${markup}</i>`).value; // eslint-disable-line no-undef
```
maciekbb: mark it as global instead?

src/docs/components/CodeBlock.jsx
```diff
@@ -63,7 +60,8 @@ class CodeBlock extends Component<PropsType> {
         .replace('<span class="hljs-tag">&lt;/<span class="hljs-name">i</span>&gt;</span>', '');
 
     } else {
-      markup = hljs.highlight(type, markup).value;
+      // $FlowFixMe
+      markup = hljs.highlight(type, markup).value; // eslint-disable-line no-undef
```
maciekbb: mark it as global instead?

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+          },
+          use: [
+            'style-loader',
+            'css-loader',
+            {
+              loader: 'postcss-loader',
+              options: {
+                ident: 'postcss',
+                plugins: [
+                  require('autoprefixer')({
```
maciekbb: probably we don't need it :D

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+const pkg = require('./package');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+const IS_PRODUCTION = Boolean(argv.production);
+const VERSION = argv.production ? pkg.version : 'dev';
+const SOURCE_DIR = path.join(__dirname, 'src');
+const DIST_DIR = path.join(__dirname, 'dist');
+const VERSIONED_DIST_DIR = path.join(DIST_DIR, VERSION);
+const SOURCE_DOCS_DIR = path.join(SOURCE_DIR, 'docs');
+const SOURCE_COMPONENTS_DIR = path.join(SOURCE_DIR, 'components');
```
maciekbb: I'd create a separate config file with all the paths

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+      }
+    },
+    plugins: [
+      new webpack.DefinePlugin({
+        'process.env.NODE_ENV': JSON.stringify(IS_PRODUCTION ? 'production' : 'development')
+      }), new HtmlWebpackPlugin({
+        template: path.join(SOURCE_DOCS_DIR, 'index.html')
+      })
+    ],
+    devtool: IS_PRODUCTION ? 'source-map' : 'eval',
```
maciekbb: We can have source maps on prod

webpack.config.js
PinkaminaDianePie: we need it, some rules we use works under prefix even in chrome, so ui is broken without it

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+    optimization: {
```
maciekbb: Do we need optimization for production? I'd rather keep in consistent with local env

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+      'hljs': 'hljs'
+    },
+    watch: true,
+    mode: 'development',
+    entry: {
+      [path.basename(jsEntry)]: jsEntry
+    },
+    output: {
+      filename: '[name].bundle.js',
+      path: jsOutput,
```
maciekbb: `config.jsOutput` or `config.distPath`

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+
+const IS_PRODUCTION = Boolean(argv.production);
+const VERSION = argv.production ? pkg.version : 'dev';
+const SOURCE_DIR = path.join(__dirname, 'src');
+const DIST_DIR = path.join(__dirname, 'dist');
+const VERSIONED_DIST_DIR = path.join(DIST_DIR, VERSION);
+const SOURCE_DOCS_DIR = path.join(SOURCE_DIR, 'docs');
+const SOURCE_COMPONENTS_DIR = path.join(SOURCE_DIR, 'components');
+
+const babelEnv = params => [
```
maciekbb: I'd just inline it

src/docs/webpackEntry.jsx
```diff
@@ -0,0 +1,22 @@
+import {hot} from 'react-hot-loader/root';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import '../main.scss';
+import './_sass/main.scss';
+import Entry from './pages/interactive';
+import {sections} from './page-components';
+
+const requireAll = r => {
```
maciekbb: This function does not require anything

src/docs/webpackEntry.jsx
```diff
@@ -0,0 +1,22 @@
...
+  r.keys().forEach(r);
+};
+
+requireAll(require.context('../images/', true, /\.svg$/));
```
maciekbb: What does it do? :D

PinkaminaDianePie: it's a weird way in webpack to load assets recursively, more details here: https://github.com/webpack/docs/wiki/context#context-module-api

src/docs/webpackEntry.jsx
PinkaminaDianePie: copy-paste from webpack documentation

webpack.config.js
PinkaminaDianePie: i use it in 2 places here and possibly will use in more places when we will introduce proper builds

webpack.config.js
PinkaminaDianePie: it's what we already have in the main config right now. lets leave it as is, because i wanna introduce proper build step some day and we will need it anyway

webpack.config.js
PinkaminaDianePie: i found that it's quite hard to reason about paths in our main app, when our configs are split into different files

package.json
PinkaminaDianePie: done

src/docs/components/CodeBlock.jsx
PinkaminaDianePie: done

src/docs/components/CodeBlock.jsx
PinkaminaDianePie: done

src/docs/components/CodeBlock.jsx
PinkaminaDianePie: done

webpack.config.js
PinkaminaDianePie: moved to other constants

webpack.config.js
maciekbb: so maybe 
```
const config = {
  ...
}
```

webpack.config.js
maciekbb: Yeah for build step I agree it's necessary, however for serving standalone styleguide not that much

webpack.config.js
PinkaminaDianePie: config for config seems weird, normal constants would be enough :D

webpack.config.js
tomasz-sodzawiczny: why not babelrc?

webpack.config.js
tomasz-sodzawiczny: My experience is that this `path.join('src', 'docs')` is an overkill and actually makes working with the config way harder. In this case I would just go for strings `'src/docs/'`.

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+        }
+      ]
+    },
+    externals: {
+      'html_beautify': {
+        root: 'html_beautify'
+      },
+      'hljs': 'hljs'
+    },
+    watch: true,
```
tomasz-sodzawiczny: Maybe use `--watch` explicitly.

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+    mode: 'development',
```
tomasz-sodzawiczny: Why hardcoding mode?

webpack.config.js
tomasz-sodzawiczny: Do we need them at all? 🤷‍♂

webpack.config.js
```diff
@@ -0,0 +1,155 @@
...
+        SOURCE_DOCS_DIR,
+        'node_modules'
+      ],
+      alias: {
+        'react-dom': '@hot-loader/react-dom'
+      }
+    },
+    plugins: [
+      new webpack.DefinePlugin({
+        'process.env.NODE_ENV': JSON.stringify(IS_PRODUCTION ? 'production' : 'development')
```
tomasz-sodzawiczny: Why, what for? webpack-cli will set it.

webpack.config.js
tomasz-sodzawiczny: But why custom config? Just use mode.

webpack.config.js
PinkaminaDianePie: done

webpack.config.js
PinkaminaDianePie: fixed

webpack.config.js
PinkaminaDianePie: copy-paste from old webpack for gulp, which had no webpack-cli :D fixed!

PR #1603, opened 2019-07-24
Add std-icons to exported icon type

matzimowski:
to make it possible to use it i.e. in `Label`

```
import Label, {ICON_TYPE} from 'style-guide/src/components/labels/Label';

<Label
    iconType={ICON_TYPE.STD_HEART}
/>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "style-guide",
-  "version": "145.2.0",
+  "version": "145.2.1",
   "description": "Brainly Front-End Style Guide",
   "author": "Brainly",
   "private": true,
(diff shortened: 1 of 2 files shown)


matzimowski:
@vergilius to other way would be to export it from Label, IconAsButton as so on, wdyt?

vergilius:
I don't have enough expertise to tell you how to do it properly, just noticed exporting two separate (though it may be the right way :O) things under one name

matzimowski:
hm, not sure what's the best way to do it here. but since `IconTypeType` accepts both TYPE and STD_TYPE it looked good to me.
anyway, it's an intermediate step, later on STD_ICON will replace renamed to ICON

maciekbb:
Don't we prefer to relay on flow type instead of importing consts?

matzimowski:
@maciekbb we do and we will, but 90% of our codebase doesn't use it and it's much easier to refactor it still using const.

PR #1608, opened 2019-07-29
added webpack dev build for other pages

PinkaminaDianePie:
mostly work for all pages, but some stuff still TBD:
- iframe pages doesn't work
- popup with source code for non-interactive pages
- search in the header

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,7 +11,8 @@
     "./node_modules/frontend-tools-configs/eslint/flow.js"
   ],
   "rules": {
-    "max-len": ["error", { "code": 120, "ignoreComments": true }]
+    "max-len": ["error", { "code": 120, "ignoreComments": true }],
+    "react/no-multi-comp": "off"
   },
   "globals": {
     "hljs": "readonly",
--- a/src/docs/layouts/IndexPage.jsx
+++ b/src/docs/layouts/IndexPage.jsx
@@ -1,0 +1,181 @@
+// @flow
+import React from 'react';
+import packageJSON from '../../../package.json';
+import ContentBox from '../../components/content-box/ContentBox';
+import ContentBoxHeader from '../../components/content-box/ContentBoxHeader';
+import ContentBoxContent from '../../components/content-box/ContentBoxContent';
+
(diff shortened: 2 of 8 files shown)


src/docs/layouts/IndexPage.jsx
```diff
@@ -0,0 +1,181 @@
+// @flow
```
PinkaminaDianePie: mostly copy-paste of original index, with some changes needed for SPA

src/docs/layouts/NavigationSPA.jsx
```diff
@@ -0,0 +1,57 @@
+// @flow
```
PinkaminaDianePie: mostly copy-paste of original navigation, but with react-router instead of normal links

.eslintrc
```diff
@@ -11,7 +11,8 @@
     "./node_modules/frontend-tools-configs/eslint/flow.js"
   ],
   "rules": {
-    "max-len": ["error", { "code": 120, "ignoreComments": true }]
+    "max-len": ["error", { "code": 120, "ignoreComments": true }],
+    "react/no-multi-comp": "off"
```
PinkaminaDianePie: our configs and eslint versions are so old, that bumping should be performed in a separate PR. in our latest configs i believe we don't have `react/no-multi-comp` rule anymore (or at least allow us to use arrows in JSX, so after bump of deps this could be removed)

src/docs/webpackEntry.jsx
```diff
@@ -3,20 +3,82 @@ import React from 'react';
...
+const version = packageJSON.version;
 
 const requireAll = r => {
   r.keys().forEach(r);
 };
 
 requireAll(require.context('../images/', true, /\.svg$/));
 
-const Wrapped = hot(Entry);
+const GeneratePage = ({padeIndex}) => navigation[padeIndex].elements.map(({name, component}) => (
```
maciekbb: I'd split it into few lines :P

PinkaminaDianePie: done

Issue #1615, opened 2019-08-13
SearchComponent "X" button added by IE by default


yutro:
on IE 11 and < browser adds the "X" button to input automatically, from time to time we need to disable it. (solution to do it through CSS) right now it's impossible cause ClassName prop passed to Input wrapper, not to Input itself.

![Screenshot from 2019-08-13 10-04-12](https://user-images.githubusercontent.com/11993787/62925066-cac0ba00-bdb1-11e9-9e0f-a3ae401cb96a.png)

PR #1616, opened 2019-08-13
SG-1615 -- inputClass name prop

Icokie:
Add the ability to add CSS class directly to Search input

https://github.com/brainly/style-guide/issues/1615

### DIFF
--- a/src/components/form-elements/Input.jsx
+++ b/src/components/form-elements/Input.jsx
@@ -70,17 +70,17 @@ export const COLOR = Object.freeze({
 });
 
 export type InputPropsType = {
-  +type?: ?InputType,
-  +value?: string | number,
-  +size?: ?InputSizeType,
-  +color?: ?InputColorType,
-  +valid?: boolean,
-  +invalid?: boolean,
-  +fullWidth?: boolean,
-  +noBorder?: boolean,
-  +withIcon?: boolean,
-  +className?: string,
-  +setInputRef?: (ref: ?HTMLElement) => mixed
+  type?: ?InputType,
+  value?: string | number,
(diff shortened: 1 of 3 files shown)


src/components/form-elements/Input.jsx
```diff
@@ -70,17 +70,17 @@ export const COLOR = Object.freeze({
...
-  size?: ?InputSizeType,
-  color?: ?InputColorType,
-  valid?: boolean,
-  invalid?: boolean,
-  fullWidth?: boolean,
-  noBorder?: boolean,
-  withIcon?: boolean,
-  className?: string,
-  setInputRef?: (ref: ?HTMLElement) => mixed
+  +type?: ?InputType,
```
PinkaminaDianePie: it will break autocomplete in VSCode. we need to bump to 0.105 before using proper exact/readonly, before that keep objects as it was before

Icokie: updated, damn vscode :(

src/components/search/Search.jsx
```diff
@@ -8,18 +8,22 @@ const {default: Input, COLOR, SIZE} = InputModule;
 
 type PropsType = {
   adaptiveIco?: boolean,
-  className?: string,
-  ...InputModule.InputPropsType
-};
+  inputClassName?: string
+} & InputModule.InputPropsType; // TODO: make back to spread (...InputModule.InputPropsType) after flow bump
```
aju: why after? It was working before

Icokie: we investigate it with @PinkaminaDianePie it doesn't see inputClassName prop, strange behavior

PR #1619, opened 2019-08-15
Fix for alignment of text of Label component

seahindeniz:
Hi

The Proxima font has an issue and it causes alignment problem to the Label component. So to avoid, this change should be considered like the `sg-button__text`.
https://github.com/brainly/style-guide/blob/febcf2c37513f46a25e3409bb0f8437796d33004/src/components/buttons/_buttons.scss#L34-L39

Please see the images below
You can count pixels if you want and this is the problem when there is no fix.

Before:
image

After: 
image

excuse me for the pixels :slightly_smiling_face:


src/components/labels/_labels.scss
```diff
@@ -40,6 +40,10 @@ $includeHtml: false !default;
       flex-shrink: 0;
       align-items: center;
       margin-right: gutter(0.25);
+      
+      // To fix alignment issue caused by Proxima font
+      position: relative;
```
burakukula: @seahindeniz 👍 but maybe we should actually add this fix to text in label, as it's Proxima issue, so it's related to text, not to icon? WDYT?

seahindeniz: Oops, sorry. I thought I was added into text. Text @burakukula

burakukula:
@seahindeniz sorry it took so long, your changes will be included in the next release of the library. BTW, we change a little the deployment flow, we gonna have bigger releases from now on. for more info you can have a look https://github.com/brainly/style-guide/blob/master/CONTRIBUTING.md or let us know.

PR #1620, opened 2019-08-16
Adjust and add icons to a new set

matzimowski:
**Adjusted icons:**

- std-facebook
- std-twitter
- std-medium
- std-youtube
- std-linkedin
- std-messages
- std-notifications
- std-counter
- std-friends

**Added icons:**

- std-logout
- std-profile_view
- std-question
- std-reload
- std-rotate
- std-settings 

Screenshot 2019-08-19 at 09 06 23


src/components/icons/Icon.jsx
```diff
@@ -195,10 +202,15 @@ export const STD_TYPE = {
...
   STD_PROFILE: 'std-profile', // new
+  STD_PROFILE_VIEW: 'std-profile_view',
+  STD_QUESTION: 'std-question',
   STD_RECENT_QUESTIONS: 'std-recent_questions', // new
+  STD_RELOAD: 'std-reload',
   STD_REPORT_FLAG: 'std-report_flag',
+  STD_ROTATE: 'std-rotate', //new
   STD_SEARCH: 'std-search',
   STD_SEEN: 'std-seen', // new
+  STD_SETTINGS: 'std-settings', // new
```
vergilius: why do we need this `// new` comment?

matzimowski: just for our purposes, (it's new, it means we've never had it before). it will be removed when legacy icons are removed after making a switch.

vergilius: it's better to add meaningful comments, even if that means they will be a bit longer

seahindeniz:
Hey @matzimowski 

I thinkt the `std-reload` icon doesn't seem right

image

matzimowski:
@seahindeniz that's a great catch! 👁 thanks!
my fix is on the way...

matzimowski:
actually the shape has to be changed (we are removing all `fill` props while generating svgs).
i will get back to it when I receive an updated icon

matzimowski:
`std-reload` 

Screenshot 2019-08-19 at 09 06 13
has been updated

PR #1626, opened 2019-08-22
Support multiple selection for Select component

seahindeniz:
Select component has failing when it meets with multiple attribute so added support for to not have rendering issues.

image


burakukula:
@seahindeniz sorry for the delay with the response. 
Regarding your PR, I am not sure if we actually need this component from the product perspective. Is it anyhow a blocker for you?

@aju wdyt about it?

seahindeniz:
> @seahindeniz sorry for the delay with the response.
> Regarding your PR, I am not sure if we actually need this component from the product perspective. Is it anyhow a blocker for you?
> 
> @aju wdyt about it?

Hi @burakukula no worries on delay.

I've added this PR to support for selecting multiple options in the select element. I know there are several ways to create an element that supports selecting multiple options but the point and my humble suggestion is SG's select component should have support for multiple select option in case of need instead of having an ugly result. 

Right now, this is how it looks when the select element has the `multiple` attribute without having additional modifiers and the result is not quite what is expected, if you agree
image

matzimowski:
IMHO, we need a design team's input here, since we do not have in our app.
i know it's extending the existing component but from design perspective it could be a completely new component.

burakukula:
Hey @seahindeniz! We are currently in the process of refactoring and total change of form elements. Selects will be changed soon somehow. I am not sure if we will create multiple selections selects, but I would suggest to close this PR and open an issue for that purpose. WDYT?

PR #1627, opened 2019-08-23
Fix for doc page of the action list component

seahindeniz:
ActionList component has miss-configured so this PR fixes

**ActionList toTop**

Before:
image
After:
image

---
**ActionListHole noShrik**

Before:
image
After:
image


matzimowski:
thanks for fixing the doc.

could you separate it from `Select` component changes, please?

PR #1629, opened 2019-08-27
Use std-icons in all examples

matzimowski:


src/components/labels/pages/labels-interactive.jsx
```diff
@@ -48,16 +48,15 @@ const Labels = () => {
...
         />
       </DocsActiveBlock>
       <DocsActiveBlock settings={settings}>
         <Label
-          text="Comment"
-          number={21}
+          text="Edit"
           size={SIZE.LARGE}
-          iconType={ICON_TYPE.COMMENT}
+          iconType={ICON_TYPE.STD_PENCIL}
```
matzimowski: there is no `comment` icon in a new set

PR #1632, opened 2019-08-28
Remove deprecated icons

matzimowski:
Screenshot 2019-08-28 at 09 51 36

the next step is to remove `std-` prefix from all icons


matzimowski:
the check will pass after merging this: https://github.com/brainly/style-guide/pull/1633

PR #1633, opened 2019-08-28
Use std-check icon in checkbox

matzimowski:
before:
Screenshot 2019-08-28 at 10 11 42

after:
Screenshot 2019-08-28 at 10 11 36

### DIFF
--- a/src/components/form-elements/Checkbox.jsx
+++ b/src/components/form-elements/Checkbox.jsx
@@ -3,7 +3,7 @@
 import classNames from 'classnames';
 import React, {PureComponent} from 'react';
 import generateRandomString from '../../js/generateRandomString';
-import Icon, {ICON_COLOR, TYPE} from '../icons/Icon';
+import Icon, {ICON_COLOR} from '../icons/Icon';
 
 export type CheckboxPropsType = {
   checked?: boolean,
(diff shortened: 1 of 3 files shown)


src/components/form-elements/_checkbox-radio.scss
```diff
@@ -52,7 +52,7 @@ $includeHtml: false !default;
 
     &__element:checked {
       & + .sg-checkbox__ghost {
-        border-color: $crInputCheckedColor;
+        border: 0;
```
matzimowski: to free some space for icon (this border was not visible anyway)

src/components/form-elements/Checkbox.jsx
```diff
@@ -42,7 +42,7 @@ class Checkbox extends PureComponent<CheckboxPropsType, CheckboxStateType> {
           {...additionalProps}
         />
         <label className="sg-checkbox__ghost" htmlFor={id}>
-          <Icon type={TYPE.CHECK} color={ICON_COLOR.ADAPTIVE} size={10} />
+          <Icon type="std-check" color={ICON_COLOR.ADAPTIVE} size={16} />
```
burakukula: maybe we could do just `color="adaptive"` here as well?

matzimowski: done

PR #1640, opened 2019-09-27
Proposal of the changes in terms of versioning and deployment

burakukula:
I would like to propose a change into our versioning process in style-guide:
Currently, we need to bump version every time we do any change, even a super small one. This is probably one of the reasons why we have currently `149.0.2` versions of the style-guide.

I would like to switch to versioning which is much more standardized and most of the open source projects that I know are following this approach - https://semver.org. We could do it as well! 😉

What will change:
 - we will bump version only once in a while (when there will be anything urgent we can obviously still bump the version and merge if that change is required for any Brainly projects.)
 - every version will contain more than one change, it could be a couple of improvements from a couple of PR or some bigger changes - it's up to maintainers to decide when we actually need another version of the library and what changes this version should include. 
 - deploy will be triggered only when the version will change, we could still merge into master, but only if the new version will be pushed we will actually deploy lib.

PROS:
 - standard versioning, easier to maintain and understand.
 - versioning will be a responsibility of maintainers, we could have
 
CONS:
 - bigger bumps in one version - it might be, that we would have more that one change in the version, so it would require from us, to actually bump more than one thing on the application side. The good news is that we don't develop style-guide so actively lately.
 - to make it work, we need to create more detailed documentation regarding versions. Currently, we don't take care of it very carefully, but for this to work we need to keep discipline regarding changelog. Which at the end should be in PROS not in CONS 🙂


CONTRIBUTING.md
```diff
@@ -110,21 +116,22 @@ Go here to find documentation about [FlEXBOX component](src/components/flex/READ
...
-
-1. Rebase your branch
-
-1. bump version in branch 
-
-1. Merge your PR to master
+1. Merge approved, pending PR's into master.
+2. Create a branch with a bumped version according to the rules above, create tag and merge it (get approvals of course).
+3. Create proper, specific and detailed release note, with all the changes included in the release
+4. Remember to let know contributors and Brainly engineers, that new version was released and it's ready to use
```
vergilius: How would you like to achieve that? Calling in PR? wdyt?

burakukula: I guess so, that this is the best way to let know external contributors - let them know in their PR with mentioning that we are merging they changes and newest version will be available soon with theirs changes. For internal we do have slack channel

vergilius: that's okay, but maybe add that to this point? otherwise someone might look for additional contact ✌️

burakukula: @vergilius sure, on my way. thx for feedback 👌

PinkaminaDianePie:
> bigger bumps in one version
I don't agree that we will have such cons. No one forces us to have only big releases. If we have enough changes to bump SG version, we can bump it, if not enough (like fixing typo in readme), we wont bump it. This change about giving us more flexibility only, not about switching to large updates only :) It's actually not really different from the current process, because we can have the same issue if we will update SG version multiple times and bump it in the main app only once at the end.

burakukula:
@PinkaminaDianePie sure, what I meant is that big bump now will be possible. but yeah, it's still on our side to decide. I hope no one will actually prepare huge updates, that we would need to then implement on the application side. This approach is definitely more flexible. I hope it will improve our workflow in SG.

aju:
We just need to be sure that all changes are included in changelog and someone that creates PR with release should take a look what was merged.

PR #1655, opened 2019-10-25
Upgrade dependencies due to vunlerabilities

SebastianCzoch:
I've updated a lot of dependencies (locked in yarn.lock) due to security reasons.

I'm strongly recommend to remove gulp because it is not maintained frequently

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,43 +2,42 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


aju:
Maybe we should bump dependencies in packages.json ? @tomasz-sodzawiczny @burakukula ?

SebastianCzoch:
@aju I think it can be done in separate PR

PR #1656, opened 2019-10-28
Counter component

burakukula:
Counter component, for now, it will create or basic version, which will be used in notifications and alert and points counter component. It might be that we would like to create later on more advanced component where it might be possible to use different icons in the counter, but for now we decided to force only points.
Additionally, `withAnimation` and `withPointer`, first needed for the basic counter in notifications, second one useful for points counter.

This is the first step on the way to merge current `<Badge />` into `<Labels />` and set "Badge" name free.

counter mov


src/components/counters/Counter.jsx
```diff
@@ -0,0 +1,144 @@
...
+export type CounterPropsType = {
+  /**
+   * Children to be rendered inside Counter
+   * @example <Counter type="basic">
+   *            text
+   *          </Counter>
+   */
+  children?: Node,
+  /**
+   * Specify type of the counter that you want to use, to types for now
```
ElaLukaszewicz: did you mean '...two types for now'? :)

src/components/counters/_counters.scss
```diff
@@ -0,0 +1,111 @@
...
+        }
+      }
+    }
+
+    &--with-animation {
+      transform: scale(0);
+      animation: sgBadgeAnimation 0.6s 2s forwards;
+    }
+
+    &--with-pointer {
```
matzimowski: not sure if it's should be here, if it's needed, it can be added on user's end with `className` or `style` prop.

matzimowski: wdyt?

burakukula: @matzimowski yeah, but I just figured that it might be useful, I guess there are these two cases when this component is a part of a text or can be clicked. But maybe it is overkill.

src/components/counters/Counter.jsx
burakukula: @ElaLukaszewicz yeah, sorry. on my way to fix it.

PR #1661, opened 2019-11-08
Adjust paddings of Button component

matzimowski:
Adjustments based on:

Screenshot 2019-11-08 16 10 05

after:
after1
after2

before:

before1
before2


src/components/buttons/_buttons.scss
```diff
@@ -41,13 +41,15 @@
     fill: $white;
     display: inline-flex;
     margin-right: spacing(xs);
+    margin-left: -4px;
```
matzimowski: i've decided to go with negative margins, no to introduce another modifier to container class like `--with icon` and override `--small` and `--large` modifiers. still looks clean imho

src/components/buttons/_buttons.scss
```diff
@@ -58,15 +60,15 @@
...
+    padding: 0 28px;
     font-size: 21px;
     line-height: 21px;
   }
 
   &--small {
     border-radius: 16px;
     height: 32px;
-    padding: 0 20px;
+    padding: 0 16px;
```
burakukula: @matzimowski you can use `spacing()` here I think.

PR #1669, opened 2019-11-13
Adjust counters

burakukula:
More generic solution for counters.

Screenshot 2019-11-13 at 15 54 57


src/components/counters/Counter.jsx
```diff
@@ -26,7 +21,7 @@ export type CounterPropsType = {
    *            text
    *          </Counter>
    */
-  children?: string,
+  children?: string | number,
```
aju: maybe just React.Node ?

PR #1671, opened 2019-11-17
Round buttons

burakukula:
This component was implemented in `style-guide-next` and it's used on new HP for not logged in user. We would like to introduce it into SG to maybe replace `style-guide-next` with the current one and in the future, for this component to replace `IconAsButton` component:

Screenshot 2019-11-17 at 22 44 19


src/components/round-buttons/RoundButton.jsx
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Icon types for round buttons
+   * @example <RoundButton iconType="answer" />
+   * @see type="points" https://styleguide.brainly.com/latest/docs/interactive.html?type="points"#round-buttons
+   * @see type="heart" https://styleguide.brainly.com/latest/docs/interactive.html?type="heart"#buttons
+   */
+  iconType?: $Values<typeof TYPE>,
+  /**
+   * Color variants for round buttons
+   * @example <RoundButton color="black-base-500" />
```
burakukula: I introduce the same naming as we do in `style-guide-next` for colors. The plan is to switch to this naming with all the colors in the future. WDYT?

matzimowski: but it means in app we will need to use different namings for colors? i am not sure if it's good idea and if it should not be introduced separately.

PR #1680, opened 2019-11-28
Fix Round Button PropsType and Docs

bartoszlorek:


src/components/round-buttons/RoundButton.jsx
```diff
@@ -52,6 +52,7 @@ export type RoundButtonPropsType = {
    * Additional class names
    */
   className?: string,
+  ...
```
bartoszlorek: because later `...props`

PR #1688, opened 2019-12-10
New inputs

burakukula:
Changes:

 - All form elements have 2 variants, default is always gray, additionally, we have white version.
 - Sizes are set in two variants - normal and large. 
 - updates in search - 2 versions with gray icon and with black round button
 - small README for form elements - it supposes to be extended in the future
 
form-elements mov


src/components/form-elements/_textarea.scss
```diff
@@ -77,23 +74,32 @@ $includeHtml: false !default;
...
+      height: 168px;
     }
 
     &--auto-height {
-      line-height: rhythm(1);
+      line-height: 24px;
       height: auto;
     }
+
+    &--nested {
```
matzimowski: remove me

src/components/form-elements/pages/input-interactive.jsx
```diff
@@ -48,17 +40,14 @@ const Inputs = () => {
...
     <div>
-      <DocsActiveBlock settings={settings}>
-        <Input placeholder="Simple input" />
-      </DocsActiveBlock>
-      <DocsActiveBlock settings={settings}>
+      <DocsActiveBlock backgroundColor="dark" settings={settings}>
         <Input
-          placeholder="Big wrong input"
+          backgroundColor="none"
+          placeholder="Big input"
```
matzimowski: maybe use a `large` word here too? wdyt? instead of `big`

burakukula: sure thing.

src/components/form-elements/_textarea.scss
burakukula: done

src/components/form-elements/pages/input-interactive.jsx
burakukula: solved

src/components/form-elements/Input.jsx
```diff
@@ -3,9 +3,9 @@
 import React from 'react';
 import classnames from 'classnames';
 
-type InputSizeType = 'small' | 'normal' | 'large';
+type InputSizeType = 'normal' | 'large';
 
-type InputColorType = 'normal' | 'light' | 'light-alt';
+type InputColorType = 'default' | 'white';
```
vergilius: so in one place we have `normal` and in other `default`, why is that?

burakukula: nope, colors are set to `default` and `white` and sizes to `normal` and `large`

src/components/form-elements/Input.jsx
```diff
@@ -53,28 +53,67 @@ export const TYPE = Object.freeze({
...
   size?: ?InputSizeType,
+  /**
+   * There are two color variants for form elements, default does not have to be specified
+   * @example <Input color="white" placeholder="placeholder" />
+   * @see color="default" https://styleguide.brainly.com/latest/docs/interactive.html?color="default"#input
+   * @see color="white" https://styleguide.brainly.com/latest/docs/interactive.html?color="white"#input
+   */
   color?: ?InputColorType,
+  /**
+   * Optional additional boolean to specified if it's valid
```
vergilius: do we need both `Optional additional` next to each other?

burakukula: nope, we don't I will fix it. thx

src/components/form-elements/Input.jsx
```diff
@@ -53,28 +53,67 @@ export const TYPE = Object.freeze({
...
+   * @example <Input valid placeholder="placeholder" />
+   */
   valid?: boolean,
+  /**
```
vergilius: would that be valid to have `valid` and `invalid` exclude each other? Right now it's not, and the CSS makes difference

src/components/form-elements/Input.jsx
burakukula: done.

src/components/form-elements/Textarea.jsx
```diff
@@ -5,25 +5,81 @@ import classnames from 'classnames';
...
+
 export const SIZE = {
   SHORT: 'short',
   NORMAL: 'normal',
   TALL: 'tall',
   XTALL: 'xtall',
 };
 
+export const TEXTAREA_COLOR = {
+  DEFAULT: 'default',
```
vergilius: isn't the no-option indicate the default behaviour (and look)?

src/components/form-elements/_select.scss
```diff
@@ -1,40 +1,43 @@
-$selectHeight: rhythm(3 / 2);
-$selectBackground: $white;
+$selectHeight: 40px;
```
vergilius: I'd like to see where these sizes come from

src/components/form-elements/Input.jsx
burakukula: It never was. We can refactor the logic of the components, but maybe let's do it separately. WDYT? Your point is valid for more than just inputs.

src/components/form-elements/Textarea.jsx
burakukula: We need it for export to sketch.

src/components/form-elements/_select.scss
```diff
@@ -1,40 +1,43 @@
...
     font-size: $selectFontSize;
     color: $selectTextColor;
-    box-shadow: $componentShadow;
     border-radius: $selectBorderRadius;
     height: $selectHeight;
+    border: 2px solid $graySecondaryLightest;
 
     &__element {
+      position: relative;
+      top: -2px;
```
vergilius: how about extract that to some `$borderRelatedVariable`? that would be clear what's happening here without need to check the border

src/components/form-elements/_select.scss
burakukula: this is the same size, as we have in Buttons, I can move it to config if you like.

src/components/form-elements/_select.scss
burakukula: 👍

src/components/form-elements/_select.scss
```diff
@@ -43,35 +46,136 @@ $includeHtml: false !default;
...
+        height: $selectLargeHeight;
+        padding: 0 50px 0 spacing(m);
+        
+        &::-ms-expand {
+          display: none;
+        }
+      }
+
+      &::placeholder {
+        font-size: 18px;
```
vergilius: why don't we reuse variables? like this `$inputLargeFontSize` - its 18px

src/components/form-elements/_text-input.scss
```diff
@@ -1,148 +1,137 @@
...
+    &--white {
+      background: $white;
+      border: 2px solid $white;
+      border-radius: $inputStandardBorderRadius;
 
-      &::placeholder {
-        color: $white;
-        font-weight: $fontWeightBold;
+      &:hover {
+        background-color: mix($white, $graySecondaryUltraLight, 12%);
```
vergilius: where does `12%` comes from? is the final color part of our palette?

src/components/form-elements/Input.jsx
vergilius: is that a updated standard for colors? there was `normal` previously in component

src/components/form-elements/Input.jsx
vergilius: gotit! 👍

src/components/form-elements/Textarea.jsx
vergilius: oh I see

src/components/form-elements/_select.scss
vergilius: I mean, even comment would be enough, that base size of our components is 40px and that comes from designs for example

👍

src/components/form-elements/_text-input.scss
```diff
@@ -1,148 +1,137 @@
...
+      &:active,
+      &:focus:hover,
+      &:active:hover {
+        background-color: $white;
+        border: 2px solid $inputFocusInvalidBorderColor;
+        box-shadow: none;
+      }
+
+      &:hover {
+        background-color: mix($inputBackground, $graySecondaryLight, 80%);
```
vergilius: mix again, do we reuse these colors in other components?

src/components/form-elements/_text-input.scss
burakukula: it's not, this is the same pattern as we use for `Button`. For filled element, we use 80% mix and for light version 12%. I will try to describe it in some part of the README.

src/components/form-elements/_select.scss
burakukula: I am moving it into configuration. I will let you know as soon as it's refactored

src/components/form-elements/Input.jsx
burakukula: yeah, this would be the new approach for form elements.

PR #1689, opened 2019-12-16
Fixme for form elements

burakukula:
Small adjustments for inputs and other for elements (moved from fixme)
The main visual change is icon size for select:

For normal size icon x16 for large we keep x24:

Screenshot 2019-12-16 at 10 45 26


src/components/form-elements/Select.jsx
```diff
@@ -10,7 +10,7 @@ type OptionsPropsType = {
   ...
 };
 
-type SelectSizeType = 'small' | 'normal';
+type SelectSizeType = 'normal' | 'large';
```
burakukula: ups, I forgot about it in prev PR

PR #1690, opened 2019-12-18
Use new mobile logotypes in HomeButton and get ride of old ones

matzimowski:
- brainly plus stays untouched

before:
Screenshot 2019-12-18 at 16 36 33

after:
Screenshot 2019-12-18 at 16 40 40


src/components/home-button/_home-button.scss
```diff
@@ -1,5 +1,5 @@
-$homeButtonMobileHeight: 36px;
-$homeButtonDefaultHeight: 36px;
+$homeButtonMobileHeight: 32px;
```
matzimowski: change of size requested by design team

PR #1693, opened 2019-12-23
New labels

burakukula:
They will replace Badge (currently we use only mustard badge on prod, so this will be quick migration)

Screenshot 2019-12-23 at 15 03 20
Screenshot 2019-12-23 at 15 03 28

labels mov

### DIFF
--- a/src/components/labels/Label.jsx
+++ b/src/components/labels/Label.jsx
@@ -54,7 +54,7 @@ export const LABEL_COLORS_SET = {
   MONO: 'mono',
 };
 
-type PropsType = {
+type PropsType = $ReadOnly<{
   /**
    * Specify type of label
    * @example <Label
@@ -123,7 +123,7 @@ type PropsType = {
    */
   className?: string,
   ...
-};
+}>;
 
 const Label = ({
   children,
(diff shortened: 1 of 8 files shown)


src/components/labels/Label.jsx
```diff
@@ -0,0 +1,183 @@
+// @flow strict
+
+import React from 'react';
+import type {Node} from 'react';
+import classNames from 'classnames';
+import Text from '../text/Text';
+import * as IconModule from '../icons/Icon';
```
matzimowski: import Icon, {TYPE: ICON_TYPE} from '../icons/Icon';`?

src/components/labels/Label.jsx
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * Children to be rendered inside Label
+   * @example <Label
+   *           type="default"
+   *           color="blue"
+   *          >
+   *            example label
+   *          </Label>
+   */
+  children: Node,
```
matzimowski: should we allow to define children? i know we did in the past, but i know not sure if it's needed here now. wdyt?

src/components/labels/Label.jsx
burakukula: 🧐

burakukula: @matzimowski This import would not work I think, there is no export of `ICON_TYPE`. I need to also use `IconTypeType` which I get from `IconModule `. I saw we do it this way in `IconAsButton`. But maybe there is a better way...

matzimowski: ok, let's stick to what we have. 👌

Issue #1412, opened 2018-03-07
RWDhelper should work with multiple elements


matikrk:
Add to RWDhelper possibility to add multiple children.
Use React.Children.map for this, if problems with returning, return fragment with all childrens

useful links:
 https://reactjs.org/docs/react-api.html#reactchildren

PR #1700, opened 2020-01-03
Add handling of array and React.Fragment as children in RWDHelper

matzimowski:
Fixes https://github.com/brainly/style-guide/issues/1412

### DIFF
--- a/src/components/helpers/RwdHelper.jsx
+++ b/src/components/helpers/RwdHelper.jsx
@@ -32,13 +32,17 @@ const RwdHelper = ({hide, children}: PropsType) => {
 
   const hideClass = `sg-hide-for-${hide}`;
 
-  if (typeof children === 'string') {
+  if (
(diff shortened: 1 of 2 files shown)


matzimowski:
good point @matikrk. actually it's your issue i was trying to solve. i agree we should not use this component if not needed, maybe let's close this issue then?

src/components/helpers/RwdHelper.jsx
```diff
@@ -32,13 +32,17 @@ const RwdHelper = ({hide, children}: PropsType) => {
 
   const hideClass = `sg-hide-for-${hide}`;
 
-  if (typeof children === 'string') {
+  if (
+    typeof children === 'string' ||
```
maciekbb: @matzimowski I think you could type it with flow https://flow.org/en/docs/react/children/

PR #1706, opened 2020-01-14
Prepare style-guide for publishing to npm

maciekbb:
This will allow to use style-guide without compiling it with babel first.

To generate a `build/` folder I added a `package` task.


.babelrc.js
```diff
@@ -8,9 +8,9 @@ const babelEnv = params => [
 module.exports = api => {
   api.cache(true);
   return {
-    presets: [babelEnv({modules: false}), '@babel/preset-react'],
+    presets: [babelEnv({modules: false}), '@babel/preset-react', '@babel/preset-flow'],
     plugins: [
-      '@babel/plugin-transform-flow-strip-types',
+      '@babel/plugin-transform-runtime',
```
maciekbb: This is needed so babel helpers are imported instead of being inlined.

This probably means `@babel/runtime` should be added as peep dependency, am I right?

.babelrc.js
```diff
@@ -8,9 +8,9 @@ const babelEnv = params => [
 module.exports = api => {
   api.cache(true);
   return {
-    presets: [babelEnv({modules: false}), '@babel/preset-react'],
+    presets: [babelEnv({modules: false}), '@babel/preset-react', '@babel/preset-flow'],
     plugins: [
-      '@babel/plugin-transform-flow-strip-types',
```
vergilius: can this affect our app current build process?

build/components/layout/iframe-pages/layout_large.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
vergilius: @maciekbb can we exclude this path: `iframe-pages`? it looks like this is used only for building demo page

package.json
```diff
@@ -84,7 +86,8 @@
     "watch": "gulp build && gulp watch & yarn start",
     "test": "yarn run scss-lint && yarn run scss-unused-variables && yarn flow && yarn run js-lint && jest",
     "eslint": "./node_modules/.bin/eslint",
-    "beta:watch": "webpack-dev-server --hot --inline --open --watch --mode=development"
+    "beta:watch": "webpack-dev-server --hot --inline --open --watch --mode=development",
+    "package": "yarn babel src --out-dir build --only 'src/components' --only 'src/index.js' --ignore '**/*.spec.jsx' --ignore '**/pages/**' --ignore '**/__mocks__/**'"
```
vergilius: lets ignore `iframe-pages`

.babelrc.js
maciekbb: I put flow preset there instead, I think it does the same thing

vergilius: aw right, didn't connect the dots immediately, sorry 👍

aju:
@maciekbb did you consider building it on install instead on committing to repo ?

maciekbb:
@aju This would mean one more step in build pipeline.

Looking at some resources online (like https://betterstack.dev/blog/npm-package-best-practices/#heading-directory-structure-conventions) I think this is a good practise to have dist/ folder published.

aju:
@maciekbb yeah but style guide is not published :) - I was thinking about using npm post install script option not a build step :)

maciekbb:
@aju I think we could publish it to npm actually.

`yarn package` can be run in `preversion` hook to guarantee we have build folder up to date.

aju:
@maciekbb yeah its an option as well :)
I was thinking still for now about postinstall  https://docs.npmjs.com/misc/scripts - it would have to be build on developer machine on install time

PR #1713, opened 2020-01-29
Remove outline none property and fix hover in IconAsButton on touch devices

ElaLukaszewicz:
- remove ```outline: none``` property
- fix hover in IconAsButton on touch devices: change css on `active` instead of `hover` -> hover state stays after mouse click on mobile and it should be visible only during click

Before:
old

After:
new

### DIFF
--- a/src/components/icon-as-button/_icon-as-button.scss
+++ b/src/components/icon-as-button/_icon-as-button.scss
@@ -61,7 +61,6 @@ $includeHtml: false !default;
     display: inline-flex;
     background: none;
     border: none;
-    outline: none;
     padding: 0;
     cursor: pointer;
     overflow: visible;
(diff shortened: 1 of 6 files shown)


src/components/icon-as-button/_icon-as-button.scss
```diff
@@ -141,20 +159,22 @@ $includeHtml: false !default;
...
 
     &--action-active {
       background: $iconAsButtonColor;
       fill: $iconAsButtonActionColor;
       border-radius: 50%;
 
-      &:hover {
-        fill: $iconAsButtonActionColor;
-      }
-
```
ElaLukaszewicz: it's unnecessary,  this fill is in all states

PR #1714, opened 2020-01-30
add onKeyDown prop, role and tabIndex to close Icon in TopLayer

ElaLukaszewicz:
add onKeyDown prop, role and tabIndex to close Icon in TopLayer for accessibility reasons

### DIFF
--- a/src/components/toplayer/TopLayer.jsx
+++ b/src/components/toplayer/TopLayer.jsx
@@ -16,7 +16,9 @@ export const SIZE = {
 export type PropsType = {
   children: Node,
   onClose?: (SyntheticMouseEvent<HTMLDivElement>) => mixed,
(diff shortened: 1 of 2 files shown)


src/components/toplayer/TopLayer.jsx
```diff
@@ -74,7 +76,13 @@ const TopLayer = (props: PropsType) => {
   return (
     <div {...additionalProps} className={topLayerClassName}>
       {onClose ? (
-        <div className="sg-toplayer__close" onClick={onClose}>
+        <div
+          className="sg-toplayer__close"
+          onClick={onClose}
+          onKeyDown={onKeyDown}
```
matzimowski: `onCloseButtonKeyDown` or similar, wdyt?

PR #1717, opened 2020-02-08
Fix licence ID

seahindeniz:
GPLv3 is not a valid SPDX license expression


matzimowski:
@seahindeniz from what i can see we cannot merge this PR, i guess it's because it's some 'uknown repository'.

seahindeniz:
Ah! I'll make another PR @matzimowski

PR #1722, opened 2020-02-10
align text to center and fix bugs

ElaLukaszewicz:
Align text to center in ```<FlashMessage />``` when text breaks into more than one line - requested by designers
Before:
image

After:
image

### DIFF
--- a/src/components/header/iframe-pages/small-device.jsx
+++ b/src/components/header/iframe-pages/small-device.jsx
@@ -6,7 +6,7 @@ import HeaderLeft from '../HeaderLeft';
 import HeaderMiddle from '../HeaderMiddle';
 import HeaderRight from '../HeaderRight';
 import HomeButton from '../../home-button/HomeButton';
-import Search, {COLOR as SEARCH_COLOR} from 'search/Search';
+import Search from 'search/Search';
 import RWDHelper, {TYPE as RWD_TYPE} from 'helpers/RwdHelper';
 import IconAsButton, {
   TYPE as ICON_TYPE,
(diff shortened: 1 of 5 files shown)


src/components/header/iframe-pages/small-device.jsx
```diff
@@ -35,7 +35,6 @@ const SmallDeviceExample = () => (
                 placeholder="Find all the answers..."
                 fullWidth
                 color={SEARCH_COLOR.LIGHT}
-                adaptiveIco
```
ElaLukaszewicz: old non-existing prop

src/components/header/iframe-pages/small-device.jsx
```diff
@@ -35,7 +35,6 @@ const SmallDeviceExample = () => (
                 placeholder="Find all the answers..."
                 fullWidth
                 color={SEARCH_COLOR.LIGHT}
```
seahindeniz: `SEARCH_COLOR.LIGHT` is also not exist should be `SEARCH_COLOR.WHITE` according to this
https://github.com/brainly/style-guide/blob/499e637d5ccc49deab32e75d87293e185ba591fa/src/components/form-elements/Input.jsx#L60-L63

ElaLukaszewicz: yes, thank you very much, I removed this prop because otherwise the search icon is not visible (if white)

seahindeniz: Oh, I think it's visible pretty visible :smiley: 
image

seahindeniz: What do you think @ElaLukaszewicz?

ElaLukaszewicz: yes, sorry for that, it rebuild that way for me 
image so i thought this prop was to change color of the icon, i didn't notice the icons didn't load yet: fix here https://github.com/brainly/style-guide/pull/1725 🤦‍♀

seahindeniz: Sure, thanks for handling it @ElaLukaszewicz :v:

PR #1733, opened 2020-03-04
Replace old dropdown

Malinicz:
Replace old dropdown from style guide with one from new home page footer (image below). 
This dropdown has been added to 'components' section with a 'deprecated' label since it should be removed in the future.

image

image


src/components/dropdowns/Dropdown.jsx
```diff
@@ -1,67 +1,59 @@
...
-    {
-      'sg-dropdown--full-width': fullWidth,
-      'sg-dropdown--opened': opened,
-    },
-    className
-  );
-  const itemsClass = classNames('sg-dropdown__items', {
-    'sg-dropdown__items--fixed': fixed,
-  });
+type PropsType = {
```
matzimowski:
```type PropsType = $ReadOnly<{
  name: string,
  links: <LinksType>,
}>```

all the props seem to be ReadOnly. wdyt?

File: src/components/dropdowns-deprecated/DropdownDeprecated.jsx [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+import classNames from 'classnames';
+
+import Icon from '../icons/Icon';
+
+type LinksType = {
+  label: string,
+  url: string,
+};
+
+type PropsType = {
```
matzimowski:
```
type PropsType = $ReadOnly<{
  name: string,
  links: <LinksType>,
}>
```

all the props seem to be ReadOnly. wdyt?

Malinicz: fixed

PR #1736, opened 2020-03-06
Dropdown component update

Malinicz:
- update Dropdown with 2 new props (fullWidth, colorMode)
- fix issue with run-sequence in watch mode

image

### DIFF
--- a/src/components/dropdowns/Dropdown.jsx
+++ b/src/components/dropdowns/Dropdown.jsx
@@ -62,11 +62,7 @@ const Dropdown = ({
       <Icon type={open ? 'arrow_up' : 'arrow_down'} size={24} color="gray" />
 
       {open && (
-        <div
-          className={classNames('sg-dropdown__items', {
-            'sg-dropdown__items--light': lightMode,
-          })}
-        >
+        <div className="sg-dropdown__items">
           {links.map((link, index) => (
             <a key={index} href={link.url} className="sg-dropdown__item">
               {link.label}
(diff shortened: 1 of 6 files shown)


src/components/dropdowns/Dropdown.jsx
```diff
@@ -14,13 +14,23 @@ type PropsType = $ReadOnly<{
   name: string,
   links: Array<LinksType>,
   initiallyOpened?: boolean,
+  colorMode?: 'dark' | 'light',
```
bartoszlorek: I'd keep what we have for selects `color?: 'default' | 'white'`

src/components/dropdowns/Dropdown.jsx
```diff
@@ -41,7 +51,11 @@ const Dropdown = ({name, links, initiallyOpened}: PropsType) => {
 
   return (
     <div
-      className={classNames('sg-dropdown', {'sg-dropdown--opened': open})}
+      className={classNames('sg-dropdown', {
+        'sg-dropdown--opened': open,
+        'sg-dropdown--light': lightMode,
```
bartoszlorek: and then `sg-dropdown--white`

src/components/dropdowns/Dropdown.jsx
Malinicz: sure, fixed - I guess in this case it makes sense to keep it consistent with Select component, however I'm not a huge fan of naming things with color names as if in the future we change the color to e.g. light gray, giving it prop color="white' could be slightly confusing :) @bartoszlorek

src/components/dropdowns/Dropdown.jsx
Malinicz: fixed

PR #1739, opened 2020-03-12
Adjust sizes of logotypes in HomeButton

matzimowski:
## after
in style-guide:
Screenshot 2020-03-12 at 09 17 52
on website:
Screenshot 2020-03-12 at 08 11 00
Screenshot 2020-03-12 at 08 11 21
Screenshot 2020-03-12 at 08 16 59

## before
in style-guide:
Screenshot 2020-03-12 at 09 18 09

on website:
Screenshot 2020-03-12 at 08 07 39
Screenshot 2020-03-12 at 08 08 06
Screenshot 2020-03-12 at 08 08 29

### DIFF
--- a/src/components/home-button/_home-button.scss
+++ b/src/components/home-button/_home-button.scss
@@ -36,18 +36,19 @@ $includeHtml: false !default;
 
     &--brainly-plus &__logo-big {
       height: 41px;
+      width: 62px;
     }
 
     &--eodev &__logo-big {
-      width: 114px;
+      width: 68px;
     }
 
     &--nosdevoirs &__logo-big {
-      width: 161px;
+      width: 97px;
     }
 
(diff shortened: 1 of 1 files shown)


src/components/home-button/_home-button.scss
```diff
@@ -36,18 +36,19 @@ $includeHtml: false !default;
 
     &--brainly-plus &__logo-big {
       height: 41px;
+      width: 62px;
     }
 
     &--eodev &__logo-big {
-      width: 114px;
```
matzimowski: i could not define width at all and base only on height (`width: auto`) but I noticed some issues on IE11 so I think it's more safe to do it this way.

bartoszlorek: agreed 👍

PR #1743, opened 2020-03-13
Header update

Malinicz:
Update header component:

1) set white mode as default 
2) remove "light" prop (according to design team we should leave only default white version of the header)
3) add "withDivider" prop to show/hide bottom border for header

image
image


src/components/header/HeaderContainer.jsx
```diff
@@ -6,16 +6,16 @@ import classnames from 'classnames';
 
 type PropsType = {
   children: Node,
-  light?: boolean,
+  dark?: boolean,
```
matzimowski: hm, i see you point, but `dark` sounds a bit strange when we know it's just a `blue` color. maybe it could be a `color` prop where we accept `blue` as a string only?
wdyt? @Malinicz @bartoszlorek

Malinicz: I kinda don't like passing color props, but you're right that it doesn't make much sense to name it "dark" unless we agree we have light and dark mode only 🤔 I think main Brainly colors are mint and blue, maybe we could pass color prop as "primary" and "secondary", so when it changes in the future we won't have to update this prop (e.g. material ui is doing similar thing in their lib). What do you think @matzimowski ?

matzimowski: actually, do we even need a blue topbar? any reason that we want to keep it?

Malinicz: I spoke with Karol and he said he would leave blue one in style guide, so I did. But yeah, for now we don't need it :)

matzimowski: and, so let's keep it for now, and yea, maybe we could try this approach with primary/secondary. @bartoszlorek wdyt?

Malinicz: actually, looking at our core colors https://styleguide.brainly.com/158.1.1/docs/basics.html#colors we really should set 4 options: `white/dark/primary/secondary` (or `light/dark/primary/secondary`), so to get white one it would be `<Header color="light" />`

seahindeniz: There are too many darks and lights in other color variations on the core colors section. White and black is more obvious than dark and light. Why don't go with standard color names?

bartoszlorek: I'd go with `color?: 'white' | 'blue'` and the white as a default

Malinicz: Sure, if you think it's better to follow already established patterns that's fine with me :) I proposed this solution after looking at our legacy code (user profile) where class '.mint' became totally different color after rebranding. So in case of rebranding, if whole codebase uses something like `<Component color="blue" />`, we would have to change that in all the places in the codebase, whereas when using `<Component color="primary" />` the only thing we would have to do is to switch `primary` to different color and update style guide package. But I understand that this is much bigger change and probably should be discussed with larger audience :)

matzimowski: agree, let's go with simple solution for now and consider introducing a new pattern later on after some discussion.
i am  even not sure if we gonna support blue one and keep it in style-guide if designers not gonna use it.

Malinicz: yup just got decision from design team that we should remove blue completely - I'll update code shortly

src/components/header/_header.scss
```diff
@@ -17,11 +17,15 @@ $includeHtml: false !default;
...
+      border-bottom: 2px solid $graySecondaryLightest;
+    }
+
     &__container {
-      background-color: $bluePrimary;
+      background-color: $white;
 
-      &--light {
-        background-color: $white;
+      &--dark {
```
matzimowski: out

src/components/header/_header.scss
```diff
@@ -7,7 +7,7 @@ $includeHtml: false !default;
   .sg-header {
     width: 100%;
     margin-bottom: rhythm(3) - $headerHeight;
-    box-shadow: $defaultBottomShadow;
+    box-shadow: none;
```
matzimowski: maybe we can just remove this line?

Malinicz: true, removed 👍

PR #1746, opened 2020-03-13
Add Mobile Icons Set

bartoszlorek:
- icon set for our Mobile Team
- React implementation for Design Team (to Sketch)

Screen Shot 2020-03-13 at 11 47 46


src/components/mobile-icons/MobileIcon.jsx
```diff
@@ -0,0 +1,77 @@
...
+  | 'heart-outline'
+  | 'keyboard'
+  | 'options'
+  | 'sent'
+  | 'share-on-ios'
+  | 'star-outline';
+
+export type SizeType = 'small' | 'medium' | 'normal';
+
+export const TYPE: {[name: string]: IconTypeType, ...} = {
```
Malinicz: Not sure if that's possible in Flow, but in typescript you could create union type from either object key or value. So in case you add new value, type is updating automatically

src/components/mobile-icons/MobileIcon.jsx
```diff
@@ -0,0 +1,77 @@
...
+  size?: SizeType,
+  color?: IconColorType,
+  className?: string,
+  ...
+};
+
+const MobileIcon = ({
+  type,
+  size = 'normal',
+  color = 'light',
```
Malinicz: It's not a big deal but I'm slightly confused why sometimes color is "white" and sometimes color is "light", light/dark seems to be a kind of "mode", not a color 🤔

src/components/mobile-icons/MobileIcon.jsx
bartoszlorek: It's also possible but in Flow union type from object values like `{a: 'small', b: 'big'}` will be `string` not a union `'small' | 'big'`

src/components/mobile-icons/MobileIcon.jsx
bartoszlorek: yeah, it's confusing! but here I'm using the same color values as other icon sets so it's a global issue that we can discuss with design team.

src/components/mobile-icons/MobileIcon.jsx
Malinicz: ok I see, that's a bummer :(

src/components/mobile-icons/MobileIcon.jsx
Malinicz: ok!

src/components/mobile-icons/MobileIcon.jsx
```diff
@@ -0,0 +1,77 @@
...
+  FLASHLIGHT_ON: 'flashlight-on',
+  HEART_OUTLINE: 'heart-outline',
+  KEYBOARD: 'keyboard',
+  OPTIONS: 'options',
+  SENT: 'sent',
+  SHARE_ON_IOS: 'share-on-ios',
+  STAR_OUTLINE: 'star-outline',
+};
+
+type PropsType = {
```
ElaLukaszewicz: we don't want ReadOnly here?

src/components/mobile-icons/_mobile-icon.scss
```diff
@@ -0,0 +1,19 @@
+@import '../icons/icons-mixin';
+
+.sg-mobile-icon {
+  height: 64px;
```
ElaLukaszewicz: maybe put those values into variables?

src/components/mobile-icons/MobileIcon.jsx
ElaLukaszewicz: and maybe some documentation/description like in Box or Label?

PR #1756, opened 2020-03-26
Add error messages for inputs

anna-wro:
https://jira.z-dn.net/browse/ANSPRIO-2854

Screenshot 2020-03-26 at 15 22 26
Screenshot 2020-03-26 at 15 32 18

### DIFF
--- a/src/components/form-elements/Input.jsx
+++ b/src/components/form-elements/Input.jsx
@@ -2,6 +2,7 @@
 
 import React from 'react';
 import classnames from 'classnames';
+import Text from '../text/Text';
 
 type InputSizeType = 'normal' | 'large';
 
@@ -108,6 +109,11 @@ export type InputPropsType = {
    */
   withIcon?: boolean,
   /**
+   * Optional string if there should be an error message displayed
+   * @example <Input errorMessage="This is an error" />
+   */
+  errorMessage?: string,
+  /**
    * Additional class names
    */
   className?: string,
@@ -130,6 +136,7 @@ const Input = (props: InputPropsType) => {
     invalid,
     className,
     setInputRef,
+    errorMessage,
     ...additionalProps
   } = props;
 
@@ -153,14 +160,35 @@ const Input = (props: InputPropsType) => {
     className
   );
 
+  const errorMessageClass = classnames(
+    'sg-input__error-message',
+    {
+      [`sg-input__error-message--${String(size)}`]: size !== SIZE.NORMAL,
+    },
+    className
+  );
+
   return (
-    <input
-      {...additionalProps}
-      type={type}
-      ref={setInputRef}
-      className={inputClass}
-      value={value}
-    />
+    <>
+      <input
+        {...additionalProps}
+        type={type}
+        ref={setInputRef}
+        className={inputClass}
+        value={value}
+      />
+      {errorMessage !== undefined &&
+        errorMessage !== null &&
+        errorMessage !== '' && (
+          <Text
+            size={size === SIZE.NORMAL ? 'xsmall' : 'small'}
+            color="peach-dark"
+            className={errorMessageClass}
+          >
+            {errorMessage}
+          </Text>
+        )}
+    </>
   );
 };
 
--- a/src/components/form-elements/Input.jsx
+++ b/src/components/form-elements/Input.jsx
@@ -3,6 +3,7 @@
 import React from 'react';
 import classnames from 'classnames';
 import Text from '../text/Text';
+import Flex from 'flex/Flex';
 
 type InputSizeType = 'normal' | 'large';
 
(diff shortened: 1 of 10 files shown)


src/components/form-elements/Textarea.jsx
```diff
@@ -123,13 +130,34 @@ const Textarea = (props: TextareaPropsType) => {
...
-      value={value}
-    />
+    <>
+      <Type
+        {...additionalProps}
+        className={textareaClass}
+        ref={textareaRef}
+        value={value}
+      />
+      {errorMessage !== undefined &&
```
matzimowski: checking just `errorMessage && ()` should be enough, or `flow` has some issues here?

anna-wro: yup :(

anna-wro: `Sketchy null check on string [1] which is potentially an empty string. Perhaps you meant to check for null or undefined [1]? (`sketchy-null-string`) (Input.jsx:180:8)`

src/components/form-elements/Textarea.jsx
```diff
@@ -123,13 +130,34 @@ const Textarea = (props: TextareaPropsType) => {
...
+        errorMessage !== null &&
+        errorMessage !== '' && (
+          <Text
+            size={size === SIZE.NORMAL ? 'xsmall' : 'small'}
+            color="peach-dark"
+            className={errorMessageClass}
```
matzimowski: maybe we could achieve all of this using `<Flex />` here?

anna-wro: Yes! I'll change it ASAP

anna-wro: done

matzimowski:
@anna-wro could u please add it to interactive pages too, please?

anna-wro:
@matzimowski I added them

src/components/form-elements/Input.jsx
```diff
@@ -154,13 +162,27 @@ const Input = (props: InputPropsType) => {
...
 
   return (
-    <input
-      {...additionalProps}
-      type={type}
-      ref={setInputRef}
-      className={inputClass}
-      value={value}
-    />
+    <div>
```
matzimowski: @bartoszlorek wdyt about wrapping it with `<div />` here?
maybe we should wrap it only when it has an errror message?
or it's better to always wrap it.

src/components/form-elements/Textarea.jsx
```diff
@@ -124,12 +132,26 @@ const Textarea = (props: TextareaPropsType) => {
...
+        className={textareaClass}
+        ref={textareaRef}
+        value={value}
+      />
+      {errorMessage !== undefined &&
+        errorMessage !== null &&
+        errorMessage !== '' && (
+          <Flex marginTop="xxs" marginLeft={size === SIZE.NORMAL ? 's' : 'm'}>
+            <Text
+              size={size === SIZE.NORMAL ? 'xsmall' : 'small'}
```
ElaLukaszewicz: maybe be consistent and just 'normal' instead of SIZE.NORMAL, wdyt?

src/components/form-elements/Input.jsx
```diff
@@ -154,13 +162,27 @@ const Input = (props: InputPropsType) => {
...
+      <input
+        {...additionalProps}
+        type={type}
+        ref={setInputRef}
+        className={inputClass}
+        value={value}
+      />
+      {errorMessage !== undefined &&
+        errorMessage !== null &&
```
bartoszlorek: it cannot be `null` so you can remove it.

src/components/form-elements/Input.jsx
bartoszlorek: `input` gets `component()` mixins and it's a `display: inline-block` so we want to keep it, not rewrite to just a `block`. I think it can be just a react `fragment` and let others decide about structure in implementation.

matzimowski: agree, can we stick to `</>` @anna-wro ?

bartoszlorek: hmmm, on the other hand `message` will be always displayed as a `block`. My only concern is when div will be full width and input inside not.

bartoszlorek: we don't want such case 😢 
Screen Shot 2020-03-27 at 09 31 50
Screen Shot 2020-03-27 at 09 30 17
Screen Shot 2020-03-27 at 09 30 24

matzimowski: @anna-wro 
after discussion with @bartoszlorek we think we could stick to `<div />`as a wrapper.
but not the break anything from layout perspective we need to:
1. add there a className, i.e. `sg-input__wrapper` and styles:
```
.sg-input__wrapper {
  display: inline-block;
  $--full-width {
    width: 100%;
  }
}
```
so fullWith prop would be used on both wrapper and input
2. let's wrap it only when there is an errorMessage, when there is no, let's return just an `<input />`.

@anna-wro wdyt?

anna-wro: ok, I'll change it

anna-wro: do we want the same with textarea?

anna-wro: @bartoszlorek @matzimowski @ElaKrk Please check one more time :)

PR #1765, opened 2020-04-01
Prioritize margin props and small

### LINKED ISSUE #1764: Prioritization of margin on Flex component
Margin property of Flex has miss-prioritized. Let's say, if Flex has used with `margin="m" marginLeft="s"` properties. Since the size "m" is bigger than "s", the "marginLeft" prop can't override "margin" prop.

This is the output
image

https://github.com/brainly/style-guide/blob/443228ba367c69f8365cd4c2c5f3a3bdd25a00f6/src/components/flex/_flex.scss#L129-L149

seahindeniz:
Fixes #1764 

Before: 
image

After:
image


matzimowski:
@seahindeniz thanks a lot for you contribution!

Issue #1584, opened 2019-06-04
Removed button colors on v144.0.0


seahindeniz:
Hello guys and @burakukula !

_**TL;DR:**_
Is it fully necessary to remove the peach color from the style guide and could you make it available again?

---

>I understand that the colors that you've removed in the #1581 may/are not used in the company based(internal) applications.
And I also know that the style-guide is not ment to be only for the company based projects, otherwise it wouldn't be available on the Github as a public repository; am I right?
Other people can able to use the style-guide in everywhere.
Since I'm using those colors- especially the Peach colored button in my project(like a lot) and your changes also affecting it. 

I have a project and it's for a small group of users of Brainly and it's called Brainly Tools. Espacially for the moderators, admins and the content managers. It's a browser extension and it gives us a bit more better user experience and it's allow us to make proper actions faster than usual. This makes users happy enough :).

My project uses the style-guide.css file that loads under the head tag in new brand pages. Like, feed/home page or the question view page.
But my project is also uses the style-guide.css file in the legacy pages like the user profile page or the user contents page by injecting the css file in the head tag.

Whenever you change something from the style guide, it makes me worried as you can imagine :sweat_smile:
The buttons that I've used in the new brand pages is based on the _untouched_ style-guide.css file that comes from the head tag. The legacy pages uses the injected css file.

So I would like to ask; is it fully necessary to remove the peach color from the style guide and could you make it available again?

Please see the example usage in the screenshots below

**Before**:
![image](https://user-images.githubusercontent.com/5789670/58881788-88a93700-86e3-11e9-81ee-41c7461c04ee.png)

**After**:
![image](https://user-images.githubusercontent.com/5789670/58881859-bbebc600-86e3-11e9-83cb-2ef467202b16.png)

burakukula:
hey @seahindeniz! 🙂Nice to hear that you use our style-guide for your widget! Awesome! 🚀
I guess, in general, if you don't want to be affected by our changes right away, you could also use versioning in your project as we do in the main application or encapsulate your widget somehow.
Actually, the changes that you mention (on the second screenshot) are not on production yet, we are still testing them. The current version of style-guide on production is still on `https://styleguide.brainly.com/143.9.1/style-guide.css`. Do you have versioning on your side? 
[...]

PR #1766, opened 2020-04-03
remove destructive and warning button type

ElaLukaszewicz:
remove destructive and warning button type (https://jira.z-dn.net/browse/ANSPRIO-2965)

### DIFF
--- a/src/components/buttons/README.md
+++ b/src/components/buttons/README.md
@@ -87,7 +87,7 @@ import {Button} from 'style-guide';
 ```
 
 #### Additional buttons
-There are also 1 additional button created for special case: facebook button, which has custom color from outside the palette.
+There is also 1 additional button created for special case: facebook button, which has custom color from outside the palette.
 
 ```jsx
 import {Button} from 'style-guide';
--- a/src/components/buttons/README.md
+++ b/src/components/buttons/README.md
@@ -87,18 +87,10 @@ import {Button} from 'style-guide';
 ```
 
 #### Additional buttons
-There are also 3 additional buttons created for special cases - destructive and warning, and facebook button, what have custom set of colors from outside the palette.
+There are also 1 additional button created for special case: facebook button, which has custom color from outside the palette.
 
 ```jsx
 import {Button} from 'style-guide';
-
-<Button type="destructive">
-    Mustard destructive button
-</Button>
-
-<Button type="warning">
-    Peach warning
-</Button>
 
 <Button type="facebook">
     I am FB button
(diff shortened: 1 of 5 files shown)


src/components/buttons/README.md
```diff
@@ -87,19 +87,11 @@ import {Button} from 'style-guide';
 ```
 
 #### Additional buttons
-There are also 3 additional buttons created for special cases - destructive and warning, and facebook button, what have custom set of colors from outside the palette.
+There are also 1 additional button created for special case: facebook button, which has custom color from outside the palette.
```
matzimowski, line 90: "There is"

seahindeniz:
@ElaKrk please see https://github.com/brainly/style-guide/issues/1584#issuecomment-509237104

matzimowski:
Hi @seahindeniz,
thanks for being alert 👌 
Could you please use `link-button-peach` and `link-button-mustard` instead the ones we are removing?
It's a changed requested by our design team, they are not planning to use them anymore in our projects, so there's no reason for us to maintain them and we have to stay in sync with our Design System. too
We are planning to introduce further changes to buttons soon, maybe some new types will appear.

seahindeniz:
Hi @matzimowski 

Thanks for letting me know and I always in wait for new plans, changes.
But I just want you to remind that, this project is not going to be used only on the Brainly web app or Brainly's sub-projects. So please make changes that would not affect others.

Also I can't open link in description but I would like to ask; is it absolutely necessary to remove colors from buttons?
I'm already using the `link-button-peach, along with `sg-button--destructive` and both are represents different things. I would like to ask changing names to `primary-peach` and `primary-mustard` to comply the naming standards in Colors- instead of `destructive` or `warning` before, but I didn't bother because either way buttons would be exists.

seahindeniz:
Anyways :) 
I don't want to get in your way if it's one of the job requirements. So I hope design team would consider extending additional colors on the primary buttons.
Thanks!

matzimowski:
@seahindeniz 
thanks for your understanding 👌 

I see your point, but unfortunately, in case someone still wants to use those types of buttons, they need to stick to a particular version of style-guide.

changes that we gonna do are not public yet from design perspective and i am not sure when exactly we introduce them, but yea, that's our plan.

seahindeniz:
@matzimowski the version topic from my part has already been discussed in here https://github.com/brainly/style-guide/issues/1584#issuecomment-498733034

> For the newer UI, my extension uses the components of the style-guide.css that loaded in DOM. That means, extension doesn't have a control of <style /> tag of the style-guide.css file in the <head /> tag added by Brainly.

If you make changes, that means I get affected as well and I have to comply accordingly so that is why I'm tried to explain. Please understand my concerns too :slightly_smiling_face:

ElaLukaszewicz:
@seahindeniz I will let you know when we will release new version of SG. I believe it will happen on Monday or Tuesday next week.

matzimowski:
@seahindeniz yea, I see your point.

Actually, regarding these buttons we removed, it's just a removal of modifiers, so they will both fallback to `sg-button--primary` and become "dark" ones. 🕶 

I guess it breaks your UI, but yea, for now what we can suggest is just to switch to `sg-button--link-button-peach` and `sg-button--link-button-mustard`.

thanks again and sorry for inconvenience.

seahindeniz:
@ElaKrk got it, thanks :wink: 

@matzimowski I'm trying not to create customized styles from my side but that's OK. I can manage somehow for having the `primary-peach` and `primary-mustard` modifiers exists :rocket:

PR #1782, opened 2020-04-17
Remove suspicious Button

matzimowski:


src/components/buttons/Button.jsx
```diff
@@ -191,8 +191,4 @@ const Button = ({
   );
 };
 
-export const Lol = () => {
```
matzimowski, line 194: let's remove it, it looks dangerous 🕵️

PR #1785, opened 2020-04-24
Exclude "default" type from className

seahindeniz:
I think there is no point having the `default` as a class name.

I would like to suggest having a boolean based `type` checking since there is no additional `type` than `strong` but that might cause backward compatibility issues and may require additional changes

Before:
image

After:
image


seahindeniz:
Hi
Should I close this PR if it breaks something?

matzimowski (approved):

bartoszlorek (approved):
we have one occurrence of this modifier in our main repo so we will remove it after 🤙

matzimowski:
@bartoszlorek i've seen it, seem to be not needed

matzimowski:
@seahindeniz thanks for your contribution 🙏

PR #1790, opened 2020-05-14
Add iconOnly modifier to Button

matzimowski:
## changes:
- `iconOnly` modifier introduced
- `hover` and `toggle hover` state of `Button` removed from docs
- `RoundedButton` component became deprecated

we want to make it work in both ways: in React and in plain html

Screenshot 2020-05-14 at 13 18 31
Screenshot 2020-05-14 at 13 18 39
Screenshot 2020-05-14 at 13 18 43


src/components/buttons/Button.spec.jsx
```diff
@@ -54,6 +54,25 @@ test('icon', () => {
...
+  const icon = <span>:P</span>;
+  const button = shallow(
+    <Button icon={icon} iconOnly>
+      Some text
+    </Button>
+  );
+
+  expect(button.contains(icon)).toEqual(true);
+  expect(button.find('.sg-button--icon-only')).toHaveLength(1);
+  expect(button.find('.sg-button__icon')).toHaveLength(1);
```
Malinicz, line 67: maybe add additional expect to check that the text is not rendered?

src/components/buttons/Button.spec.jsx
```diff
@@ -54,6 +54,25 @@ test('icon', () => {
...
+});
+
+test('icon only when no icon', () => {
+  const button = shallow(<Button iconOnly>Some text</Button>);
```
Malinicz, line 71: is it possible to type it in a way that when you pass `iconOnly`, you must pass also `icon`?

bartoszlorek: it's already typed but tests are flow-free

Malinicz: yepp, but I think we don't need such test then? idk

matzimowski: good point, removed

src/components/buttons/Button.spec.jsx
matzimowski: but children are optional already, if you pass some text i will be rendered but not visible

PR #1804, opened 2020-05-28
Overlay add dark color

matikrk:
Screenshot 2020-05-28 at 12 42 39

needed in tutoring
as far I know used in payments in brainly

### DIFF
--- a/src/components/overlay/Overlay.jsx
+++ b/src/components/overlay/Overlay.jsx
@@ -1,20 +1,37 @@
 // @flow strict
 
-import React from 'react';
 import type {Node} from 'react';
+import React from 'react';
 import classnames from 'classnames';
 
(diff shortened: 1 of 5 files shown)


src/components/overlay/Overlay.jsx
```diff
@@ -1,20 +1,37 @@
...
+  DARK: 'dark',
+};
+
+type ColorType = $Values<typeof COLOR>;
+
+const Overlay = ({
+  partial,
+  children,
+  className,
+  color,
```
matzimowski, line 26: `color = "blue"`, but i see that the fallback is anyway to blue, so maybe not needed there

matikrk: ye I left fallback like it was

PR #1824, opened 2020-06-10
Rename Select sizes

anna-wro:
No visual changes
Text input will be done separately

https://brainly.atlassian.net/browse/ANSPRIO-3367
https://brainly.atlassian.net/wiki/spaces/DesignSystem/pages/33523275/FNC#SELECT

Screenshot 2020-06-10 at 11 20 58

### DIFF
--- a/src/sass/_config.scss
+++ b/src/sass/_config.scss
@@ -136,8 +136,8 @@ $breakpointsMap: (
 );
 
 // Form elements configuration
-$formElementNormalHeight: 40px; // Standard size for elements like inputs, selects or buttons
-$formElementLargeHeight: 56px;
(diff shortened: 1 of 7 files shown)


src/sass/_config.scss
```diff
@@ -136,8 +136,8 @@ $breakpointsMap: (
 );
 
 // Form elements configuration
-$formElementNormalHeight: 40px; // Standard size for elements like inputs, selects or buttons
-$formElementLargeHeight: 56px;
+$formElementSizeM: 40px; // Standard size for elements like inputs, selects or buttons
```
matzimowski, line 139: lets use elementSpacing() mixin @anna-wro

PR #1831, opened 2020-06-15
Remove RoundButtons

matzimowski:
## It's a breaking change

RoundButtons have been deprecated for a while already - finally we can get rid of them.

please use `<Button ... iconOnly />` instead.

Screenshot 2020-06-15 at 14 30 52


matzimowski:
fyi @seahindeniz

PR #1840, opened 2020-06-23
Add new comment icons

Malinicz:
Add new comment icons


src/images/icons/comment.svg
```diff
@@ -0,0 +1,4 @@
...
+</svg>
```
bartoszlorek, line 4: newline at the end

PR #1843, opened 2020-06-24
Create File Handler

ElaLukaszewicz:
Create File Handler
image
image


src/components/form-elements/pages/input-interactive.jsx
```diff
@@ -46,7 +46,6 @@ const Inputs = () => {
     <div>
       <DocsActiveBlock backgroundColor="dark" settings={settings}>
         <Input
-          backgroundColor="none"
```
ElaLukaszewicz, line 49: it was causing some errors in console

src/components/file-handler/pages/file-handlers.jsx
```diff
@@ -0,0 +1,63 @@
...
+        <FileHandler iconType="answer">text</FileHandler>
+      </DocsBlock>
+    </Flex>
+    <Flex marginRight="l">
+      <DocsBlock info="with on close">
+        <FileHandler onClose={closeCallback}>text</FileHandler>
+      </DocsBlock>
+    </Flex>
+    <Flex marginRight="l">
+      <DocsBlock info="laoding">
```
bartoszlorek, line 35: loading

src/components/file-handler/FileHandler.jsx
```diff
@@ -0,0 +1,167 @@
...
+   * @default false
+   */
+  loading?: boolean,
+  /**
+   * Callback, called by clicking on **close** button. If specified, button will be added automatically
+   * @example <FileHandler onClose={() => doSomething()}>
+   *            text
+   *          </FileHandler>
+   */
+  onClose?: (SyntheticMouseEvent<HTMLDivElement>) => mixed,
```
bartoszlorek, line 71: why is `onClose` there? is it open something that triggers this fn on close?

ElaLukaszewicz: it is onClick in button with 'close' Icon - we did it the same in Label or TopLayer and i think it makes sense, but maybe i'm wrong?

src/components/file-handler/FileHandler.jsx
```diff
@@ -0,0 +1,167 @@
...
+  setTextRef,
+  className,
+  ...props
+}: PropsType) => {
+  const fileHandlerClass = classNames('sg-file-handler', {
+    'sg-file-handler--closable': onClose,
+    [`sg-file-handler--${COLORS_MAP[color]}`]: color,
+  });
+
+  const showLink = src !== null && src !== undefined;
```
Malinicz, line 118: Not sure if null check if necessary: `Optional parameters will accept missing, undefined, or matching types. But they will not accept null.` from https://flow.org/en/docs/types/functions/

ElaLukaszewicz: you are right, thank you! ❤️

src/components/file-handler/FileHandler.jsx
bartoszlorek: I mean to re-think possibilities because for sure you can close toplayer, but closing label can be discussing 😄 for sure you can remove label.

ElaLukaszewicz: so you suggest 'onRemove' prop? or 'onClickCloseButton'? or sth else?

bartoszlorek: `onClose` is fine if we are thinking about a button/icon, not something to close. I wanted to say something

PR #1847, opened 2020-06-29
Add possibility to display button icon on the right

Malinicz:
image
image

### DIFF
--- a/src/components/buttons/Button.jsx
+++ b/src/components/buttons/Button.jsx
@@ -198,7 +198,6 @@ const Button = ({
 
   const iconClass = cx('sg-button__icon', {
     [`sg-button__icon--${size || ''}`]: size,
-    'sg-button__icon--right-aligned': reversedOrder,
   });
 
   let ico;
--- a/src/components/buttons/Button.jsx
+++ b/src/components/buttons/Button.jsx
@@ -66,10 +66,12 @@ type ButtonIconType =
   | {
       icon?: Node,
       iconOnly?: null,
+      reversedOrder?: boolean,
     }
   | {
       icon: Node,
       iconOnly?: boolean,
+      reversedOrder?: null,
     };
 
 export type ButtonPropsType = {
@@ -171,6 +173,7 @@ const Button = ({
   type,
   icon,
   iconOnly,
+  reversedOrder,
   href,
   fullWidth,
   disabled,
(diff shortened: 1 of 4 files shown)


src/components/buttons/_buttons.scss
```diff
@@ -272,6 +292,21 @@ $largeButtonSize: componentSize(l);
...
         margin-left: -3px;
       }
+
+      &--right-aligned {
+        margin-right: -2px;
+        margin-left: spacing(xs);
+      }
+    }
+
+    .sg-button__icon--right-aligned.sg-button__icon--s {
```
Malinicz, line 302: not sure why, but `&--modifier#{&}--modifier-2` doesn't work here - I had to do it this way :(

src/components/buttons/Button.jsx
```diff
@@ -66,10 +66,12 @@ type ButtonIconType =
   | {
       icon?: Node,
       iconOnly?: null,
+      reversedOrder?: boolean,
     }
   | {
       icon: Node,
       iconOnly?: boolean,
+      reversedOrder?: null,
```
Malinicz, line 74: If there's no icon, flow will complain when we pass reversedOrder. But technically, we can add `icon`, `iconOnly` and `reversedOrder` (which doesn't make sense), and flow will not complain. I would leave this as is, because fixing that would introduce more unnecessary complexity imho (we would have to add more interfaces here). Wdyt?

matzimowski: i think it's acceptable for now till we find a solution how to make it simpler

src/components/buttons/Button.jsx
```diff
@@ -188,12 +191,14 @@ const Button = ({
...
       'sg-button--icon-only': Boolean(icon) && iconOnly,
       [`sg-button--${String(type)}-toggle-${String(toggle)}`]: toggle,
+      'sg-button--reversed-order': reversedOrder,
     },
     className
   );
 
   const iconClass = cx('sg-button__icon', {
     [`sg-button__icon--${size || ''}`]: size,
+    'sg-button__icon--right-aligned': reversedOrder,
```
matzimowski, line 201: i am wondering if we could get rid of this modifier and stick to `sg-button--reversed-order` only

src/components/buttons/_buttons.scss
```diff
@@ -47,14 +47,30 @@ $largeButtonSize: componentSize(l);
     margin-right: spacing(xs);
     margin-left: -4px;
 
+    &--right-aligned {
```
matzimowski, line 50:
```
.sg-button--reversed-order & {
      margin-right: -4px;
      margin-left: spacing(xs);
}
```
here and accordingly in other places. wdyt?

PR #1850, opened 2020-06-30
Icons needed for tutoring

wasilakkf:
Confirmed with @matikrk that these icons should be included in styleguide.

Screenshot 2020-06-30 at 14 32 21


matzimowski:
let's discuss it first, there were some changes from design team

PR #1854, opened 2020-07-06
Rename text sizes from normal to medium

matzimowski:
no visual changes

this change applies to `Text`, `Link`, `TextBit` and `Headline` components

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "172.2.2",
+  "version": "173.0.0",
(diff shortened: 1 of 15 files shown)


seahindeniz:
Will be there any changes in shortening the names of the size @matzimowski? 
Like `medium` to `m`

matzimowski:
@seahindeniz we conduct all this kind of changes in sync with our design system team (we export our style-guide to their design tools so we use common naming). in case of texts, i asked and they prefer to stick to those longer names.

PR #1860, opened 2020-07-09
Label / Change color of close icon

anna-wro:
Requested by design

Before:
Screenshot 2020-07-09 at 10 30 13

After:
Screenshot 2020-07-09 at 10 29 56


src/components/labels/Label.spec.jsx
```diff
@@ -76,54 +76,102 @@ describe('Label', () => {
...
+
     expect(label.hasClass('sg-label--mint-primary')).toEqual(true);
-    expect(label.find(Icon).props().color).toEqual('light');
+    expect(closeIcon.props().color).toEqual('light');
+    expect(heartIcon.props().color).toEqual('light');
     expect(label.find(Text).props().color).toEqual('white');
   });
 
-  test('has proper styles if transparent', () => {
+  xtest('has proper styles if transparent', () => {
```
ElaLukaszewicz, line 123: remove x please :)

anna-wro: Nice catch!

PR #1884, opened 2020-07-30
Add typescript definitions

webDevAdvocate:
**Description**: As tutoring project was migrated to TypeScript we lost type checking provided by flow in style-guide.
**This PR contains starting point to create type definitions file for typescript, and should be developed incrementally when some new components will be used in some TypeScript project**

- As only tutoring use TS currently, I was able to check only parts of types used by tutoring-frontend project
- Rest part of types can contain some errors as it was hard to check it.
- For now I have exported all components from index.js and few type definitions. 
- The rest of types and some const's should be added to declaration file when there will be a need for that

First question is, if it is a good place to add this file?
- Maybe we should have it internally in tutoring project?
- Or maybe we should already publish them as part of @types namespace

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,8 +104,8 @@
     "eslint": "./node_modules/.bin/eslint",
     "beta:watch": "webpack-dev-server --hot --inline --open --watch --mode=development",
     "package-clean": "yarn rimraf esm/ commonjs/",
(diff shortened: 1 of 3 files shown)


webDevAdvocate:
As agreed with @maciekbb better approach at this stage will be adding this file internally to tutoring-project so I am closing PR

PR #1914, opened 2020-09-09
Add useRef type to input's setInputRef prop type.

### LINKED ISSUE #1913: Input component doesn't allow to pass useRef hook
There is a `setInputRef` prop to pass ref to input element, but it is typed to only allow the previous/"class" way of adding references.

kubajasiolek:
closes #1913

### DIFF
--- a/src/components/form-elements/Input.jsx
+++ b/src/components/form-elements/Input.jsx
@@ -122,7 +122,9 @@ export type InputPropsType = {
   /**
    * Additional function to set ref for input
    */
-  setInputRef?: { current: ?HTMLInputElement } | (ref: ?HTMLElement | ?HTMLInputElement) => mixed,
+  setInputRef?:
+    | {current: ?HTMLInputElement}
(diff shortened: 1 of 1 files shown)


src/components/form-elements/Input.jsx
```diff
@@ -122,7 +122,7 @@ export type InputPropsType = {
   /**
    * Additional function to set ref for input
    */
-  setInputRef?: (ref: ?HTMLElement) => mixed,
+  setInputRef?: { current: ?HTMLInputElement } | (ref: ?HTMLElement | ?HTMLInputElement) => mixed,
```
ElaLukaszewicz, line 125: did you thought about just passing ref like in FIleHandler? not a function?

kubajasiolek: That's how I'm passing it. But I don't want to break the API, so I left the previous function-way as well

ElaLukaszewicz (approved):

mcekiera (approved):

ziejka (approved):

PR #1915, opened 2020-09-09
Switch to forked library @brainly/s3

Malinicz:
Package "s3" has been forked to @brainly organization. One of dependencies of this package (graceful-fs) was causing our builds to fail. Since this package is no longer maintained, we decided to fork it and make needed updates. As a long-term solution, we need to switch "s3" package to something more up-to-date.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -353,7 +353,7 @@
     "@babel/traverse" "^7.6.2"
     "@babel/types" "^7.6.0"
 
-"@babel/highlight@^7.10.4":
+"@babel/highlight@^7.10.4", "@babel/highlight@^7.8.3":
   version "7.10.4"
   resolved "https://registry.yarnpkg.com/@babel/highlight/-/highlight-7.10.4.tgz#7d1bdfd65753538fabe6c38596cdb76d9ac60143"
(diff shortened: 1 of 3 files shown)


anna-wro (approved):

coderitual (approved):

ElaLukaszewicz (approved):

yarn.lock
```diff
@@ -353,7 +353,7 @@
     "@babel/traverse" "^7.6.2"
     "@babel/types" "^7.6.0"
 
-"@babel/highlight@^7.10.4":
+"@babel/highlight@^7.10.4", "@babel/highlight@^7.8.3":
```
Malinicz, line 356: not sure why yarn changed that 🤔  anyways, it should not cause any issues

coderitual: @Malinicz 

it means that one of dependency was found to be using: "@babel/highlight@^7.8.3" but is satisfied by the the same: version "7.10.4". This is how yarn is trying to merge different lock version in a defensive way.

This is correct since eg. @babel/code-frame@7.8.3 and @babel/code-frame@^7.0.0 are present and are using different version. yarn lock will grow over time and will have such a duplicates unless is cleaned.

Malinicz: cool, good to know 👍

PR #1919, opened 2020-09-10
Turn off side effects flag for the package

coderitual:
Side effect flag may have impact on consumers package size after build and we need to have it set to `false`. Style guide itself doesn't have side effect other than css file which is being loaded separately from the package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,5 +130,5 @@
       ".*": "babel-jest"
     }
   },
-  "sideEffects": true
+  "sideEffects": false
 }


Malinicz (approved):

package.json
```diff
@@ -130,5 +130,5 @@
       ".*": "babel-jest"
     }
   },
-  "sideEffects": true
+  "sideEffects": false
```
anna-wro, line 133: do you need to include it if false? What's the default value?

anna-wro (approved):

package.json
Malinicz: it was set to false previously, I set it to true when I tested storybook build, and forgot to change that 🙈

Malinicz: because of that we had errors in bundle size checks when bumping style guide version in brainly-frontend (it was importing more things than needed)

coderitual: @anna-wro default value is `true`. It can have false but also an array which files have side effects and which don't.

PR #1927, opened 2020-09-24
Add space between util

coderitual:
## What?

Add **space between** util for controlling the space between child elements (Lobotomized Owls).

### Basic
Creates space between children. Ignoring children with `.sg-space-ignore` class
```html
<ul class="sg-space-y-xxl">
  <li>1</li>
  <li>2</li>
  <li class="sg-space-ignore">3</li> <!--  Ignored child because of class -->
  <li>4</li>
</ul>
```

### Auto emtpy node remove

```html
<ul class="emtpy:sg-space-y-xxl">
  <li>1</li>
  <li>2</li>
  <li class="sg-space-ignore">3</li> <!--  Ignored child because of class -->
  <li>4</li>
  <li></li> <!--  Ignored child because empty node -->
</ul>
```

### Responsive variants

`md:sg-space-y-xs`, `lg:sg-space-y-xs`

```html
<!--  XS margin by default, xl margin on medium screen, xxl margin on large -->
<ul class="sg-space-y-xs md:sg-space-y-xl lg:sg-space-y-xxl"> 
  // ...
</ul>
```

It works in similar way for `empty` variant

```html
<!--  XS margin by default, xl margin on medium screen, xxl margin on large -->
<ul class="empty:sg-space-y-xs md:empty:sg-space-y-xl lg:empty:sg-space-y-xxl"> 
  // ...
</ul>
```

### Easily applicable and testable from dev tools (leverages autocomplete)

image

## Bundle size

This approach generates quite a lot of classes. Normally it is easily mitigated by using purge css againt the website. Since we don't have it yet we need to be careful here.

Let's discuss if result below is acceptable. If not we can easily get rid of some variants (eg. large screens, or empty variant becomes default)

**before:**
image

**after**
image

## Context reference

https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,1 @@
+module.exports = require('eslint-config-brainly/prettier.config');
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,1 +1,0 @@
-module.exports = require('eslint-config-brainly/prettier.config');
--- a/src/sass/_utils.scss
+++ b/src/sass/_utils.scss
@@ -3,7 +3,7 @@
 
 // Space between (Lobotomized owl) util for controlling the space between child elements.
 // You can take the child out of the layout area by giving the child a .sg-space-ignore class.
-// To automaticaly ignore empty children use "empty" class variant e.g. empty:sg-spacey-m
+// To automaticaly ignore empty children use "empty" class variant e.g. empty:sg-space-y-m
 @each $breakpoint, $variant in $responsiveVariants {
   @include sgResponsive($breakpoint) {
     @each $sizeName, $size in $sizesSetup {
--- a/src/sass/_utils.scss
+++ b/src/sass/_utils.scss
@@ -6,7 +6,7 @@
 // To automaticaly ignore empty children use "empty" class variant e.g. empty:sg-space-y-m
 @each $breakpoint, $variant in $responsiveVariants {
   @include sgResponsive($breakpoint) {
-    @each $pseudo, $name in ('': '', ':not(empty)': 'empty\\:')
+    @each $pseudo, $name in ('': '', ':not(:empty)': 'empty\\:')
     {
       .#{$variant}#{$name}sg-space-y-0 > * + *:not(.sg-space-ignore)#{$pseudo} {
         margin-top: 0 !important;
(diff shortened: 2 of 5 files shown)


.prettierrc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('eslint-config-brainly/prettier.config');
```
coderitual, line 1: Please ignore that, it satisfies vscode when prettier formatting is disabled for projects not having config file.

matzimowski (approved):
👏

Fasosnql (approved):
I like this approach. I'm curious, do you have some plan on wider implementation of that? I mean if that a start point to push everyone to use more classes than components and overrides? Or it is just util for internal usage?

Malinicz (approved):
This looks awesome :)

src/sass/_utils.scss
```diff
@@ -0,0 +1,33 @@
+// Utils are meant to override other styles (See 7-1 or ITCSS for reference)
+// sass-lint:disable no-important
+
+// Space between (Lobotomized owl) util for controlling the space between child elements.
+// You can take the child out of the layout area by giving the child a .sg-space-ignore class.
+// To automaticaly ignore empty children use "empty" class variant e.g. empty:sg-space-y-m
+@each $breakpoint, $variant in $responsiveVariants {
```
Malinicz, line 7: @coderitual I'm just wondering if we should add some docs on styleguide.brainly.com about how to use those classes? WDYT?

coderitual: @Malinicz I totally agree. For sure there will be section in our new storybook docs, but maybe it is also worth putting docs into our basic section of current style guide. Let me do this in separate PR.

coderitual:
> I like this approach. I'm curious, do you have some plan on wider implementation of that? I mean if that a start point to push everyone to use more classes than components and overrides? Or it is just util for internal usage?

@Fasosnql Preferably I would like to expand on this idea to have other util classes. At this point lack of purging mechanism in our frontend app is a big limitation. In other words we need to stop linking directly to style guide and have a build step which will handle preparing css for every entry point of our app. I think we need to explore that topic in the first place before we go any further. but this use case was great opportunity to start.

bartoszlorek (approved):

PR #1939, opened 2020-10-06
Remove image offset in avatar

coderitual:
Removing offset from avatar image. Not sure why it was introduced but we cannot move element by border size here. Normally nested element should be moved by `border` to the right if they were absolutely positioned with `border box` sizing on wrapper.

Please let me know if you are aware what use case was meant to be fixed by that :)

image

### DIFF
--- a/src/components/avatar/_avatar.scss
+++ b/src/components/avatar/_avatar.scss
@@ -61,11 +61,6 @@ $includeHtml: false !default;
 
     &--with-border {
       border: solid $avatarBorderColor $avatarBorderWidth;
-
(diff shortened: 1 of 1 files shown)


Malinicz (approved):

bartoszlorek (approved):
I have introduced it during my avatar refactoring to keep the same size of avatar with or without border when there is an icon (fixed size) instead of image. In that case `border-box` does not work because it anyways push fixed icon down.

Possible solution can be replacing border with inset box-shadow with hard edge.

matzimowski (approved):

PR #1952, opened 2020-10-21
Storybook setup - HTML markup code generation for mdx and CSF

coderitual:
📗  Storybook adjustments to work perfectly for Brainly Style Guide:

This will enable us to use storybook as a main dev tool in brainly style guide. Once we move all components docs to the new format I prepare ci/cd to publish storybook as our main style guide documentation. 

- **JSX and HTML source code** is available out of the box (with **prettier** formatting for html)
- **Custom Canvas and Preview** extended to work within storybook (no changes for user).
- **Transparent code block supported in MDX and CSF** - also with multiple stories inside one canvas. All changes are applied behind the scene so creating stories is exactly the same as in official documentation. We're not limited to mdx. Args stories and and all benefits they provide( automatically generated controls, actions etc.) are in place.
- **Custom background in stories** - we use storybook@next, once 6.1 lands we can remove some custom configuration for that.
- **Extended code blocks** - based on `@storybook/components` and `@storybook/addon-docs/blocks`  we keep the same look and feel while have our customisation applied.
- **a11y plugin** - opened by default in dev mode
- **Some visual changes to have Brainly vibe** - it will be further discussed with frontend devs and designers

Screenshot 2020-10-23 at 13 44 08
Screenshot 2020-10-23 at 13 44 21
Screenshot 2020-10-23 at 13 44 42
Screenshot 2020-10-23 at 13 44 46
Screenshot 2020-10-23 at 13 44 59
Screenshot 2020-10-23 at 13 45 11
Screenshot 2020-10-23 at 13 45 57
Screenshot 2020-10-23 at 13 46 06
Screenshot 2020-10-23 at 13 57 07
image


.storybook/main.js
```diff
@@ -1,17 +1,123 @@
 const path = require('path');
+const argv = require('yargs').argv;
+
+const IS_PRODUCTION = Boolean(argv.production);
+const VERSION = argv.production ? pkg.version : 'dev';
```
Fasosnql, line 5:
```suggestion
const VERSION = IS_PRODUCTION ? pkg.version : 'dev';
```

.storybook/main.js
```diff
@@ -1,17 +1,123 @@
...
-    });
 
-    return config;
-  }
+    const _config = {
+      ...config,
+      module: {
+        ...config.module,
+        rules: [
+          ...config.module.rules.slice(1),
```
Fasosnql, line 25: could you add comment, what you're removing?

.storybook/main.js
```diff
@@ -1,17 +1,123 @@
...
+            query: {
+              name: 'static/media/[name].[hash:8].[ext]',
+            },
+          },
+        ],
+      },
+    };
+
+    _config.resolve.modules.push(
+      ...[
```
Fasosnql, line 113: Likely you don't need this spread here and you can just do:
```
_config.resolve.modules.push(
        SOURCE_COMPONENTS_DIR,
        SOURCE_DOCS_DIR,
        path.join(SOURCE_DIR, 'images'),
        'node_modules',
    );
```

.storybook/main.js
```diff
@@ -1,17 +1,123 @@
...
-    config.module.rules.push({
-      test: /\.scss$/,
-      use: ['style-loader', 'css-loader', 'sass-loader'],
-      include: path.resolve(__dirname, '../src/main.scss'),
-    });
 
-    return config;
-  }
+    const _config = {
+      ...config,
```
Fasosnql, line 21: I'm not sure if those all spread operation are needed, what I'm seeing is only removing one element from `config.module.rules` and pushing new properties, so maybe it is better to do:
```
config.module.rules = [...config.module.rules.slice(1), ...[custom rules]]
```

Malinicz (approved):
awesome stuff! 🎖️

coderitual:
thanks for taking the first step in this direction @Malinicz 💪

src/_docs/blocks/Stories.jsx
```diff
@@ -0,0 +1,39 @@
...
+  includePrimary?: boolean;
+}
+
+export const Stories = ({title, includePrimary = false}: StoriesProps) => {
+  const context = useContext(DocsContext);
+  const componentStories = getDocsStories(context);
+
+  let stories: Array<DocsStoryProps> = componentStories;
+
+  if (!includePrimary) stories = stories.slice(1);
```
Fasosnql, line 22: instead of `let` maybe is better to have:
```
const stories = [!includePrimary ? ...componentStories.slice(1) : ...componentStories]
```

matzimowski (approved):
great work 👏  it looks so impressive!

bartoszlorek (approved):

.storybook/main.js
coderitual: yea, good point

.storybook/main.js
coderitual: Yep I didn't need that. I got this code from storybook author to make alpha version working, hopefully I will remove big part of it once 6.1 version is released.

.storybook/main.js
coderitual: I think I cannot, push takes comma separatet args, if I push array it will be array in array
image

coderitual: right

.storybook/main.js
coderitual: done

ziejka (approved):

PR #1953, opened 2020-10-29
Fix static build and font loading in manager ui and preview ui

coderitual:
**Fix css and icons loading** - Set `sideEffects` flag to point to files which contains side effects. This enable webpack to properly inlcude those assets into the bundle.

**Fix font loading in stories and preview ui** - by removing conflicting file loader (there is default file loader provided in storybook)

**Use brainly font in manager ui** - by using the same font served from public dir by storybook static server.

image


bartoszlorek (approved):

Fasosnql (approved):

package.json
```diff
@@ -136,5 +136,8 @@
       ".*": "babel-jest"
     }
   },
-  "sideEffects": false
+  "sideEffects": [
```
Malinicz, line 139: does it have any impact on npm package size?

coderitual: @Malinicz it does affect webpack while using our style guide package. This array means that everything beside those 2 matches are side effects free and can be removed automatically by webpack when specific module is not imported.

All css and scss containing side effects by default (they are added to the DOM). The same is for icons (but we're not importing them from sources so this shouldn't have any impact.

For scss, we still want to have tree shaking working and this should be handled by `usedExports` (know as tree shaking in webpack). This technique is not the same as side effect flag and relies on terser to guess whether particular statements are side effects free and can be removed if not used (it works worse than sideeffects flag but still optimisation is present)

Expected result for our frontend app should be the same as now (I didn't touched js files, these are still marked as). but **we gonna be 100% sure once I will merge new version of style guide to the app**. If now I will take a closer look at  files responsible for importing selected scss file from sg

Thanks for pointing this out though :)

Malinicz (approved):

PR #1955, opened 2020-11-02
Add Accordion Component

coderitual:
Accordion based on accordion compoent from `components/src/ui`:
- changed interface (`spacing` prop uses `sg-spacing-[...]` utility class internally)
- added storybook stories (next docs)
- added current docs. Note: Interactive part has problem related to `useLayoutEffect` for jsx rendering. Current docs don't handle that. I won't fix it as we move to storybook.
- edit: fixed responsiveness of component after expanding
- edit: changed transition function to ease out (which fits better for expanding, for collapsing it's better to have ease in but it's not a big issue)

responsive bug will be fixed in separate pr

Screenshot 2020-11-02 at 17 11 09
Screenshot 2020-11-02 at 17 11 58
Screenshot 2020-11-02 at 17 12 37

### DIFF
--- a/src/components/accordion/Accordion.jsx
+++ b/src/components/accordion/Accordion.jsx
@@ -25,7 +25,7 @@ export const AccordionContext = createContext<AccordionContextType>({
   onChange: () => undefined,
 });
 
-const spaceClasses = {
+export const spaceClasses = {
   xxs: 'sg-space-y-xxs',
   xs: 'sg-space-y-xs',
   s: 'sg-space-y-s',
@@ -63,7 +63,7 @@ const Accordion = ({
 
   const classes = cx(
     {
-      [`${spaceClasses[spacing]}`]: !!spacing,
+      [`${spaceClasses[spacing]}`]: spaceClasses[spacing],
     },
     className
   );
--- a/src/components/accordion/Accordion.jsx
+++ b/src/components/accordion/Accordion.jsx
@@ -1,0 +1,77 @@
+//@flow strict
+
+import React, {createContext, useCallback, useState} from 'react';
+import cx from 'classnames';
+
+type PropType = $ReadOnly<{
+  allowMultiple?: boolean,
+  children: React$Node,
+  className?: string,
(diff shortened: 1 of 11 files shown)


src/components/accordion/Accordion.jsx
```diff
@@ -0,0 +1,78 @@
+//@flow strict
+
+import React, {createContext, useCallback, useState} from 'react';
+import cx from 'classnames';
+
+type PropType = $ReadOnly<{
+  allowMultiple?: boolean,
+  children: React$Node,
+  className?: string,
+  spacing?: 'xxs' | 'xs' | 's' | 'm' | 'l' | 'xl' | 'xxl' | 'xxxl' | 'xxxxl',
```
seahindeniz, line 10: Hi @coderitual 

I haven't fully adapted to the `sg-space` utility but do you have any plans for creating a new file/module that contains some sort of definition/types related to the utility class?

In this line, it's kind of defined statically, and this makes me wonder; is this going to statically defined on future components as well?
IMO, having these space modifiers to be imported from a module/module-like file, may simplify the process by referencing them on future components 🤔

coderitual: Hey @seahindeniz , That kind od utility classes started to live recently as a complementary adjustment to components as immediate response to the needs which arose on frontend app. We don't have yet full interface prepared. **It's early stage of this idea and the usage is somehow decoupled and similar to the one from `tailwind`**. Use them as normal css classes so any other react component or static html can be tweaked by them. This is not ideal by any means but because we build our css from scss separately from react components I don't wan't to depend on any of scss values yet. Another reason is there are no benefits on bundle size of imported modules. Implementation you commented on is just small improvement compared to how we used this component before (`<Accordion className="sg-space-y-m">`).

The long term plan is to move slowly towards better modularisation both on components and scss level (css on cdn will still live but as style guide is constantly growing it becomes suboptimal). This would also allow us to have better tooling for such extensions. There is a clear need to be able to extend layout components using design tokens (like spacing, maybe typography etc). Please treat those utility more like WIP towards final solution. In ideal world we would achieve something similar to what `styled system` is in styled components world but without styled components. Before that happens we need explore some ideas.

matzimowski (approved):
lgtm

PR #1958, opened 2020-11-03
Add modal component to the Style Guide

coderitual:
Add modal component to the style guide

Screenshot 2020-11-04 at 14 55 01
Screenshot 2020-11-04 at 14 51 26
Screenshot 2020-11-04 at 14 53 53

### DIFF
--- a/src/components/modal/Modal.jsx
+++ b/src/components/modal/Modal.jsx
@@ -2,6 +2,7 @@
 
 import * as React from 'react';
 import TopLayer from '../toplayer/TopLayer';
+import type {PropsType as TopLayerPropsType} from '../toplayer/TopLayer';
 import Overlay from '../overlay/Overlay';
 
 const OVERLAY_CLASS_NAME = 'sg-overlay';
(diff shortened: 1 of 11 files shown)


src/components/modal/Modal.jsx
```diff
@@ -0,0 +1,71 @@
+// @flow strict
+
+import * as React from 'react';
+import TopLayer from '../toplayer/TopLayer';
+import Overlay from '../overlay/Overlay';
+
+const OVERLAY_CLASS_NAME = 'sg-overlay';
+
+type PropsType = {
```
bartoszlorek, line 9: we should think about these props, I believe most of them aren't needed

maciekbb: I guess we should think if our goal here is to be backward-compatible or to provide the best API

bartoszlorek: exactly, I think modal is not widely used because of this complex API

coderitual: I agree with you @bartoszlorek  @maciekbb . Moving this component to the SG made me think about how many props are just passed down to the `TopLayer` and many of them are not working in modal in expected way (speaking gently). My initial idea was to keep the same interface for compatibility but if you're not happy from current implementation of Modal this is great case to improve that.

coderitual:
@dsmileybrainly I will close this PR as modal in current form is not something we would be happy to have in SG. The only benefit would be removing it from `ui` folder.

PR #1960, opened 2020-11-23
Accordion update

Malinicz:
- add variant for collapsed items in accordion
- enable Accordion to have several items opened by default

image
image

Desktop/mobile style approved by designers


src/components/accordion/AccordionItem.jsx
```diff
@@ -28,30 +35,47 @@ const AccordionItem = ({
...
+      payload: {id, value},
+    });
+  }
+
+  useEffect(() => {
+    if (defaultOpened) {
+      handleOpen(true);
+    }
+    //eslint-disable-next-line
+  }, [defaultOpened]);
```
coderitual, line 77: After rethinking we shouldn't use only one prop in this list. Let's keep it empty to indicate our intention (component did mount). Also allowing change of state after component is rendered would be against common behaviour of `default/initial` props

src/components/accordion/accordionReducer.js
```diff
@@ -0,0 +1,33 @@
+// @flow strict
```
coderitual, line 1: Can we move this reducer to accordion component itself? It's much easier to follow when such code is close to the component (to keep changes in sync). We can argue whether we also should move AccordionItem to the same file and it would be correct thinking as well

src/components/accordion/accordionReducer.js
```diff
@@ -0,0 +1,33 @@
...
+  type: 'accordion/SET_OPENED',
+  payload: {id: string, value: boolean},
+};
+
+export type StateType = $ReadOnly<{
+  opened: {
+    [key: string]: boolean,
+  },
+  noGapBetweenElements: boolean,
+  allowMultiple: boolean,
```
coderitual, line 13: let's not sync our props with state. Moving reducer to the component will help resolving the issue.

src/components/accordion/Accordion.jsx
```diff
@@ -43,34 +42,17 @@ const Accordion = ({
...
-  const handleChange = useCallback(
-    (item: string, value: boolean) => {
-      if (allowMultiple) {
-        setOpened({...opened, [item]: value});
-      } else {
-        const newOpened = {...opened};
+  const [state, dispatch] = useReducer(accordionReducer, {
+    opened: {},
+    noGapBetweenElements: spacing === 'none',
+    allowMultiple,
```
coderitual, line 48: This will break when allowMultiple prop changes

Malinicz: I think this is similar to `defaultOpened` - if we remove it from deps, it won't update if anyone decides that it should be dynamic. Here `allowMultiple` also should not be dynamic I guess (I can't find a use case when it should 🤔 )

src/components/accordion/Accordion.jsx
```diff
@@ -43,34 +52,34 @@ const Accordion = ({
...
+    opened: {},
+  });
 
-  const handleChange = useCallback(
-    (item: string, value: boolean) => {
-      if (allowMultiple) {
-        setOpened({...opened, [item]: value});
-      } else {
-        const newOpened = {...opened};
+  function reducer(state: StateType, action: ActionType): StateType {
```
bartoszlorek, line 59: I'd move reducer outside component scope

Malinicz: @coderitual ? 😄

coderitual: @bartoszlorek reducer uses component prop so it's easier to keep it that way. Alternatively we could have custom hook with reducer inside and pass prop parameters as hook params but imho this reducer is simple enough to leave it as it is.

bartoszlorek: or have separate actions for a different purposes and via different context dispatches proper one.

coderitual: What do you mean it's harder to test? Reducers should be tested separately. It's implementation detail of component. We should only test public api of component.

coderitual: @bartoszlorek we can jump on a call later if you want so we can discuss that :)

coderitual: @bartoszlorek moreover reducers from useReducer are working more in cheat mode. Please read:

https://twitter.com/dan_abramov/status/1102010979611746304

coderitual (approved):
👍

matzimowski (approved):

anna-wro (approved):

PR #1965, opened 2020-11-25
Add padding prop to AccordionItem

Malinicz:
Add padding prop to AccordionItem

### DIFF
--- a/src/components/accordion/AccordionItem.jsx
+++ b/src/components/accordion/AccordionItem.jsx
@@ -16,12 +16,15 @@ import Link from '../text/Link';
 import Text from '../text/Text';
 import {AccordionContext} from './Accordion';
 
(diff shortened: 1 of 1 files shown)


ElaLukaszewicz (approved):

src/components/accordion/AccordionItem.jsx
```diff
@@ -16,12 +16,15 @@ import Link from '../text/Link';
 import Text from '../text/Text';
 import {AccordionContext} from './Accordion';
 
+type PaddingType = 'xxs' | 'xs' | 's' | 'm' | 'l' | 'xl';
```
Malinicz, line 19: @coderitual do we want to decouple it from Box types or I should import it here?

coderitual: yep, let's keep it decoupled for now 👍

matzimowski (approved):

coderitual (approved):

PR #1969, opened 2020-11-26
Add bullet icon

Malinicz:
Add bullet icon to web icons set:

image

image

### DIFF
--- a/src/images/icons/bullet.svg
+++ b/src/images/icons/bullet.svg
@@ -1,7 +1,4 @@
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
   <title>bullet</title>
-  <g fill="none" fill-rule="evenodd">
-    <path d="M0 0h512v512H0z"/>
-    <circle fill="#000" cx="256" cy="256" r="64"/>
-  </g>
+  <circle cx="256" cy="256" r="64" fill-rule="evenodd"/>
 </svg>
--- a/src/images/icons/bullet.svg
+++ b/src/images/icons/bullet.svg
@@ -1,0 +1,6 @@
+<svg width="512" height="512" xmlns="http://www.w3.org/2000/svg">
+  <title>bullet</title>
+  <g fill="none" fill-rule="evenodd">
+    <path d="M0 0h512v512H0z"/>
+    <circle fill="#000" cx="256" cy="256" r="64"/>
+  </g></svg>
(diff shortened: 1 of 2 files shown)


src/images/icons/bullet.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
```
Malinicz, line 1: went through svgomg

bartoszlorek (approved):

src/images/icons/bullet.svg
```diff
@@ -0,0 +1,7 @@
...
+  <title>bullet</title>
+  <g fill="none" fill-rule="evenodd">
```
coderitual, line 3: can we remove this g entirely?

coderitual (approved):

coderitual:
@Malinicz Could you please add icon screenshot from storybook/style guide for release purposes?

src/images/icons/bullet.svg
coderitual: @Malinicz ignore that comment, looks we need it

coderitual: @Malinicz alternatively you can remove group along with path. Circle alone is enough

matzimowski:
@Malinicz could u please add a screenshot? 🍭

matzimowski (approved):

src/images/icons/bullet.svg
```diff
@@ -0,0 +1,7 @@
...
+    <path d="M0 0h512v512H0z"/>
+    <circle fill="#000" cx="256" cy="256" r="64"/>
```
coderitual, line 5: we need to remove fill color for web icons

coderitual (changes requested):
@Malinicz To make color overrides working properly, resolving commented issues should do the job (removing fill from cricle, remove group and remove path)
image

src/images/icons/bullet.svg
Malinicz: @coderitual I wanted to set fill as `currentColor`, but then I checked other icons and most of them have fill set to `#000`. However, I left below and it seems to work fine:

```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
  <title>bullet</title>
  <circle cx="256" cy="256" r="64" fill="currentColor" fill-rule="evenodd"/>
</svg>
```

coderitual: @Malinicz could you give me an example of web icon having fill set to #000? if there is any it's a bug

coderitual (approved):
❤️  thanks!

I confirm there are other icons with fill color set to #000. This is of course a bug so you might have been tricked because of them.

PR #1971, opened 2020-11-27
Accordion update

Malinicz:
- set pointer only on accordion item header
- fix issue with turning off hover state after click when mouse is still on hover area

Screen Recording 2020-11-27 at 11 43 19


anna-wro (approved):

matzimowski (approved):
lgtm

coderitual (approved):

PR #1979, opened 2020-12-03
Adjust Accordion to meet accessibility requirements

Malinicz:
- enable accordion items to be expanded/collapsed with space and enter
- disable focus on collapsed content
- avoid unnecessary collapsing on the first render and fix the issue with defaultOpened prop not working
- display accordion item outline on focus only when focused by keyboard
- add accessibility properties needed for screen readers per https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html


src/components/accordion/AccordionItem.jsx
```diff
@@ -40,40 +41,52 @@ const AccordionItem = ({
   className = '',
   defaultOpened = false,
   padding = 'm',
+  tabIndex = 0,
 }: PropType) => {
+  const hasRendered = useRef(false);
```
Malinicz, line 46: I added this to avoid unnecessary collapsing on first render

coderitual (changes requested):
thanks @Malinicz !

It looks we also need to resolve width issue. This is how it's gonna look on production.
image

src/components/accordion/_accordion.scss
```diff
@@ -1,4 +1,8 @@
 @if ($includeHtml) {
+  .sg-accordion-container {
+    width: 100%;
```
coderitual, line 3: hey @Malinicz , let's remove 
```
.sg-accordion-container {
    width: 100%;
}
```

and keep it as default (display block).

coderitual:
@Malinicz it look that in brainly frontend we need to slightly change parents of accordion and this should be enough. Accordion itself behaves correctly (like normal div it will occupy width if there is a room in paren)

coderitual:
Hey @Malinicz I changed layout for this particular story to better fit component's needs.
Screenshot 2020-12-03 at 18 00 33
Screenshot 2020-12-03 at 18 00 40

coderitual (approved):
👍

src/components/accordion/Accordion.jsx
```diff
@@ -77,9 +143,21 @@ const Accordion = ({
...
+        opened: state.opened,
+        focusedElementId: state.focusedElementId,
+        dispatch,
+      }}
     >
-      <div className={cx(spaceClass, className)}>{children}</div>
+      <div
+        ref={wrapperRef}
+        className={cx(spaceClass, className)}
+        data-allow-multiple={allowMultiple}
```
matzimowski, line 156: sorry if i missed sth but are these two: `data-allow-multiple`, `data-allow-toggle` used somehow? or there is some a11y for having them?

matzimowski (approved):
amazing job @Malinicz  👏

src/components/accordion/Accordion.jsx
bartoszlorek: used by screen readers

bartoszlorek (approved):

src/components/accordion/Accordion.jsx
matzimowski: 👏 thanks! perfect!

anna-wro (approved):

PR #1985, opened 2020-12-04
Fix Accordion item scrolling issue part II

Malinicz:
Turned out preventDefault is not needed, what we need is `overflow-anchor: none` 😅

### DIFF
--- a/src/components/accordion/Accordion.jsx
+++ b/src/components/accordion/Accordion.jsx
@@ -79,7 +79,13 @@ const Accordion = ({
       const key = KEY_CODES[event.keyCode];
 
       if (['space', 'enter'].includes(key)) {
-        event.preventDefault();
+        if (
+          event.target instanceof HTMLElement &&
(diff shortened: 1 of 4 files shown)


coderitual (approved):

coderitual:
@Malinicz wait a sec please :)

coderitual (changes requested):
@Malinicz  Please also remove: ` event.preventDefault();` from key event. It breaks a11y on children element (eg you cannot hit enter on link, input or form)

coderitual:
@Malinicz we need to check if we are opening correct item.

We can achieve this by this condition:

```
      if (['space', 'enter'].includes(key)) {
        if (event.target.id === state.focusedElementId) {
          event.preventDefault();
        }
        dispatch({type: 'accordion/KEYBOARD_SET_OPENED'});
      }
```

coderitual (approved):

PR #1988, opened 2020-12-07
Add forwardRef to Box and Flex

Malinicz:
Without `forwardRef` we are not able to use `ref` on those elements

### DIFF
--- a/src/components/accordion/Accordion.spec.jsx
+++ b/src/components/accordion/Accordion.spec.jsx
@@ -2,7 +2,6 @@ import React from 'react';
 import {mount} from 'enzyme';
 import Accordion from './Accordion';
 import AccordionItem from './AccordionItem';
-import Box from '../box/Box';
 
 describe('<Accordion>', () => {
   it('renders', () => {
@@ -125,7 +124,7 @@ describe('<Accordion>', () => {
 
     expect(
       accordion
-        .find(Box)
+        .find('Box')
         .at(0)
         .hasClass('sg-accordion-item--no-gap')
     ).toBe(true);
@@ -142,7 +141,7 @@ describe('<Accordion>', () => {
 
     expect(
       accordion
-        .find(Box)
+        .find('Box')
         .at(0)
         .prop('borderColor')
     ).toEqual('gray-secondary-lightest');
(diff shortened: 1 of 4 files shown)


ElaLukaszewicz (approved):

src/components/accordion/Accordion.spec.jsx
```diff
@@ -124,7 +125,7 @@ describe('<Accordion>', () => {
 
     expect(
       accordion
-        .find('Box')
+        .find(Box)
```
coderitual, line 128: how about setting Box.displayName = "Box", if I understand correctly it uses displayName to find component without a need of importing real component

Malinicz, line 128: this is something worth to remember, when forwardRef is used on element we can't find by string in tests :)

Malinicz: good point 🤔  Haven't thought about it, but actually this seems like a better solution

coderitual: @Malinicz you can wrap setting display name in env check and only set in on development /tests.

something like: 

```
if(dev) {
...
}

where dev is ofc check for `process.env.NODE_ENV`
```

Malinicz: yeah but then I would have the same problem in brainly-frontend, right?

bartoszlorek (approved):

src/components/accordion/Accordion.spec.jsx
Malinicz: and if I just set displayName on component, I also would be able to find by string in our repo, or I'm wrong @coderitual?

coderitual: @Malinicz you should be able to find it in our repo too as long as we don't set production flag for tests in brainly frontend. This code will be present in production build of style guide and only removed on production build of application uses style guide.

coderitual: @Malinicz So setting `export const __DEV__ = process.env.NODE_ENV !== "production"` in some utils file and importing it into component file should work. This also should work with application which uses our component. So in brainly frontend we can test component using strings as well

matzimowski (approved):

coderitual (approved):

PR #1990, opened 2020-12-08
Select - remove custom line-height (a11y fix)

matzimowski:
It's changed to fulfill the following obligation:

"Text can be resized without assistive technology up to 200 percent without loss of content or functionality, except for captions and images of text."

ref: https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html

checked on: chrome, safari, firefox, edge and ie11

moreover, setting `line-height` there didn't result in any visual change for default (100%) size.

## before:

**when 200% of text resize**

Screenshot 2020-12-08 at 09 44 15

**when 100% of text size** (no influence)

Screenshot 2020-12-08 at 09 44 29

## after:

**when 200% of text resize**
Screenshot 2020-12-08 at 09 43 49

**when 100% of text size** (no influence)
Screenshot 2020-12-08 at 09 44 21

### DIFF
--- a/src/components/form-elements/_select.scss
+++ b/src/components/form-elements/_select.scss
@@ -16,7 +16,6 @@ $includeHtml: false !default;
       font-size: $formElementDefaultFontSize;
       color: $formElementPlacholderTextColor;
       height: $formElementSizeM;
-      line-height: $formElementSizeM - 4px; // subtraction due to top and bottom border
       background: $formElementDefaultBackgroundColor;
       border-radius: $formElementStandardBorderRadius;
 
@@ -97,7 +96,6 @@ $includeHtml: false !default;
 
       .sg-select__element {
         height: $formElementSizeL;
-        line-height: $formElementSizeL - 4px; // subtraction due to top and bottom border
         font-size: $formElementLargeFontSize;
         border-radius: $formElementLargeBorderRadius;
         padding: 0 50px 0 spacing(m);


src/components/form-elements/_select.scss
```diff
@@ -97,7 +96,6 @@ $includeHtml: false !default;
 
       .sg-select__element {
         height: $formElementSizeL;
-        line-height: $formElementSizeL - 4px; // subtraction due to top and bottom border
```
coderitual, line 100: @mateuszs what if we use percentage value?

matzimowski: i observed that it's not needed at all. do you see any need to define it using percentage value?
if it's not set here it gets default value equals `1.5rem` (inherited from `body`).

coderitual: @matzimowski yea, by percentage I meant unit less (which probably is the same :upside_down_face: ), I was wondering why was added but if it works that way we're good!

coderitual (approved):

ElaLukaszewicz (approved):

kubajasiolek:
@matzimowski FYI, this PR introduced a bug that makes birthday input not visible on regform on PL.
Actionlists sets line-height to 0, which previously was overwritten by the above, and now is not, so stays at 0.
@lukaszduraj found the bug, and will report to PBUGS soon.

lukaszduraj:
image
image

coderitual:
Thanks for letting us know @kubajasiolek !

I am not sure why action list is setting line-height to 0. we can consider replacing it with flex in frontend app. wdyt cc @matzimowski ?

matzimowski:
@coderitual oh, sorry for the bug, yea, i would prefer to use Flex instead, i will ry to fix it in our app

coderitual:
@matzimowski thanks Mati! let me know if you have any issues. We can fix it here in the style guide in that case.

PR #1995, opened 2020-12-09
Fix reduce motion check

coderitual:
Apparently on ios12 `window.matchMedia('(prefers-reduced-motion: reduce)')` returns undefined instead of proper match media object (looks like a bug, matchMedia is supported there). We need to have extensive cheking. Additionally `(prefers-reduced-motion)` is added as some devices supports only this shorter form


matzimowski (approved):
🌞

PR #1997, opened 2020-12-10
Fix reduce motion check vol. 2

matzimowski:
It fixes:
`TypeError: mediaQuery.addEventListener is not a function. (In 'mediaQuery.addEventListener('change', handleChange)', 'mediaQuery.addEventListener' is undefined)` error that is thrown on devices with  iOS v12, but also on desktop ie. Firefox v52.

According to: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener:

> "Older browsers should use addListener instead of addEventListener since MediaQueryList only inherits from EventTarget in newer browsers."

at the same time it should still work for newer browsers.

### DIFF
--- a/src/components/utils/useReducedMotion.js
+++ b/src/components/utils/useReducedMotion.js
@@ -29,9 +29,9 @@ export default function useReducedMotion(): boolean {
     };
 
     handleChange();
-    mediaQuery.addEventListener('change', handleChange);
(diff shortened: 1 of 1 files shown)


coderitual (approved):
❤️

ElaLukaszewicz (approved):

PR #2003, opened 2020-12-17
Use dart-sass

seahindeniz:
> [LibSass and Node Sass are deprecated](https://sass-lang.com/blog/libsass-is-deprecated)

Quoting from [here](https://github.com/sass/node-sass#node-sass)
Installation of node-sass kind of failing on some environments, so it's best to avoid that lib might be a good option.


coderitual:
hey @seahindeniz , thanks for the PR! We have another internal CI/CD which builds style guide and before we go with this kind of change we need to be sure we don't break current pipeline. 

I am happy to move away from it but this should be planned. Also I am not sure what we gain with this change. Please note that bugs are still fixed. https://sass-lang.com/blog/libsass-is-deprecated#what-does-quot-deprecated-quot-mean

seahindeniz:
Hi @coderitual,
Yes I agree it may not change too many things but the node-sass and node-gyp has issues with installation. Since both node-sass and dart-sass have the same API and, there is that depreciation label out there, I tought it would be good to switch from it.

I have already done my search for the installation issue with node-sass and result was not quite good 
https://github.com/sass/node-sass/issues/2528

coderitual:
Hey @seahindeniz I am happy to get back to this topic after new year. I definitely see benefits from. using recommended compiler but we need deeply investigate other factors like build performance for production and local development compatibility.

See you after holiday break!

coderitual:
Resolved in #2342

PR #2006, opened 2020-12-22
Add Lists to Storybook

Malinicz:
- add list and menu list to Storybook

I split list into 2 menu items as I think it makes more sense (they are separate components)

image
image


coderitual:
@Malinicz I agree, it makes sense to have 2 different stories. Could you also add subcomponent section to each story so users could also see `ListItem` and `MenuListItem` documentation next to their parent componetns?

Here is an example I added to accordion:
https://github.com/brainly/style-guide/blob/master/src/components/accordion/Accordion.stories.jsx#L11

Thanks!

coderitual (approved):
🍺

PR #2007, opened 2020-12-22
Add Breadcrumbs component to Storybook

Malinicz:
Add Breadcrumbs component to Storybook

image


coderitual:
@Malinicz thanks!

Could we please showcase more real world example. Current sg docs are quite nice:

image

We shouldn't use long text as our default version but only as a edge case example (preferably the last from examples)

Do we need these argTypes?
```
  short: {control: 'boolean'},
  adaptive: {control: 'boolean'},
  inlineItems: {control: 'boolean'},
```
We only need them if we modify something.

coderitual (approved):
🎄

PR #2014, opened 2021-01-15
Add label to Radio and Checkbox

coderitual:
## Changes
This is enabler for further development.

- Adding built-in label to `checkbox` and `radio` components, making upcoming implementation of redesign (almost finished by designer) a bit easier from this point. This is backward compatible with current solution if label is not provided.
- Removing deprecated label from radio and checkbox examples.

image

image

image

image

image


Malinicz (approved):

matzimowski (approved):
👏

PR #2016, opened 2021-01-29
Generate TypesScript Library Declaration

coderitual:
# Change

This adds fully automated generation of single file typescript library declaration in the form of public API - `brainly-style-guide.d.ts`

https://user-images.githubusercontent.com/8572321/106401121-b13a8480-6422-11eb-9462-fd3228dcd534.mp4

## Important notes ✅
- Whole Style Guide api is available as a one declaration file (the same as in [definitely-typed](https://github.com/DefinitelyTyped/DefinitelyTyped))  There is no need to access internal folder structure for types and hopefully we could avoid importing from path (eg. `import {IconTypeType as SubjectIconType} from 'brainly-style-guide/src/components/subject-icons/SubjectIcon';`) since it makes internal folder layout to be part of the api. Importing constants from style guide has similar downside and using literals might be preferred way. Common types will be explicitly exported with unique name and available as a style guide globals (e.g. icon and text related types, props types)
(see [Caveats](##Caveats))
- All components declaration and their PropsTypes exposed (each prop type has unique name now).
- Script is incorporated into CI so we can quickly catch future incompatibilities after new code is added.
- Type generation script runs locally and it is a part of the npm publishing process.
- Codebase remains JSX/Flow based until we fully switch to TS.
- There are still a few things to be fixed in the near future but I would like to test types declaration as soon as possible.  (see [Caveats](##Caveats))

## Process 💡

image

→ Manual changes to `.js/.jsx` files whenever change is compatible between Flow and TypeScript.
→ Codemod and AST adjustments resolve problems which cannot be resolved simply by changing source files.
→ flow-to-ts produces `.ts/.tsx` from `.js/.jsx`
→ Typescript compiler produces `d.ts` files for each source `.ts/.tsx`.
→ API Extractor rollups all declaration files into one public API declaration

## Caveats 🦄
- While all typings are available now, some of them have strange names. **This will be resolved along the way**. Those types though still provide perfectly correct definition. It's minor issue atm and mostly visible during extending types.  
**For now only public types with unique name are exported.**
**Details:**
→ All types became publicly available. It means that they should have unique names. Unfortunately some components expose their types without worrying about uniqueness of the type. For instance `PropTypes` is often exported from component and later extended in application. This will be fixed in codemod step to not create too many breaking changes to the code (I expect only small breaking changes with flow - for example mixed type returned by event handlers is not compatible with ts)
→ Some types are defined multiple times and have exactly same meaning. This results in redundant names and causes extractor to autogenerate suffixes. Sometimes we will solve such a problem in a similar way to the previous point. Other times, providing a more general type will be a better choice.
- There are still some errors in autogenerated typescript files (~12). Those are internal component problems and don't affect public api declaration so are treaded like warnings. API extractor is to decide whether api was successfully produced based on TS declaration files.


bartoszlorek (approved):
looks good to me

dsmileybrainly (approved):

webDevAdvocate (approved):

Fasosnql (approved):

src/components/bubble/Bubble.jsx
```diff
@@ -21,17 +20,43 @@ type AligmentType = 'start' | 'center' | 'end';
 
 type DirectionType = 'left' | 'right' | 'top' | 'bottom';
 
-export const ALIGNMENT = {START: 'start', CENTER: 'center', END: 'end'};
-export const DIRECTION = {
+export const ALIGNMENT: {
```
matikrk, line 23: missing here TS `as const` :) 

How about trick with `Object.freeze()`? I know right now you made it ok, but have you tested it?

coderitual: hey @matikrk, thanks!, Object.freeze is not working as expected with TS neither with flow-to-ts (I hoped that freeze will be converted to as const). I tested `as const` and it works nice.

I started from codemod for this but I decided to get back to compatible solution over codemod becuase I would like to avoid using those const at all.

scripts/build-types.js
```diff
@@ -0,0 +1,162 @@
...
+    DEST_DIR,
+    relativeSourceFile.replace(sourceExtension, destinationExtension)
+  );
+
+  fs.outputFileSync(outputFile, typescriptCode, noop => noop);
+});
+
+const tsFiles = glob.sync('.typescript/**/*.{ts,tsx}');
+
+const options = {
```
matikrk, line 72: Can we extend from  `tsconfig.json`?

matikrk (approved):

scripts/build-types.js
coderitual: good idea

coderitual: @matikrk done

matikrk (approved):

PR #2017, opened 2021-02-01
Add extra large breakpoint to scss

coderitual:
Exposing extra large breakpoint which is part of our design system.

Available as 
- **scss variable: `$breakpointXLarge`**
- **mobile first mixin: `sgResponsive(xl)`** - this should be preferred way. Mobile first approach.
- class: `sg-hide-for-xlarge-up`
- mixin `sgBreakpoint(xlarge-up)` -  this mixin shuld be avoided as it mixes 3 different approaches (strict resolution, mobile first and desktop first)

Notes:
- **`sgBreakpoint`, `sg-hide-for-xxx` classes and `RwdHelper` (uses internally sg-hide) are marked as deprecated**
- I added `large-up` to breakpoint map which is the same as wrongly named `large-only`. I leave `large-only` as it is for backward compatibility but `large-up` should be used from now on.

### DIFF
--- a/src/sass/_config.scss
+++ b/src/sass/_config.scss
@@ -137,7 +137,7 @@ $breakpointsMap: (
   'medium-only': '(min-width: #{$breakpointMedium}) and (max-width: #{$breakpointLarge - 1px})',
   'medium-down': '(max-width: #{$breakpointLarge - 1px})',
   'medium-up': '(min-width: #{$breakpointMedium})',
-  'large-only': '(min-width: #{$breakpointLarge})' // deprecated
+  'large-only': '(min-width: #{$breakpointLarge})',
   'large-up': '(min-width: #{$breakpointLarge})',
   'xlarge-up': '(min-width: #{$breakpointXLarge})',
 );
--- a/src/sass/_config.scss
+++ b/src/sass/_config.scss
@@ -133,6 +133,7 @@ $breakpointsMap: (
   'medium-only': '(min-width: #{$breakpointMedium}) and (max-width: #{$breakpointLarge - 1px})',
   'medium-down': '(max-width: #{$breakpointLarge - 1px})',
   'medium-up': '(min-width: #{$breakpointMedium})',
+  'large-down': '(max-width: #{$breakpointLarge - 1px})',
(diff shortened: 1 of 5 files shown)


anna-wro (approved):

src/sass/_config.scss
```diff
@@ -126,26 +126,31 @@ $toplayerZIndex: 1000;
...
 $breakpointLarge: 1024px !default;
+$breakpointXLarge: 1280px !default;
 
 $breakpointsMap: (
   'small-only': '(max-width: #{$breakpointMedium - 1px})',
   'medium-only': '(min-width: #{$breakpointMedium}) and (max-width: #{$breakpointLarge - 1px})',
   'medium-down': '(max-width: #{$breakpointLarge - 1px})',
   'medium-up': '(min-width: #{$breakpointMedium})',
-  'large-only': '(min-width: #{$breakpointLarge})'
+  'large-only': '(min-width: #{$breakpointLarge})',
```
Malinicz, line 136: since we are adding xl now, you don't have to add max-width here, like for medium-only?

bartoszlorek: that's true

coderitual: @Malinicz This is breaking change and we can do that in separate PR. A lot of places uses `large-only` in the meaning of `large-up`

Malinicz: ok I see, sure let's do that separately 👍

Malinicz (approved):

src/sass/_config.scss
bartoszlorek: Second thing, for `sg-hide` class, we need `sg-hide-for-large-DOWN` because it works in opposite way. It should hide content until wide of the screen is XL.

coderitual: @bartoszlorek makes sense

coderitual: @bartoszlorek added `large-down` to generate `sg-hide-for-large-down` class

matzimowski (approved):

bartoszlorek (approved):

PR #2023, opened 2021-02-11
Add utility classes for fade-in animation

Malinicz:
Add 2 classes for fade-in animation, widely used in brainly-frontend app.

### DIFF
--- a/src/sass/_utils.scss
+++ b/src/sass/_utils.scss
@@ -31,3 +31,25 @@
 }
 
 // sass-lint:enable no-important
+
+@keyframes sgAnimationFadeIn {
+  0% {
+    opacity: 0;
+  }
+
+  100% {
+    opacity: 1;
+  }
+}
+
+.sg-animation-fade-in-slow {
+  animation-name: sgAnimationFadeIn;
+  animation-duration: 600ms;
+  animation-timing-function: ease-in;
+}
+
+.sg-animation-fade-in-fast {
+  animation-name: sgAnimationFadeIn;
+  animation-duration: 400ms;
+  animation-timing-function: ease-in;
+}


matzimowski (approved):
lgtm

src/sass/_utils.scss
```diff
@@ -31,3 +31,25 @@
...
+  0% {
+    opacity: 0;
+  }
+
+  100% {
+    opacity: 1;
+  }
+}
+
+.sg-animation-fade-in-slow {
```
matzimowski, line 45: what about fadeOut? fadeIn gonna miss it 🤗

src/sass/_utils.scss
```diff
@@ -31,3 +31,25 @@
...
+  animation-name: sgAnimationFadeIn;
+  animation-duration: 600ms;
+  animation-timing-function: ease-in;
+}
+
+.sg-animation-fade-in-fast {
+  animation-name: sgAnimationFadeIn;
+  animation-duration: 400ms;
+  animation-timing-function: ease-in;
```
coderitual, line 54: I found only 3 usages of this particular easing in application which is kind of wrong. Animation principles sugges easing out or in out like for things we're showing, can we please discuss? :)

src/sass/_utils.scss
coderitual: @matzimowski that's true :) please keep in mind this is kind of temporary. To cover fade out properly we need transition component taking care of keeping instance until animation ends. I hope we get there when we will have motion designer :)

src/sass/_utils.scss
coderitual: ah probably you meant usage of these: 
image

coderitual (approved):

bartoszlorek (approved):

PR #2025, opened 2021-02-12
Update docs with information about new animation classes

Malinicz:
image


coderitual (approved):
I just realised that the name is wrong. This is transition (it applies to class names as well). Let's keep it that way though and we refactor later. We gonna have different section for animations

PR #2028, opened 2021-02-24
Add cup icon to web icons set

Malinicz:
image


matzimowski (approved):
🥤  🥇

src/images/icons/cup.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
Malinicz, line 1: I removed `fill` from here (comparing to mobile icon), should I actually do that @coderitual?

bartoszlorek (approved):

src/images/icons/cup.svg
coderitual: great, we don't use `fill` one color icons

coderitual (approved):
✏️

PR #2038, opened 2021-03-26
Add second option for filters icon

patrycja-b:
Screenshot 2021-03-26 at 12 16 54


src/components/icons/Icon.jsx
```diff
@@ -36,6 +36,7 @@ export type IconTypeType =
   | 'exclamation_mark'
   | 'facebook'
   | 'filters'
+  | 'filters_2'
```
matzimowski, line 39: maybe we could consider some different name? ie. `funnel`? wdyt?

coderitual: let's wait a bit, I will consult naming. This icon will show up in design tools as well so let's make sure we use the best name :)

patrycja-b: Yeah, makes sense to me. I focused on the usage in this case (as probably this will be our second option for filters) but I think `funnel` is way better name 😅  @coderitual  ?

coderitual: Let's go with @matzimowski propose - `funnel`

matzimowski: i feel very honoured 🥨

patrycja-b: done ✅

matzimowski (approved):
fantastic 🥇

coderitual (approved):
🐦 👍

PR #2051, opened 2021-05-14
[WIP] Accordion external control

patrycja-b:


src/components/accordion/AccordionItem.jsx
```diff
@@ -23,6 +23,7 @@ export type AccordionItemPropsType = $ReadOnly<{
   defaultOpened?: boolean,
   padding?: PaddingType,
   tabIndex?: number,
+  customId?: string,
```
coderitual, line 26: just `id` maybe? customId sounds like a small leaky abstraction

src/components/accordion/Accordion.jsx
```diff
@@ -80,14 +88,53 @@ const Accordion = ({
...
+  const isControlled = useRef<boolean>(index !== undefined && !!onChange);
   const [state, dispatch] = useReducer(reducer, {
     opened: {},
     focusedElementId: null,
   });
   const hasReduceMotion = useReducedMotion() || reduceMotion;
 
+  useEffect(() => {
+    const isCallbackMissing = index !== undefined && !onChange;
+    const isComponentChangedToUncontrolled = isControlled.current && !index;
```
coderitual, line 104: I think we need to be more explicit here - `typeof index !== "undefined" - to avoid false positives (`index = 0`).

src/components/accordion/Accordion.stories.jsx
```diff
@@ -100,3 +100,32 @@ export const NoGaps = () => (
...
+    ))}
+  </Accordion>
+);
+
+Controlled.args = {
+  index: currentAccordionId,
+};
+
+Controlled.argTypes = {
+  index: {control: {type: 'select', options: CONTROLLED_ACCORDION_IDS}},
```
coderitual, line 130: 🚀

src/components/accordion/Accordion.stories.jsx
```diff
@@ -100,3 +100,32 @@ export const NoGaps = () => (
...
+  'accordion item 3',
+];
+
+let currentAccordionId = null;
+const setCurrentAccordionId = id => {
+  currentAccordionId = id;
+};
+
+export const Controlled = (args: any) => (
+  <Accordion onChange={setCurrentAccordionId} {...args}>
```
coderitual, line 116: hey, just fyi, this shouldn't work :) - calling `setCurrentAccordionId`  has no impact as this is function outside component (it won't cause component to rerender as in useState/useReducer). This code works only because we're overwriting state in accordion reducer.

I will fix this in review pr so no action from your side is required

patrycja-b: 🤦  You are totally right, I overlooked that. I see - it works only because we change `index` prop via controls.

src/components/accordion/Accordion.spec.jsx
```diff
@@ -163,4 +163,101 @@ describe('<Accordion>', () => {
...
+      </Accordion>
+    );
+
+    expect(accordion.find('[aria-labelledby]').hostNodes()).toHaveLength(
+      accordionIds.length
+    );
+    expect(accordion.find('[aria-expanded=true]').hostNodes()).toHaveLength(1);
+  });
+
+  it('expands controlled item when index is type of string', () => {
```
coderitual, line 203: I think this test is duplicated :)

patrycja-b:
The work is moved to https://github.com/brainly/style-guide/pull/2053

PR #2052, opened 2021-05-19
Adjust title prop for AccordionItem

patrycja-b:
`title` is type of `React.Node`

For the most common case when `title` is `string`  we wrap it in `Link` component to have default styling:
Screenshot 2021-05-19 at 11 34 34

For other cases we allow it to stretch to the whole width:
Screenshot 2021-05-19 at 11 34 42

### DIFF
--- a/src/components/accordion/Accordion.spec.jsx
+++ b/src/components/accordion/Accordion.spec.jsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {mount} from 'enzyme';
 import Accordion from './Accordion';
 import AccordionItem from './AccordionItem';
+import Link from '../text/Link';
 
 describe('<Accordion>', () => {
   it('renders', () => {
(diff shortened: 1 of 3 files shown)


src/components/accordion/Accordion.spec.jsx
```diff
@@ -163,4 +163,28 @@ describe('<Accordion>', () => {
     // hostNodes returns html elements and skip react components
     expect(accordion.find('[aria-expanded=true]').hostNodes()).toHaveLength(2);
   });
+
+  it('renders span with link styles when title is string', () => {
```
coderitual, line 167: `span` in this context is saying more than test does, maybe we can rephrase it?

src/components/accordion/Accordion.spec.jsx
```diff
@@ -163,4 +163,28 @@ describe('<Accordion>', () => {
...
+  it('renders span when title is not string', () => {
+    const accordion = mount(
+      <Accordion>
+        <AccordionItem title={<div>info</div>}>
+          Accordion Item Description
+        </AccordionItem>
+      </Accordion>
+    );
+
+    expect(accordion.find('.sg-text--link').exists()).toBe(false);
```
coderitual, line 187: How about checking with `find(Link)` (of course we need to import Link to test or use string literal `Link`)

coderitual (approved):

src/components/accordion/Accordion.spec.jsx
patrycja-b: yes, makes more sense 👍 
https://github.com/brainly/style-guide/pull/2052/commits/41349ae98be711232f036fdd6fbdbe915bdbf91a

src/components/accordion/Accordion.spec.jsx
patrycja-b: https://github.com/brainly/style-guide/pull/2052/commits/41349ae98be711232f036fdd6fbdbe915bdbf91a

coderitual:
@patrycja-b good work! Do you think is it worth to add/change one of the stories to show customised title - maybe based on your product example?

coderitual:
after conversation with @patrycja-b we decided to not add customised title example, from my perspective it's ready to merge

PR #2057, opened 2021-05-29
Add loading state to button

coderitual:
## Summary
- added loading state to button (using spinner adjustments from https://github.com/brainly/style-guide/pull/2055)
- unify props by replacing discriminated unions with simple types and invariants
- improve story docs (all props are editable, icon prop selectable from list)

image
image


bartoszlorek (approved):

maciekbb (approved):

ziejka (approved):

simka (approved):

src/components/buttons/Button.jsx
```diff
@@ -1,7 +1,10 @@
 // @flow strict
 
 import * as React from 'react';
+import Spinner, {SPINNER_SIZE, SPINNER_COLOR} from '../spinner/Spinner';
 import cx from 'classnames';
+import {__DEV__} from '../utils';
```
matikrk, line 6: why we don't have index.js in this folder to export invariant?

src/components/buttons/Button.jsx
```diff
@@ -62,45 +56,57 @@ export const BUTTON_TOGGLE: {
...
+  | 'solid-inverted'
+  | 'solid-blue'
+  | 'solid-mint'
+  | 'transparent-peach'
+  | 'transparent-mustard'
+  | 'transparent-blue'
+  | 'transparent-inverted'
+  | 'facebook';
+
+type ButtonToggleType = 'peach' | 'mustard' | 'blue';
```
matikrk, line 96: Pls drop it someday 🙏  I would even move it directly to ButtonType

matikrk (approved):

src/components/buttons/Button.jsx
coderitual: this is not as easy as moving it to button type. It behaves differently for different types (sometimes we respect 3 values, sometimes only one is correct). definitely something to be further discussed though but more aboout having 2 buttons instead of 1 which we effectively have inside button. 

image

src/components/buttons/Button.jsx
coderitual: good point, done

radekwarisch (approved):

kubajasiolek (approved):

ensisf (approved):

anna-wro (approved):

PR #2067, opened 2021-06-22
Extend select with grouped options

ensisf:
### Summary
For now existed custom component `Select` has no possibility to group options as regular HTML `select` has
using `<optgoup>` tag. This PR adds the possibility to render "grouped" options in select without breaking changes by extending the currently existed `option` prop.

#### Changelog:
- prop `option` extended with object type for grouped elements;
- updated README.md;
- added tests;
- added interactive and static examples;

Screenshot 2021-06-22 at 12 52 40


src/components/form-elements/Select.jsx
```diff
@@ -103,6 +122,14 @@ const Select = (props: SelectPropsType) => {
     };
   }
 
+  if (options.length && groupedOptions.length) {
+    throw {
+      name: 'WrongOptions',
+      message:
+        'Select: should passed be either options or groupedOptions prop!',
+    };
+  }
```
patrycja-b, line 131: Maybe we can use `invariant` function here, @coderitual wdyt?

src/components/form-elements/Select.jsx
```diff
@@ -115,11 +142,16 @@ const Select = (props: SelectPropsType) => {
...
+
+  const optionsElements =
+    (options.length && options.map(getOptionElement)) ||
+    (groupedOptions.length &&
+      groupedOptions.map(({options, label}) => (
+        <optgroup key={label} label={label}>
+          {options.map(getOptionElement)}
+        </optgroup>
+      ))) ||
+    [];
```
patrycja-b, line 154: Wdyt about splitting this into consts to increase readability? 

```suggestion
  const flatOptionsElements = (options.length && options.map(getOptionElement));
  const groupedOptionsElements = (groupedOptions.length &&
      groupedOptions.map(({options, label}) => (
        <optgroup key={label} label={label}>
          {options.map(getOptionElement)}
        </optgroup>);
        
  const optionsElements = flatOptionsElements || groupedOptionsElements || [];
```

src/components/form-elements/pages/select.jsx
```diff
@@ -57,6 +65,14 @@ const selects = () => (
     <DocsBlock info="Full width">
       <Select {...exampleProps} fullWidth />
     </DocsBlock>
+    <DocsBlock info="With grouped options">
+      <Select
+        {...exampleProps}
+        options={[]}
```
patrycja-b, line 71: I can see that `options` are optional, so we can remove this line not to confuse people how to use it properly.

src/components/form-elements/Select.jsx
ensisf: This was my first solution. I did not find it more readable at that moment, it just enforces to create 1 extra unnecessary variable. But up to you, I do not mind changing it back.

src/components/form-elements/Select.jsx
coderitual: yes, exactly

patrycja-b: @ensisf you can check the example usage of this function in `Accordion` component.

src/components/form-elements/Select.jsx
```diff
@@ -75,13 +81,25 @@ export type SelectPropsType = {
    * @example <Select size="m" options={[{value: 'option1', text: 'Option1'},{value: 'option2', text: 'Select selector'}]} />
    */
   options?: Array<OptionsPropsType>,
+  /**
+   * Optional array of grouped options of the select
+   * @example <Select size="m" options={[{"label": "Label title","items": [{ "value": "option1", "text": "Option1" },{ "value": "option2", "text": "Select selector" }]}]
+} />
+   */
+  groupedOptions?: Array<GroupedOptionsPropsType>,
```
coderitual, line 89: Hey @ensisf , I think we need to figure out better interface than this. If we need to mimic native solution (and this is great opportunity to do it). We need to allow options to be grouped and without groups at the same time. 

I would be more to allow having `option` and `optgroup` as children of select but this would change completely your implementation. If we want to keep current way of passing options we need to extend current prop.

We could jump on a call if you you'd like to.

coderitual (changes requested):

src/components/form-elements/Select.jsx
```diff
@@ -115,11 +128,22 @@ const Select = (props: SelectPropsType) => {
...
-  const optionsElements = options.map(({value, text}) => (
-    <option key={value} value={value}>
-      {text}
-    </option>
-  ));
+
+  const optionsElements = options.map(item => {
+    if (typeof item.label === 'string' && Array.isArray(item.options)) {
+      return (
+        <optgroup key={item.label} label={item.label}>
```
ensisf, line 135: @coderitual I think I need your help here :)
After checks in line 133, flow still do not distinguish which array element type is here - should be GroupedOptionsPropsType 

Screenshot 2021-06-25 at 09 47 40

[Here](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDyOGAlnAHYAquBAvGAN6phgZ4AeGAXGAM4YATqQoBzANwswANwCGMAK55eA4WMkBfdNnxgA4oLgL8AE2JlKNXQ2asYsgEZ4YKoSIlS4JchT68AgoKCslgAPOY+VngAfJratETelHxRYAyBwWERlgkAPvqGxnhmSdS00egAxskYYJyyALY4MHjZvmlgANpSjHKKymAA5F4WFACMAEwAzIMANGycPENtY4Masz19SrzDpRMTcwtcOwDKzniVtXznl3CCaxustqxg9k4uQwAyjs5HGHNSVgjHx+LqAl69eTbIbAyiTKYAFmmh3Yx2WpVW63BTy2A12ZGmCMo+xRi1ONyuFLuD3BAF1HmAsTioXjYRRpjN5qiloMzi1LvwqfcsbTJKhqr5ag1ZDhTG0+B0ABRs0HyqIASjS0SYUkEeAwCkEFDAKoAdNKcIrSOwGlqdS8wKQoIqdHg4FBHTbTW9fnQ-UNVO5BmAAGQhsAZEKm0h8SNYK1elXqzXPB16g1Gz14Bren4wMDgrTgp1gBPZ03c0Ph63l3Ep7FgdOG42ph1-Xg1nPchltmQsjte3E9l5aB2jl5NzMUBQwGDoVgadWoDRAA) is an example in flows playground,  but there is everything ok.
I was struggling with this some time and for now dont have any idea how to fix it.

coderitual: @ensisf Thanks for example in playground, it might be related to outdated version of flow. I will take a look at your branch and get back to you.

src/components/form-elements/Select.jsx
```diff
@@ -10,6 +10,12 @@ type OptionsPropsType = {
   ...
 };
 
+type GroupedOptionsPropsType = {
+  label: string,
+  options: $ReadOnlyArray<OptionsPropsType>,
+  ...
```
coderitual, line 16:
```suggestion
```

src/components/form-elements/Select.jsx
```diff
@@ -10,6 +10,12 @@ type OptionsPropsType = {
   ...
```
coderitual, line 10:
```suggestion
```

src/components/form-elements/Select.jsx
```diff
@@ -115,11 +128,22 @@ const Select = (props: SelectPropsType) => {
...
     className
   );
-  const optionsElements = options.map(({value, text}) => (
-    <option key={value} value={value}>
-      {text}
-    </option>
-  ));
+
+  const optionsElements = options.map(item => {
+    if (typeof item.label === 'string' && Array.isArray(item.options)) {
```
coderitual, line 133:
```suggestion
    if (item.options) {
```

src/components/form-elements/Select.jsx
coderitual: Hey, there is some problem with understanding `typeof` in the flow version we're using. This probably was introduced later. also checking isArray is redundant in the code and example in playground.

coderitual: Also we can't use `label` in key here. Label is not unique, maybe combining label plus index will be better here.

ensisf: @coderitual Thanks for helping! All changes are applied.

patrycja-b (approved):
⭐️

src/components/form-elements/pages/select-interactive.jsx
```diff
@@ -38,6 +45,10 @@ const Selects = () => {
       <DocsActiveBlock backgroundColor="dark" settings={settings}>
         <Select options={exampleOptions} size="m" color="white" />
       </DocsActiveBlock>
+
+      <DocsActiveBlock backgroundColor="dark" settings={settings}>
+        <Select groupedOptions={exampleGroupedOptions} size="m" color="white" />
```
coderitual, line 50: @ensisf is this example still valid (`groupedOptions`)?

ensisf: Oh, sorry, I've missed here. I will in a moment

ensisf: Done

coderitual (approved):
🥇

PR #2090, opened 2021-08-11
v191.1.0

anna-wro:
* add `sg-cursor-pointer` class

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "191.0.0",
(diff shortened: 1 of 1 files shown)


coderitual:
hej @anna-wro, I think you missed publishing tags 
image

you can do that with `git push --follow-tags`.

you can also set this option globally to always push tags: `git config --global push.followTags true`

anna-wro:
@coderitual I pushed and changed the config, thanks!

coderitual (approved):

PR #2092, opened 2021-08-11
Trim long texts in Label

### LINKED ISSUE #2068: Label and text-overflow: ellipsis
we could consider adding a way to handle longer texts in labels using `text-overflow: ellipsis` option by default, i.e

Screenshot 2021-06-22 at 18 36 30

otherwise, it has to be handled manually (like [here](https://github.com/brainly/brainly-frontend/pull/15835)) and requires overriding style-guide related styles.

clxandstuff:
### Summary

From now label component is single line of text only and will will trim text if it takes more horizontal space than 100% of container. Ellipsis will be added at the end of trimmed text. 
I was thinking about adding title attribute to have access to full text when trimmed but I wasn't sure whether it can conflict with tooltips. Could you add some thoughts?

Since I'm fairly new in the team I've been thinking about other solution but not sure if it fits principles. Let me explain in few points:
1. Add option (via prop) to Text component to be single line only
2. Use single line Text on  Label component, since it's using Text

Screenshot 2021-08-11 at 14 45 58

Fixes #2068

### DIFF
--- a/src/components/labels/Label.jsx
+++ b/src/components/labels/Label.jsx
@@ -257,7 +257,12 @@ const Label = ({
         {children}
       </Text>
       {onClose ? (
-        <button className="sg-label__close-button" onClick={onClose}>
+        <button
+          className="sg-label__close-button"
+          onClick={onClose}
+          title="close"
+          aria-label="close"
+        >
           <Icon type="close" color={closeIconColor} size={16} />
         </button>
       ) : null}
--- a/src/components/labels/Label.jsx
+++ b/src/components/labels/Label.jsx
@@ -248,11 +248,14 @@ const Label = ({
           <Icon type={iconType} color={iconColor} size={16} />
         </div>
       )}
-      <span className="sg-label__text">
-        <Text size="small" weight="bold" color={textColor}>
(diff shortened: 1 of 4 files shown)


matzimowski (approved):
lgtm 👏 

+1 for adding a title, imho it should not collide with tooltip

clxandstuff:
@matzimowski 
Do you think we can use Text to control whether it is single line and trimmed if needed and then use that functionality in Label? I added it to PR description.

matzimowski:
@clxandstuff i like your idea, if there is nobody against, let's do it.

i can see some cases in our web app repo already, where it could be used instead of custom css (just look for `text-overflow: ellipsis;` there).

coderitual (approved):
I like idea with title as well, let's discuss it further. Otherwise it looks good

clxandstuff:
@matzimowski
Few thoughts after chat with @coderitual.

Since most of us like the idea of using Text component to trim texts, we will add to our backlog and make research how we can reuse it later throughout the app. For adding title we stay with current Label props accepting html attributes, so no changes required 😃

PR #2095, opened 2021-08-23
nesting Text, Link, Headline with inherited styles

clxandstuff:
Text, Link, Headline will now accept new prop `inherited` which makes component styles inherited. This will allow us to nest those components without extra effort of duplicating styles on nested text component.

Example: 
To make nested bold text same colour:

```js
// before
<Text color="mustard">
  Some text
  <Text color="mustard" weight="bold">bolded part</Text>
</Text>

// after
<Text color="mustard">
  Some text
  <Text inherited weight="bold">bolded part</Text>
</Text>
```

### DIFF
--- a/src/components/text/Headline.jsx
+++ b/src/components/text/Headline.jsx
@@ -53,25 +53,28 @@ export type HeadlinePropsType = {
   align?: ?HeadlineAlignType,
   className?: ?string,
   extraBold?: ?boolean,
+  inherited?: ?boolean,
   ...
 };
 
 const Headline = ({
   children,
   type = HEADLINE_TYPE.H1,
-  size = HEADLINE_SIZE.MEDIUM,
+  size,
   extraBold,
   transform,
   align,
   color,
   className,
+  inherited = false,
   ...props
 }: HeadlinePropsType) => {
   const Type = type;
   const headlineClass = classNames(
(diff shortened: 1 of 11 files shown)


matzimowski (approved):
really cool idea ❤️ 
we were missing this option 🥇

coderitual:
@clxandstuff great work!

could you please to add appropriate sections to the current docs as well?

bartoszlorek (approved):
🎉 ❤️ 🙌

bartoszlorek:
@clxandstuff Did you check if it works for Text, Link, and Headline? I remember that Headline cannot be inlined, and it will break the inherited solution.

coderitual:
@clxandstuff 

oh @bartoszlorek is right, we need to unset ‘display’ attr when inherit prop is set

clxandstuff:
@bartoszlorek @coderitual 
yep this `display: block` should be also reset

coderitual:
I think I was meant `revert` not unset (unset will pick inherit or initial, and initial is always inline)

clxandstuff:
@coderitual 
That's right. Changed to revert.

matzimowski (approved):
👏

coderitual (approved):

PR #2096, opened 2021-08-23
v191.3.0

clxandstuff:
### Features

- new prop `inherited` has been added to Text, Link, Headline

type: Boolean
default value: false

It makes easier to nest those components and keep parent-child styles in sync.

```js
// before
<Text color="mustard" size="xlarge" weight="bold">
  Some text
  <Text color="mustard" size="xlarge" weight="bold">nested text</Text>
</Text>

// after
<Text color="mustard" size="xlarge" weight="bold">
  Some text
  <Text inherited>nested text</Text>
</Text>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "191.2.0",
+  "version": "191.3.0",
   "description": "Brainly Front-End Style Guide",
   "repository": "https://github.com/brainly/style-guide.git",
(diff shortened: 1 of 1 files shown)


bartoszlorek (approved):

coderitual:
@clxandstuff I think it would be worth to add `type="span"` to examples if you would like to reuse them in release notes.

coderitual (approved):

PR #2119, opened 2021-09-16
add stories from current docs to storybook

clxandstuff:
This is another part of docs migration to Storybook. Goal of this part was to transform current docs examples to SB stories.
This is list of component stories to add in next part:
* Text
* Link
* Headline
* TextBit
* Card

### DIFF
--- a/.storybook/storybook.scss
+++ b/.storybook/storybook.scss
@@ -1,0 +1,25 @@
+@import '../src/sass/config';
+@import '../src/sass/mixins';
+@import '../src/components/buttons/buttons.stories';
+
+.sg-story {
+  display: flex;
+
+  &-variant {
+    margin-bottom: spacing(l);
+
+    &__name {
+      margin-bottom: spacing(s);
+    }
+
+    &-dark-box {
+      background: $graySecondary;
+      padding: spacing(m);
+    }
+
+    &-border-box {
+      border: 1px solid $graySecondary;
+      padding: spacing(m);
+    }
+  }
+}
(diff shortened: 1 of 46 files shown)


.storybook/storybook.scss
```diff
@@ -0,0 +1,25 @@
+@import '../src/sass/config';
+@import '../src/sass/mixins';
+@import '../src/components/buttons/buttons.stories';
+
+.sg-story {
```
coderitual, line 5: do we need this css? layout  and background of the story should be customised on story param level and handled in story block

src/components/accordion/Accordion.stories.jsx
```diff
@@ -117,21 +85,43 @@ export const Controlled = (args: any) => {
...
+  'xxl',
+  'xxxl',
+  'xxxxl',
+  'none',
+];
+
+export const Spacing = (args: any) => (
+  <div>
+    {spacings.map(spacing => (
+      <StoryVariant key={spacing} label={`spacing - ${spacing}`}>
```
coderitual, line 116: I don't like this idea of story variant. It bloats code preview and feels redundant. In the context of accordion it's an overkill. We don't need to present every attribute and possible variants. Accordion in current docs present only 2 variants. Could we discuss it further? I will setup review call for us.

.storybook/storybook.scss
clxandstuff: This styles are for util component to present story variant in consistent way. Not sure if we can do this with SB story blocks

src/components/accordion/Accordion.stories.jsx
clxandstuff: It might be good to present all necessary variants for story for VR story reuse. I agree though, for some cases it can be too much. Ok, let's have a separate discussion for this.

coderitual (approved):
Summary of discussion

For this PR:
- remove unnecessary accordion stories 

Next PR:
- introduce representative story for an avatar
- introduce representative story for button
- Introduce chromatic only stories with accordion as a first example

PR #2120, opened 2021-09-20
docs: storybook stories look update

clxandstuff:
Another part of changes for incremental migration docs to storybook:

- avatar stories update based on figma templates
- storybook utils moved to docs folder
- accordion stories updated
- button stories look update


coderitual (approved):
Table looks great! Let's go with it. As for "sticky row", we don't need it now. Just be sure to create jira ticket so we can get back to it in the future.

PR #2122, opened 2021-09-21
add codebuild buildspec

clxandstuff:
We need buildspec for Storybook publish automation. This file describes how AWS Codebuild will be building our project Storybook before publishing

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -1,0 +1,18 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - yarn
+  build:
+    commands:
+      - yarn build-storybook
+  post_build:
+    commands:
+      - GIT_BRANCH=${CODEBUILD_WEBHOOK_HEAD_REF:-$CODEBUILD_SOURCE_VERSION}
+      - export GIT_BRANCH=$(echo $GIT_BRANCH | sed "s|branch/||" | sed "s|/|-|g" | sed "s|refs-heads-||")
+
+artifacts:
+  name: '$GIT_BRANCH'
+  files:
+    - '.storybook/public/**/*/'


buildspec.yml
```diff
@@ -0,0 +1,18 @@
...
+      - yarn build-storybook
+  post_build:
+    commands:
+      - GIT_BRANCH=${CODEBUILD_WEBHOOK_HEAD_REF:-$CODEBUILD_SOURCE_VERSION}
+      - export GIT_BRANCH=$(echo $GIT_BRANCH | sed "s|branch/||" | sed "s|/|-|g" | sed "s|refs-heads-||")
+
+artifacts:
+  name: '$GIT_BRANCH'
+  files:
+    - '.storybook/public/**/*/'
```
coderitual, line 18: I think `.storybook-static/**/*` is the correct artifact folder

coderitual: or this is only to expose for build?

clxandstuff: Nice catch 😄 Thanks

buildspec.yml
```diff
@@ -0,0 +1,18 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - yarn
+  build:
+    commands:
+      - yarn build-storybook
```
coderitual, line 9: we need additional step to build icons -> `yarn build`. Alternatively we can run this command from `yarn build-storybook`. Building icons for storybook is something we need to fix anyway

clxandstuff: No worries, I will add `yarn build` before that step. Thanks for catch 😄

clxandstuff: Done

buildspec.yml
clxandstuff: Done

PR #2124, opened 2021-09-22
update storybook to stable v6

clxandstuff:


package.json
```diff
@@ -38,14 +38,14 @@
...
     "@microsoft/api-extractor": "^7.13.0",
-    "@storybook/addon-a11y": "@next",
-    "@storybook/addon-actions": "@next",
-    "@storybook/addon-docs": "@next",
-    "@storybook/addon-essentials": "@next",
-    "@storybook/addon-links": "@next",
-    "@storybook/addons": "@next",
-    "@storybook/react": "@next",
-    "@storybook/theming": "@next",
+    "@storybook/addon-a11y": "6",
```
coderitual, line 41: @clxandstuff  hmm I am wondering why yarn specified it that way instead of `^6.3.8`. Do you know the reason?

clxandstuff: I think this is my choice :) I will install them without specifying version. This should update package.json  to proper version

clxandstuff: Done :)

PR #2125, opened 2021-09-24
improve typings

clxandstuff:
- icon typings are safer to use, have safer names (not mangled)
- all html attributes can be passed as props

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -101,7 +101,10 @@ export type {ListItemIconPropsType} from './components/list/ListItemIcon';
 export {default as MenuList} from './components/list/MenuList';
 export type {MenuListPropsType} from './components/list/MenuList';
 export {default as Logo} from './components/logo/Logo';
-export type {LogoPropsType} from './components/logo/Logo';
+export type {
+  LogoPropsType,
+  LogoTypeType as LogoType,
+} from './components/logo/Logo';
 export {default as MathSymbol} from './components/math-symbols/MathSymbol';
 export type {
   MathSymbolPropsType,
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -73,7 +73,10 @@ export type {IconAsButtonPropsType} from './components/icon-as-button/IconAsButt
 export {default as Icon} from './components/icons/Icon';
 export type {IconPropsType} from './components/icons/Icon';
 export {default as MobileIcon} from './components/mobile-icons/MobileIcon';
-export type {MobileIconPropsType} from './components/mobile-icons/MobileIcon';
+export type {
+  MobileIconPropsType,
+  IconTypeType as MobileIconTypeType,
```
coderitual, line 78: How about changing naming to something like: `MobileIconTypeType`?

clxandstuff: No problem for me, but keep in mind that all types should have Type postfix and icon type apparently ends the same 😄 It will pass linting but can be less consistent with other type names

coderitual: True, but I would rather remove Type suffix everywhere. It is there only because of flow. Currently we export only 2 kind of types. PropsType for each component and now second one for Icons names. All the other types are access by index and we dont really care about their names.

clxandstuff: Yes, I feel the same :) I changed both exported types to `MobileIconType` and `SubjectIconType`

coderitual:
@clxandstuff how about `IconType` from `/components/icons/Icon`?

src/index.js
```diff
@@ -97,7 +103,10 @@ export type {MenuListPropsType} from './components/list/MenuList';
 export {default as Logo} from './components/logo/Logo';
 export type {LogoPropsType} from './components/logo/Logo';
```
coderitual, line 104: @clxandstuff I believe we talked that changing type in logo component would be good

coderitual (approved):

src/index.js
clxandstuff: Sorry, I misunderstood that it just 'would be good' but we will cover it later😄

clxandstuff: Logo also exported now :)

coderitual: No worries, I wasn't sure either :)

PR #2126, opened 2021-09-27
v193.2.0

clxandstuff:


coderitual:
@clxandstuff i think this should be bumped as a breaking change. we change global type names, even if we know application doesn't use it.

clxandstuff:
Ah crap, you're right. Closing this one

PR #2128, opened 2021-09-29
update flash message stories

clxandstuff:
update stories based on figma templates
Screenshot 2021-09-29 at 11 35 13

### DIFF
--- a/src/components/flash-messages/FlashMessage.stories.jsx
+++ b/src/components/flash-messages/FlashMessage.stories.jsx
@@ -33,7 +33,7 @@ export const Types = args => (
             {type}
           </Headline>
         </td>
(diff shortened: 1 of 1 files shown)


domidomi:
Do we want to reflect the component the same as it is in the handoff file? If so, then I think each flash message should be in a wider container so it can fill it, as at the moment it looks a bit squished. Figma preview:
Screenshot 2021-09-29 at 16 08 08

clxandstuff:
Well, why not? :)

clxandstuff:
Screenshot 2021-09-29 at 16 27 42

domidomi (approved):

coderitual (approved):

PR #2130, opened 2021-09-29
align Label stories to figma templates

clxandstuff:
'types' story
part of it on screenshot below

Screenshot 2021-09-30 at 13 55 59

'trimmed text' story
Screenshot 2021-09-30 at 13 57 03

### DIFF
--- a/src/components/labels/Label.stories.jsx
+++ b/src/components/labels/Label.stories.jsx
@@ -1,6 +1,37 @@
 import * as React from 'react';
-import Label, {ICON_TYPE, LABEL_COLORS_SET, LABEL_TYPE} from './Label';
-import Flex from '../flex/Flex';
+import Label, {
+  ICON_TYPE,
+  LABEL_COLORS_SET,
+  LABEL_TYPE,
+  COLORS_DEFAULT_MAP,
+  COLORS_SOLID_MAP,
+} from './Label';
+import {StoryVariantTable} from '../../_docs/utils';
+import Headline from '../text/Headline';
+import hex from '../colors/hex';
+
+const onCloseMock = () => null;
+const onClickMock = () => null;
+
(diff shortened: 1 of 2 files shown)


src/components/labels/Label.stories.jsx
```diff
@@ -1,16 +1,46 @@
...
+  LABEL_COLORS_SET,
+  LABEL_TYPE,
+  COLORS_DEFAULT_MAP,
+  COLORS_SOLID_MAP,
+} from './Label';
+import {StoryVariantTable} from '../../_docs/utils';
+import Headline from '../text/Headline';
+import hex from '../colors/hex';
+
+const onCloseMock = () => null;
```
coderitual, line 13: nitpick: maybe just:

`function noop() {}` instead of those 2 mock functions?

src/components/labels/Label.stories.jsx
```diff
@@ -804,14 +428,10 @@ export const LongText = args => {
...
-const onCloseMock = () => {};
-
-LongText.args = {
-  ...Default.args,
+TrimmedText.args = {
   title: 'Long long long text',
   children: 'Very very long label',
   onClose: {control: onCloseMock},
+  color: LABEL_COLORS_SET.BLUE,
+  onClick: onClickMock,
```
coderitual, line 436: why do we need onClick?

src/components/labels/Label.stories.jsx
clxandstuff: Good idea! 👍

src/components/labels/Label.stories.jsx
clxandstuff: Actually it's unnecessary, adde by mistake. Thanks for catch

coderitual (approved):

PR #2134, opened 2021-10-04
Fix info icon

coderitual:
Fixing `info` icon by removing fill property so icon can get proper color.

Before:
image

After:
image


matzimowski (approved):
sorry, my bad, i didn't change it. thanks @coderitual

clxandstuff (approved):

PR #2143, opened 2021-10-05
reduce number of vulnerable dependencies

clxandstuff:
Dependencies were updated with `yarn upgrade`. After update of eslint import plugin, I had to add this:

```
"settings": {
    "import/resolver": {
      "node": {
        "extensions": [".js", ".jsx"]
      }
    }
  }
```
to `import/extensions` rule to work properly.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,17 +5,24 @@
   "parser": "babel-eslint",
   "extends": ["brainly-react"],
   "rules": {
-    "max-len": ["error", { "code": 120, "ignoreComments": true }],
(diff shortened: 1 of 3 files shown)


coderitual:
Hey @clxandstuff , could we please sync about this? I think that this kind of update might be potentially harmfull. 

Yarn.lock is being updated extensively but it is not cleaned up really. Also we see breaking changes happend but our package.json is not touched at all.

let’s think about strategy of updating deps. I think we may need more granular approach upgrading particular package one by one with possible help of —latest flag and upgrade interactive mode.

PR #2176, opened 2021-11-03
Export Dialog component

radekwarisch:
I have noticed this component is not exported yet. We would love to use it. 
Let me know whether it is not production-ready.


coderitual (approved):

coderitual:
@radekwarisch it's not being used on prod yet but it's production ready, probably first usage will be language selector.

Please keep in mind that dialog will be placed in the dom exactly where it's used, if you need to use portals or implement advanced displaying logic (e.g some queues, state managers etc) it should be entirely on app level.

Also, some current uses of the modal component existing in the application would need to be redesigned to fit the new dialog, but this is a natural process of adopting a new component and will happen in product teams when they need to use the new look for their feature.

radekwarisch:
@coderitual sure, we intend not to solve state mgmt of modals yet. We will keep you in the loop once migrations with dialog will be implemented.

PR #2177, opened 2021-11-03
Accordion a11y improvements

katarzynatobis:
- role `heading` and `headingLevel` prop (default value: `2`) added to `AccordionItem` 
- `aria-disabled` set `true` for expanded item in controlled accordion

### DIFF
--- a/src/components/accordion/Accordion.jsx
+++ b/src/components/accordion/Accordion.jsx
@@ -67,7 +67,6 @@ type ContextType = {
   dispatch: (action: ActionType) => void,
   reduceMotion: boolean,
   onItemSelect: (id: string, value: boolean) => void,
-  isControlled: boolean,
   ...
 };
 
@@ -280,7 +279,6 @@ const Accordion = ({
       dispatch,
       reduceMotion: hasReduceMotion,
       onItemSelect,
-      isControlled,
     }),
     [
       hasReduceMotion,
@@ -288,7 +286,6 @@ const Accordion = ({
       onItemSelect,
       state.focusedElementId,
       state.expanded,
-      isControlled,
     ]
   );
 
--- a/src/components/accordion/Accordion.jsx
+++ b/src/components/accordion/Accordion.jsx
@@ -67,6 +67,7 @@ type ContextType = {
   dispatch: (action: ActionType) => void,
   reduceMotion: boolean,
   onItemSelect: (id: string, value: boolean) => void,
+  isControlled: boolean,
   ...
 };
 
@@ -279,6 +280,7 @@ const Accordion = ({
       dispatch,
       reduceMotion: hasReduceMotion,
       onItemSelect,
+      isControlled,
     }),
     [
       hasReduceMotion,
@@ -286,6 +288,7 @@ const Accordion = ({
       onItemSelect,
       state.focusedElementId,
       state.expanded,
+      isControlled,
     ]
   );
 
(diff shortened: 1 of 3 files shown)


src/components/accordion/Accordion.jsx
```diff
@@ -67,6 +67,7 @@ type ContextType = {
   dispatch: (action: ActionType) => void,
   reduceMotion: boolean,
   onItemSelect: (id: string, value: boolean) => void,
+  isControlled: boolean,
```
clxandstuff, line 70: I think `isControlled` is unnecessary here in context. Maybe we can pass list of disabled accordion items controlled by accordion? Item could be marked as disabled when for instance `allowMultiple` is false and item is already expanded?

clxandstuff (changes requested):

src/components/accordion/AccordionItem.jsx
```diff
@@ -53,10 +55,12 @@ const AccordionItem = ({
     dispatch,
     reduceMotion,
     onItemSelect,
+    isControlled,
   } = useContext(AccordionContext);
   const [isHovered, setIsHovered] = useState(false);
 
   const isCollapsed = !expanded.includes(id);
+  const isDisabled = !isCollapsed && isControlled;
```
clxandstuff, line 63: I found this does not work when we're in `allowMultiple` mode. I think this value can be controlled by accordion and passed as prop. Take a look at my other comment on `src/components/accordion/Accordion.jsx`

clxandstuff (changes requested):

src/components/accordion/Accordion.jsx
katarzynatobis: `allowMultiple` is not working in controlled Accordion, in that case only one accordion item can be expanded and then must be disabled (it can be closed only by expanding some other item).

clxandstuff: You are right, controlled Accordion doesn't care about `allowMultiple`. As we discussed it, If someone wants to control Accordion, he should also control `aria-disable` attributes of accordion items.

src/components/accordion/AccordionItem.jsx
clxandstuff: We discussed with @katarzynatobis that we don't need it anymore. Component which controls Accordion also has to control `aria-disable` attributes

coderitual: I don't fully agree. 

Considering the rule:
> If the accordion panel associated with an accordion header is visible, and if the accordion does not permit the panel to be collapsed, the header button element has aria-disabled set to true.

and how our version of accordion is working we need to implement is as follows:
-  When `Accordion` is within **uncontrolled** mode, aria-disabled in `AccordionItem` should be always `false`. `AllowMultiple` flag has nothing to do here because it's different property from keeping at least one item always opened. We don't provide that property on purpose because in our implementation we decided that each item can be always collapsed by default.
- When `Accordion` is in **controlled** mode, developers can implement mechanism to keep at leas one element opened and we need to provide him a way to pass aria-disabled to the item they decide to keep in that state.

Let's discuss tomorrow

clxandstuff: Thats exactly same conclusion we had with Kasia. Let's meet tomorrow :)

katarzynatobis: commit with disabled state is reverted now :)

coderitual (approved):

clxandstuff (approved):

PR #2178, opened 2021-11-03
v198.3.0

katarzynatobis:
## Features

- Improved `Dialog` accessibility:
	 - focus dialog when there are no focusable descendants
	 - `aria-*` attributes for `Dialog` and `DialogCloseButton`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "198.2.0",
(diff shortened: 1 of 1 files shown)


clxandstuff:
Hey @katarzynatobis 
This release should be v198.3.0 as long as we're adding functionality. Let me know if you need any help changing this to 198.3.0 (FYI: 198.2.1 tag is already pushed)

PR #2184, opened 2021-11-12
Add additional zIndex property to Dialog

Malinicz:
[Features]

- add additional zIndex prop in order to avoid problems with stacking:

https://user-images.githubusercontent.com/14280367/141443102-c5062973-f8f3-4764-bb32-72c311a87680.mov

### DIFF
--- a/src/components/dialog/Dialog.jsx
+++ b/src/components/dialog/Dialog.jsx
@@ -25,6 +25,7 @@ export type DialogPropsType = $ReadOnly<{
    * Fires on user actions like clicking outside
    * the Dialog or the Escape key.
    */
+  zIndex?: number,
   onDismiss?: () => void,
   onEntryTransitionEnd?: () => void,
   onExitTransitionEnd?: () => void,
@@ -56,6 +57,7 @@ function BaseDialog({
   labelledBy,
   label,
   describedBy,
+  zIndex = 'auto',
   onDismiss,
   onEntryTransitionEnd,
   onExitTransitionEnd,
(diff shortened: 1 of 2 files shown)


anna-wro (approved):

coderitual:
@Malinicz I think that setting z-index shouldn't be part of the current implementation of dialog. Dialog placement is entirely up to the app. Probably in new part of the app portal might be used. 

My suggestion for now is to create layer in app which will handle z-index. The other option we may consider is to expose assigning classnames for overlay/container but maybe we can avoid that, wdyt?

Malinicz:
> @Malinicz I think that setting z-index shouldn't be part of the current implementation of dialog. Dialog placement is entirely up to the app. Probably in new part of the app portals might be used.
> 
> My suggestion for now is to create layer in app which will handle z-index. The other option we may consider is to expose assigning classname for overlay

Ok, I spoke with Bery about it and he was ok with this prop, but you are probably right. We will create reusable wrapper for this dialog in the app for now. @coderitual closing this PR

coderitual (approved):

src/components/dialog/Dialog.jsx
```diff
@@ -25,6 +25,7 @@ export type DialogPropsType = $ReadOnly<{
    * Fires on user actions like clicking outside
    * the Dialog or the Escape key.
    */
+  zIndex?: number,
```
clxandstuff, line 28: Could you move this to next line? Comment is misaligned

clxandstuff (changes requested):

clxandstuff (approved):

PR #2185, opened 2021-11-12
chromatic init

clxandstuff:


scripts/chromatic-stories.js
```diff
@@ -0,0 +1,34 @@
+const fs = require('fs');
```
coderitual, line 1: Can we change the name of this script to something like `genetate-chromatic-stories`?

.storybook/main.js
```diff
@@ -7,8 +7,14 @@ const SOURCE_DIR = path.join(__dirname, '../src');
 const SOURCE_DOCS_DIR = path.join(SOURCE_DIR, '_docs');
 const SOURCE_COMPONENTS_DIR = path.join(SOURCE_DIR, 'components');
 
+console.log(process.env);
+
 module.exports = {
-  stories: ['../src/**/*.stories.@(jsx|mdx)'],
+  stories: [
+    process.env.CHROMATIC === 'true'
+      ? '../src/**/*.chromatic-stories.@(jsx|mdx)'
```
coderitual, line 15: I am thinking about changing the name of chromatic stories from `chromatic-stories` to `chromatic.stories.@(jsx|mdx)`

The only challenge will be to filter out `chromatic.stories` from normal storybook run. Fortunately stories property in main.js config also takes async function so we can dynamically collect all needed stories and filter them out in js function.

reference: https://storybook.js.org/docs/react/configure/overview#configure-story-loading

yarn.lock
```diff
@@ -7967,7 +7967,7 @@ fast-glob@^2.2.6:
     merge2 "^1.2.3"
     micromatch "^3.1.10"
 
-fast-glob@^3.1.1:
+fast-glob@^3.1.1, fast-glob@^3.2.7:
```
coderitual, line 7970: I already use glob in other scripts, is fast-glob any better?

scripts/chromatic-stories.js
```diff
@@ -0,0 +1,34 @@
...
+const path = require('path');
+const fg = require('fast-glob');
+
+const storiesPaths = fg.sync('src/**/*.stories.jsx');
+
+storiesPaths.forEach(storiesPath => {
+  let componentName = storiesPath.split('.')[0].split('/');
```
coderitual, line 8: I suggest using `path.parse()`

scripts/chromatic-stories.js
```diff
@@ -0,0 +1,34 @@
...
+    .join('/');
+
+  fs.writeFileSync(
+    path.resolve(
+      __dirname,
+      `../${chromaticStoriesPath}`,
+      `${componentName}.chromatic-stories.jsx`
+    ),
+    `import * as ${componentName} from './${componentName}.stories.jsx';
+import {mergeStories} from '${new Array(subdirsCount)
```
coderitual, line 25: let's use `path.relative(from, to)` here

yarn.lock
clxandstuff: I can't see glob as project dev dependency. You could use it because other modules install it. I think it will be safer to add it to our package json, and then I can use it in this script also. What do you think?

coderitual: Oh defintely, let’s add glob explicitly then

coderitual (approved):

scripts/chromatic-stories.js
clxandstuff: Good idea. I added section in docs and renamed script.

.storybook/main.js
clxandstuff: Updated. Now we can use `.chromatic.stories` and `.stories` extensions

scripts/chromatic-stories.js
clxandstuff: Done, thanks for tip.

scripts/chromatic-stories.js
clxandstuff: Nice, thanks for tip!

CONTRIBUTING.md
```diff
@@ -59,32 +61,49 @@ All components are **responsive** by default, so there is no need for modifiers
...
+
+There is also Chromatic visual regression tool which uses storybook stories to run tests. It is integrated with Github and being run on pull requests.
+
+If you already have storybook stories, you can use them also for Chromatic. Run script:
+
+```
+$ node ./scripts/generate-chromatic-stories.js
+```
+
+which looks for `{component_name}.chromatic.stories.jsx` files and generates `{component_name}.chromatic.stories.jsx` for each of them(if not exists). Created files have single exported story(Default). You can extend it with your stories if needed.
```
coderitual, line 93: @clxandstuff did you mean `{component_name}.stories.jsx` in first part?

I would also rephrase: `You can extend it with your stories if needed.` with `You can extend it with chromatic specific stories if needed.`

package.json
```diff
@@ -138,7 +139,8 @@
     "prepublishOnly": "yarn package-prepare",
     "storybook": "start-storybook -p 6006 -s ./.storybook/public",
     "storybook-docs": "start-storybook --docs -p 6006 -s ./.storybook/public",
-    "build-storybook": "build-storybook -c .storybook -o .storybook-static -s ./.storybook/public"
+    "build-storybook": "build-storybook -c .storybook -o .storybook-static -s ./.storybook/public",
+    "chromatic": "CHROMATIC=true npx chromatic --project-token=940abc073970 --exit-zero-on-changes"
```
coderitual, line 143: Shouldn't we store token in secrets like described [here](https://www.chromatic.com/docs/github-actions)?

clxandstuff: Yes, even more in fact style guide repo is public. I will move it to secrets

CONTRIBUTING.md
clxandstuff: Done

package.json
clxandstuff: Done

Issue #2035, opened 2021-03-22
Implement accessibility for Icon component


ElaLukaszewicz:
Implement accessibility for Icon component.
Suggestion of solution to this issue:
- remove title from svg's (we would like to be able to pass translated aria-labels for icons)
- add ```role=img```
- remove ```div``` container if possible to pass additional props to svg itself

coderitual:
https://github.com/brainly/style-guide/issues/1651

PR #2188, opened 2021-11-15
Improve `Icon` accessibility

katarzynatobis:
- added accessible `title` & `description` props for `Icon`
- added `role` to `svg` in `Icon` (default: `"img"`)
- adjusted colors for high contrast mode (only Windows):
normal colors: 
dark1
high contrast mode before change:
dark2
high contrast mode after change:
dark3

closes: #2035


src/components/icons/Icon.jsx
```diff
@@ -488,6 +488,18 @@ export type IconPropsType =
...
+       */
+      title?: string,
+      /**
+       * An accessible, long-text description
+       */
+      description?: string,
+      /**
+       * Provides semantic meaning to content, defaults to "img"
+       */
+      role?: string,
```
coderitual, line 502: @katarzynatobis can it be different than img?

domidomi: If title and description are optional, then maybe we should set default role as 'presentation'? Otherwise accessibility checks will always throw warnings as title and description may not always always provided

coderitual: My understanding is:
- if we allow only `presentation` and `img` in role, let's reflect that in type of role - I think that default to `img` is a good way.
- what is the benefit of `presentation` role over `aria-hidden` in the context of our icon component? This would be alternate way allowing aria-hidden to be passed down to the component and having role always set to `img` without `role` prop being exposed

src/components/icons/_icons.scss
```diff
@@ -9,13 +9,18 @@ $iconSizes: 16, 24, 32, 40, 56, 80, 104;
   .sg-icon {
     @include iconColor();
 
-    fill: $white;
+    color: $white;
+    fill: currentColor;
```
coderitual, line 13: Nice change!

src/components/icons/Icon.jsx
coderitual: Correct me if I am wrong but role presentation might not be the thing we want to have in here. It will just remove the native semantic meaning (which is graphical-content in our case) but the screen reader will still see svg nodes. I think that role should be always set to `img` and we should allow passing aria-hidden. wdyt?

clxandstuff: I have same thoughts after reading those materials @coderitual linked. It might be unnecessary to expose role as prop, since we know icon acts like an img. Aria hidden should do the job .

clxandstuff (changes requested):

src/components/icons/Icon.jsx
katarzynatobis: I wasn't clear enough, sorry. `presentation` was only an example - based on w3, svg can [have any role](https://www.w3.org/TR/html-aria/#el-svg). But I agree with you - in that case, only `img` makes sense.

`aria-hidden` doesn't have to be passed to `svg`, it can be used on wrapper element. What's more, on wrapper it works for custom icon as well, so the icon api is more consistent and that's why I decided to add `aria-label`.

src/components/accordion/Accordion.spec.jsx
```diff
@@ -178,9 +178,9 @@ describe('<Accordion>', () => {
       </Accordion>
     );
 
-    expect(accordion.find('[aria-labelledby]').hostNodes()).toHaveLength(
-      accordionIds.length
-    );
+    expect(
```
coderitual, line 181: I would keep changes only related to the PR and prepare separate pr to improve other component

katarzynatobis: Adding `aria-labelledby` in `Icon` caused this test to fail, without this change "Continuos Integration / Build and Test (push)" step fails.

coderitual (approved):

clxandstuff (approved):

coderitual:
hey @katarzynatobis , is there anything yet in this PR you would like to discuss?

PR #2194, opened 2021-11-18
setup jest-axe tests for a11y

katarzynatobis:


src/axe/testA11y.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @example
+ * ```jsx
+ *  it('should have no a11y violations', async () => {
+ *    await testA11Y(<MyComponent />, options);
+ *  });
+ * ```
+ */
+
+export default async function testA11y(node, axeOptions) {
```
coderitual, line 15: I am not convinced to this helper. It automatically runs `toHaveNoViolations` and has conditional mount which might result in leaky abstraction (kind of abstraction you need to know internal implementation).

We may be also testing multistep behaviour in the context of axe. e.g:
- render component
- make an user action
- **check** if there are no violations (after user action).

In that context this function won't give a lot of value but might cause test to be less readable.

I know, that fluent ui uses this approach bet let's discuss it further.

coderitual (approved):
💯

clxandstuff (approved):

PR #2195, opened 2021-11-19
publish storybook master only on version change

clxandstuff:
Publish on master will not upload new files until new version in package json.


coderitual (approved):

scripts/test-s3.js
```diff
@@ -0,0 +1,19 @@
...
+  },
+});
+
+client.s3.headObject(
+  {
+    Bucket: 'style-guide.brainly.com',
+    Key: `index.html`,
+  },
+  function(notExist) {
+    console.log(notExist);
```
coderitual, line 17: maybe more descriptive?

scripts/build-storybook.js
```diff
@@ -0,0 +1,23 @@
...
+  },
+});
+
+client.s3.getObject(
+  {
+    Bucket: 'style-guide.brainly.com',
+    Key: `.sg-version`,
+  },
+  function(err, data) {
+    if (!data || (data && data.toString('utf-8') !== version)) {
```
coderitual, line 18: maybe some console info if it's not met?

clxandstuff: Good idea, thanks

scripts/test-s3.js
clxandstuff: This is actually test file. I will remove it :) Thanks for catching it

PR #2199, opened 2021-11-22
Flex responsive props

clxandstuff:
### Features
This feature introduces responsive props for Flex component. It enables passing multiple of values for prop depending on screen width.

**Object notation**

Breakpoints keys: 'sm', 'md', 'lg', 'xl' (all optional)

`<Flex
      marginLeft={{
        sm: 's',
        lg: 'l'
      }}>content</Flex>`

result HTML:
```
<div class="sg-flex sg-flex--margin-left-s lg:sg-flex--margin-left-l">
  content
</div>
```

**Array notation**

Breakpoints items (use in the same order): 'sm', 'md', 'lg', 'xl' (all optional, can be null or undefined)

`<Flex marginLeft={['s', 'm', null, 'l']}>content</Flex>`

result HTML:
```
<div class="sg-flex sg-flex--margin-left-s md:sg-flex--margin-left-m xl:sg-flex--margin-left-l">
  content
</div>
```

List of supported responsive props:
- fullWidth
- fullHeight
- noShrink
- inlineFlex
- alignItems
- alignContent
- justifyContent
- wrap
- wrapReverse
- alignSelf
- direction
- margin
- marginTop
- marginBottom
- marginLeft
- marginRight

#### CSS size:
**without responsive props:**
original size: 119691 / gzipped size: 15822
**with responsive props:**
original size: 137531 / gzipped size: 17168

### DIFF
--- a/src/components/flex/Flex.spec.jsx
+++ b/src/components/flex/Flex.spec.jsx
@@ -103,7 +103,7 @@ describe('<Flex>', () => {
   });
 
   it('renders responsive props passed as object', () => {
-    const component = shallow(
+    const componentWithAllBreakpoints = shallow(
       <Flex
         fullWidth={{
           sm: true,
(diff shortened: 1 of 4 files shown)


src/components/flex/Flex.spec.jsx
```diff
@@ -101,4 +101,75 @@ describe('<Flex>', () => {
...
+    expect(
+      component.findWhere(
+        node =>
+          node.prop('className') ===
+          'sg-flex sg-flex--full-width md:sg-flex--full-width lg:sg-flex--auto-width xl:sg-flex--auto-width'
+      )
+    ).toHaveLength(1);
+  });
+
+  it('renders responsive class names with prop passed as array', () => {
```
coderitual, line 144: is this duplicate from above?

src/components/flex/Flex.spec.jsx
```diff
@@ -101,4 +101,75 @@ describe('<Flex>', () => {
...
+    expect(
+      component.findWhere(
+        node =>
+          node.prop('className') ===
+          'sg-flex sg-flex--full-width md:sg-flex--full-width lg:sg-flex--auto-width xl:sg-flex--auto-width'
+      )
+    ).toHaveLength(1);
+  });
+
+  it('renders responsive class names when passed null or undefined as item', () => {
```
coderitual, line 160: I think that only null was tested here. I would test 3 situations:

- `<Flex fullWidth={[true, true, null, false]}>`
- `<Flex fullWidth={[true, true, undefined, false]}>`
- `<Flex fullWidth={[,,,false]}>` - this is almost the same as above but it might be a good example of use

domidomi: I would also test cases of `<Flex fullWidth={[false,false,true,true]}>` and `<Flex fullWidth={[,,,,]}>`

clxandstuff: @domidomi @coderitual 
Wouldn't be sufficient to this one case but with 3 approaches?

- `<Flex fullWidth={[true, true, null, false]}>`
- `<Flex fullWidth={[true, true, undefined, false]}>`
- `<Flex fullWidth={[true, true, , false]}>`

I have feeling that other cases can be redundant.

clxandstuff:
```
it('renders responsive class names when passed null or undefined as item, or sparse array', () => {
    const componentWithNull = shallow(
      <Flex fullWidth={[true, true, null, false]}>
        <div>test</div>
      </Flex>
    );

    const componentWithUndefined = shallow(
      <Flex fullWidth={[true, true, undefined, false]}>
        <div>test</div>
      </Flex>
    );

    const componentWithSparseArray = shallow(
      // eslint-disable-next-line no-sparse-arrays
      <Flex fullWidth={[true, true, , false]}>
        <div>test</div>
      </Flex>
    );

    [
      componentWithNull,
      componentWithUndefined,
      componentWithSparseArray,
    ].forEach(component => {
      expect(
        component.findWhere(
          node =>
            node.prop('className') ===
            'sg-flex sg-flex--full-width md:sg-flex--full-width xl:sg-flex--auto-width'
        )
      ).toHaveLength(1);
    });
  });
```

src/components/flex/Flex.spec.jsx
clxandstuff: Yes, thats duplicated. I will remove it.

src/components/flex/Flex.spec.jsx
coderitual: I have a feeling that testing different combinations of they ways we pass responsive props is crucial for testing api. not only null vs undefined vs omitting but also different placement of those values in array e.g all 4 placements in array but also what if I pass more than 4 and only 1 value.

clxandstuff: Ok , I think below list includes all cases for array and object notation. Let me know if we need to cover something more.

Object notation:
- with all breakpoints
- with some breakpoints
- with no breakpoints
- with invalid breakpoints

Array notation:
- with all breakpoints
- with null breakpoints
- with undefined breakpoints
- with skipped breakpoints (sparse array)
- with no breakpoints
- with too many breakpoints

coderitual: cool, how about using `test.each` for all those cases above (I believe we havent used it yet but it might be good use case)?

src/components/flex/Flex.spec.jsx
clxandstuff: Done

coderitual (approved):

katarzynatobis (approved):

PR #2202, opened 2021-11-24
Apply new colors system

domidomi:
Introduced new color variables in scss next to the current colors.
Updated styleguide and storybook documentations with the new color system.

Screenshot 2021-11-24 at 14 58 46

Screenshot 2021-11-24 at 14 58 55


src/sass/_config.scss
```diff
@@ -46,6 +46,84 @@ $transparent: rgba(255, 255, 255, 0);
 $buttonPrimaryFbColor: #3c5b96;
 $buttonPrimaryFbHoverColor: #4367a9;
 
+// New colors system ⬇️
+$blue-70: #002238;
```
coderitual, line 50: @domidomi this fake comment to mitigate "no usage" issue will be needed here, right?

domidomi: @coderitual Do you mean I should move it right below old color system?

coderitual: @domidomi oh sorry, I thought it was that error we have been takimg about but apparently there are new scams linter errrors:
- small hex values
- Camel case in variables - can we change this settting to allow the new name with hyphen?

domidomi: @coderitual it seems we can only change the rule `variable-name-format` to some regex expression, as it doesn't support multiple values. In that case I have to do changes to `sass-lint-config-brainly` package

coderitual: @domidomi let's override those rules in.sass-lint for now. Frontend app uses different linter and we need to check and eventually override rules in the app during sg upgrade.

coderitual (approved):

PR #2208, opened 2021-11-26
Icon new color system

domidomi:


src/components/icons/_icons-mixin.scss
```diff
@@ -3,51 +3,43 @@
     fill: inherit;
   }
 
-  &--light { //default
-    fill: $white;
+  &--icon-black {
```
clxandstuff, line 6: Just wondering if we could skip 'icon' part in class names when we know this is icon component? On the other hand it's can be easier to match it to actual color name in figma.

clxandstuff (approved):

src/components/icon-as-button/IconAsButton.spec.jsx
```diff
@@ -15,10 +15,10 @@ test('render', () => {
 
 test('colors', () => {
   const type = TYPE.ANSWER;
-  const color = ICON_COLOR.DARK;
+  const color = ICON_COLOR['icon-black'];
   const iconAsButton = shallow(<IconAsButton type={type} color={color} />);
 
-  expect(iconAsButton.hasClass('sg-icon-as-button--dark')).toEqual(true);
+  expect(iconAsButton.hasClass('sg-icon-as-button--icon-black')).toEqual(true);
```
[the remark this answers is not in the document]
domidomi: Sure, we can definitely go through all tests in the future and remove such tests, but I think it would be "too much" for this PR

src/components/icons/_icons-mixin.scss
domidomi: It would change our classnames construction convention https://github.com/brainly/style-guide/blob/214bb15c4251249305460a1f707de1169a5aec41/src/components/icons/Icon.jsx#L539

src/components/icon-as-button/IconAsButton.spec.jsx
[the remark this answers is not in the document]
clxandstuff: Yep, this is not blocker for me. Everything looks fine :)

src/components/icons/_icons-mixin.scss
clxandstuff: Yes, then function should be adjusted to produce shorted class names. Reason for that was just some word duplication in class name. Now it's `sg-icon--icon-black`, but could be shorter `sg-icon--black`. Nevermind, I think it's not worth changing 😄

clxandstuff:
@domidomi 
Please check Chromatic UI tests, there are some changes, but I thought there shouldn't be any visual changes so far :)

domidomi:
> @domidomi Please check Chromatic UI tests, there are some changes, but I thought there shouldn't be any visual changes so far :)

We are actually introducing visual changes to Icon, as we want to fix accessibility issues. That's why icons become darker now. @clxandstuff

clxandstuff:
@domidomi 
Allright, thanks. I can see on latest Chromatic build only important changes left. Could you check and accept build on Chromatic and add me as a reviewer? This may first good use case to try Chromatic. You can find some instructions on Confluence page I'm working on at the moment.
https://brainly.atlassian.net/wiki/spaces/DesignSystem/pages/939786571/Chromatic+in+the+Style+Guide

PR #2209, opened 2021-11-29
A11y improvements icon

katarzynatobis:
- added accessible `title` & `description` props for `Icon`
- added `role` to `svg` in `Icon` (default: `"img"`)
- adjusted colors for high contrast mode (only Windows):
normal colors: 
dark1
high contrast mode before change:
dark2
high contrast mode after change:
dark3

closes: #2035

### DIFF
--- a/src/components/buttons/_buttons.scss
+++ b/src/components/buttons/_buttons.scss
@@ -11,7 +11,7 @@ $largeButtonSize: componentSize(l);
   }
 }
 
-@mixin iconColor($color) {
+@mixin buttonIconColor($color) {
   & .sg-button__icon {
     color: $color;
   }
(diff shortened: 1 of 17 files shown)


coderitual (approved):

coderitual:
@katarzynatobis we have some changes in chromatic, could you please take a look?

clxandstuff:
I added some comments. The very first outcome from Chromatic integration :)

coderitual:
@clxandstuff so this is visual bug in this PR? I was misleading by first comment :): 
> **just testing comments** icons became invisible on white background

clxandstuff:
I was not actually reviewing PR but testing Chromatic, but this PR bug comment is real :)

src/components/icon-as-button/_icon-as-button.scss
```diff
@@ -27,16 +27,16 @@ $includeHtml: false !default;
 
 @mixin colorTheme($modifierName, $color, $colorActive, $actionForegroundColor: $iconAsButtonActionColor) {
   &.sg-icon-as-button--#{$modifierName} {
-    fill: $color;
+    color: $color;
```
coderitual, line 30: @katarzynatobis @clxandstuff probably we need to add story for this deprecated button.Not necessary in this PR.

Previously we skipped this component in storybook because we didn't  know it would be useful for chromatic.

src/components/buttons/_buttons.scss
```diff
@@ -11,6 +11,12 @@ $largeButtonSize: componentSize(l);
   }
 }
 
+@mixin iconColor($color) {
```
coderitual, line 14: @katarzynatobis `buttonIconColor` ? `iconColor` is already defined mixing and this can be easily confused.

coderitual (changes requested):
we need to change mixin name

src/components/buttons/_buttons.scss
katarzynatobis: 👍🏻

coderitual (approved):

clxandstuff (approved):

PR #2211, opened 2021-12-01
A11y tests for Dialog and Accordion (using: axe, testing-library)

katarzynatobis:
Added accessibility tests for:
- `Dialog`
- `Accordion`

Full documentation [here(internal)](https://brainly.atlassian.net/wiki/spaces/DesignSystem/pages/940015977/Components+Status+IN+PROGRESS)


src/components/accordion/Accordion.spec.jsx
```diff
@@ -62,24 +64,30 @@ describe('<Accordion>', () => {
   it('expands one item at a time when "allowMultiple" is set to false', () => {
     const accordion = mount(
       <Accordion>
-        <AccordionItem title="Item 1">Accordion Item Description</AccordionItem>
-        <AccordionItem title="Item 2">Accordion Item Description</AccordionItem>
-        <AccordionItem title="Item 3">Accordion Item Description</AccordionItem>
+        <AccordionItem title="Item 1" id="item1">
```
coderitual, line 67: @katarzynatobis do we need this change? we don't need `id` if component is uncontrolled

src/components/accordion/Accordion.jsx
```diff
@@ -102,7 +102,7 @@ const Accordion = ({
     invariant(
       !(isControlled && !onChange),
       // eslint-disable-next-line max-len
-      ' You provided an `expanded` prop to a Accordion without an `onChange` handler. Users won`t be able to switch between expanded/collapsed state.'
+      'You provided an `expanded` prop to a Accordion without an `onChange` handler. Users won`t be able to switch between expanded/collapsed state.'
```
coderitual, line 105: one more typo: an accordion

katarzynatobis: 👍🏻

src/components/accordion/Accordion.spec.jsx
katarzynatobis: It's unnecessary, I added these identifiers for a11y testing and overlooked this change when I moved the a11y test to a separate file.

coderitual: cool, please make sure to remove ids whenever it's unnecessary to not make impression it's needed

coderitual (approved):

src/components/dialog/Dialog.a11y.spec.jsx
```diff
@@ -0,0 +1,110 @@
...
+      </Dialog>
+    );
+
+    expect(dialog.getByRole('dialog')).toEqual(document.activeElement);
+  });
+
+  it('closes on Esc key', () => {
+    const dialog = render(<Dialog open>content text</Dialog>);
+
+    fireEvent.keyDown(dialog.getByRole('dialog'), esc);
```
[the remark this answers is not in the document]
katarzynatobis: I switched to user-event. 
The significant change is in 'KeyboardEvent':
- in fireEvent you specify the target,
- in user-event target must be focused before (`document.activeElement` must be set).
User can focus something in 2 ways, by clicking or tabbing. Tabbing (the real tabbing) in not available here, click will not do the job, so focus needs to be fired directly, not in the standard (user-like) way. `fireEvent.focus` just fires onfocus() event handler and do not set `document.activeElement`, but `element.focus()` does.

PR #2217, opened 2021-12-09
Add Mobile Keyboard Icon as Standard Icon

bartoszlorek:
It's needed in the monorepo for the Quick Results v2 project

Screenshot 2021-12-09 at 11 13 15


matzimowski (approved):
👏

coderitual:
hey @bartoszlorek , probably it will take some time until you will be upgrade sg in the app. We are in the middle of color migrations with a lot of breaking changes. We can sync.

bartoszlorek:
@coderitual I can inline this SVG file as a workaround for our MVP experiment, and then replace it with the SG one when it will be possible.

coderitual (approved):

coderitual:
@bartoszlorek awesome. Let’s merge it then and we will take care of releasing and later let you know when it will be possible to replace in the app

PR #2220, opened 2021-12-10
Avatar a11y improvements

katarzynatobis:
* alt added
* link label added
* visible focus style

[Internal documentation](https://brainly.atlassian.net/wiki/spaces/DesignSystem/pages/949387656/Avatar)


src/components/avatar/Avatar.jsx
```diff
@@ -62,41 +64,36 @@ const Avatar = ({
...
-  }
-
-  return (
+  const avatarContent = (
     <div {...props} className={avatarClass}>
-      {link !== undefined && link !== '' ? (
-        <a href={link} title={title}>
-          {avatarContent}
-        </a>
+      {' '}
```
coderitual, line 72: why we need this?

src/components/avatar/Avatar.a11y.spec.jsx
```diff
@@ -0,0 +1,52 @@
...
+  it('renders empty avatar ', async () => {
+    const alt = 'alt';
+    const avatar = render(<Avatar alt={alt} />);
+
+    await testA11y(avatar.container);
+    expect(avatar.queryByRole('img')).toBeFalsy();
+    expect(avatar.queryByAltText(alt)).toBeFalsy();
+  });
+
+  it('renders empty avatar with link and label', async () => {
```
coderitual, line 17: What happens if link with label will be provided along with alt text for image? Is it ok? Can we test it?

src/components/avatar/Avatar.jsx
```diff
@@ -49,7 +50,8 @@ const Avatar = ({
   imgSrc,
   className,
   link,
-  title,
+  linkLabel,
```
coderitual, line 53: I am not sure if having 2 separate props is the best approach here. What is the value if i provide both (alt+linkLabel), can we discuss?

src/components/avatar/Avatar.jsx
katarzynatobis: good catch 👍🏻

src/components/avatar/Avatar.a11y.spec.jsx
katarzynatobis: Yes, this is a correct usage. Ex. link to another user's profile (label: "User's profile", alt text: "John Doe"). I'll add a test for that case.

src/components/avatar/Avatar.jsx
coderitual: I think that this name might be a bit misleading since it doesn't provide visual label to the tree. let's discuss tomorrow approach to aria props

src/components/avatar/Avatar.a11y.spec.jsx
```diff
@@ -0,0 +1,70 @@
+import * as React from 'react';
+import {render, screen} from '@testing-library/react';
+import Avatar from './Avatar';
+import {testA11y} from '../../axe';
+import userEvent from '@testing-library/user-event';
+
+describe('Avatar a11y', () => {
+  describe('imgSrc and link is not provided', () => {
+    it('renders avatar removed from a11y tree', async () => {
```
katarzynatobis, line 9: @clxandstuff here is an example of what we discussed

coderitual (approved):

clxandstuff (approved):

coderitual (approved):
Thanks @katarzynatobis for handling invariants! Regarding checka probably we hit chromatic limits this month. Let’s take a look on Mon

PR #2237, opened 2021-12-17
FileHandler a11y imprv

katarzynatobis:
* proper role on interactive element, depending on provided props
* uploaded / loading status hidden visually with `aria-live="polite"`
* a11y tests
* label for close button
* thumbnail/icon hidden from accessibility tree
* link is tabbable, when has `role="button"` and `href` is not provided

[Internal docs](https://brainly.atlassian.net/wiki/spaces/DesignSystem/pages/951517464/FileHandler)

### DIFF
--- a/src/components/file-handler/FileHandler.a11y.spec.jsx
+++ b/src/components/file-handler/FileHandler.a11y.spec.jsx
@@ -1,0 +1,65 @@
+import * as React from 'react';
+import {render} from '@testing-library/react';
+import FileHandler from './FileHandler';
+import {testA11y} from '../../axe';
+import userEvent from '@testing-library/user-event';
+
+describe('FileHandler', () => {
+  it('fires onClick on click, space and enter', () => {
+    const handleOnClick = jest.fn();
+    const fileName = 'file name';
+    const fileHandler = render(
+      <FileHandler onClick={handleOnClick}>{fileName}</FileHandler>
+    );
+
+    userEvent.click(fileHandler.getByText(fileName));
+    fileHandler.getByText(fileName).focus();
+    userEvent.keyboard('{space}');
+    userEvent.keyboard('{enter}');
+
+    expect(handleOnClick).toHaveBeenCalledTimes(3);
+  });
+
+  it('passes src to link href and does not act on space', () => {
+    const src = '#src';
+    const fileName = 'file name';
+    const fileHandler = render(<FileHandler src={src}>{fileName}</FileHandler>);
+
+    fileHandler.getByText(fileName).focus();
+    userEvent.keyboard('{space}');
+
+    expect(fileHandler.getByRole('link').href).toContain(src);
+    expect(window.location.href).not.toContain(src);
+  });
+
+  it('should have status in accessibility tree', () => {
+    const fileHandler = render(<FileHandler loading />);
+
+    expect(fileHandler.getByText('loading')).toBeTruthy();
+  });
+});
+
+describe('FileHandler a11y', () => {
+  it('should have no a11y violations', async () => {
+    await testA11y(<FileHandler />);
+  });
+  it('should have no a11y violations when loading', async () => {
+    await testA11y(<FileHandler loading>loading</FileHandler>);
+  });
+
+  it('should have no a11y violations when thumbanailSrc and src are provided', async () => {
+    await testA11y(
+      <FileHandler thumbnailSrc="#" src="#">
+        fileName.jpg
+      </FileHandler>
+    );
+  });
+
+  it('should have no a11y violations when onClick is provided', async () => {
+    const handleOnClick = jest.fn();
+
+    await testA11y(
+      <FileHandler onClick={handleOnClick}>file name</FileHandler>
+    );
+  });
+});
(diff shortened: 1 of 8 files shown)


src/components/file-handler/FileHandler.jsx
```diff
@@ -127,28 +144,36 @@ const FileHandler = ({
...
       </div>
       <span className="sg-file-handler__text" ref={textRef}>
+        <span className="sg-visually-hidden" aria-live="polite">
+          {loading
+            ? statusLabel?.loading || 'loading'
+            : statusLabel?.uploaded || 'uploaded'}
+        </span>
         {src !== undefined ? (
-          <Link {...clickProps} size="small" color="text-black">
+          <Link
```
katarzynatobis, line 171: @coderitual @domidomi @clxandstuff
Here is that troublesome tag, which is some situations (`onClick` provided`) should be a button

katarzynatobis: `role="button"` added, I followed guidelines from this article: https://www.deque.com/blog/accessible-aria-buttons/

coderitual: thanks for the update @katarzynatobis

src/components/file-handler/FileHandler.jsx
```diff
@@ -22,6 +22,11 @@ export const FILE_HANDLER_COLORS_SET = {
   WHITE: 'white',
 };
 
+export interface AriaStatusLabelType {
```
coderitual, line 25: Let's use `type` instead of interface here. Interface in flow has similar use cases to type but much wider and intention might be different - interface is the only way to use `implement` keyword on classes level or to merge types defined in different files. I would stick to using type if we don't need interface.

Also think it's internal type and we don't need to export it.

src/components/file-handler/FileHandler.jsx
```diff
@@ -118,37 +135,67 @@ const FileHandler = ({
...
+      e.preventDefault();
+      return onClick();
+    });
+
+  const isActionProvided = src !== undefined || preventedOnClick;
 
+  const buttonEventProps = {
+    onClick: preventedOnClick,
+    onKeyDown: e => {
+      if (e.keyCode === 32 || e.keyCode === 13) {
```
coderitual, line 150: maybe small comment here?
// 32 - space, 13 - enter

src/sass/_utils.scss
```diff
@@ -73,3 +73,15 @@
     box-shadow: $shadowLarge;
   }
 }
+
+.sg-visually-hidden {
```
coderitual, line 77: ❤️

src/components/file-handler/FileHandler.jsx
```diff
@@ -118,37 +135,67 @@ const FileHandler = ({
...
-        src={thumbnailSrc}
-        alt=""
-        className="cursor-pointer"
-      />
+      <img src={thumbnailSrc} alt="" className="cursor-pointer" />
     ) : (
-      <a {...clickProps}>
-        <Icon type={iconType} size={24} color={ICON_COLOR['icon-black']} />
-      </a>
+      <Icon type={iconType} size={24} color={ICON_COLOR['icon-black']} />
```
coderitual, line 169: I think we can use just `color='icon-black'` here

src/components/file-handler/FileHandler.a11y.spec.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+    userEvent.click(fileHandler.getByText(fileName));
+    fileHandler.getByText(fileName).focus();
+    userEvent.keyboard('{space}');
+    userEvent.keyboard('{enter}');
+
+    expect(handleOnClick).toHaveBeenCalledTimes(3);
+  });
+
+  it('passes src to link href and does not act on space', () => {
```
coderitual, line 23: Could we please clarify this test a bit?

I understand it as follows:
- When I pass src to file handler, src is being passed to internal link element
- When I pass src to file handler, hitting spacebar doesn't do anything
- What if I hit enter? Should I do something? maybe we cover both cases in this test?

src/components/file-handler/FileHandler.stories.jsx
```diff
@@ -30,6 +30,8 @@ export default {
   },
 };
 
+const src = 'https://source.unsplash.com/240x240/?cat';
```
coderitual, line 33: can we use more static image here? I know it's haw it was defined but maybe it's a good occasion to change it here? 

cc @clxandstuff

src/components/file-handler/FileHandler.jsx
katarzynatobis: Added

src/components/file-handler/FileHandler.jsx
katarzynatobis: ok 👍🏻

src/components/file-handler/FileHandler.a11y.spec.jsx
katarzynatobis: I hope now it's a bit more clear

src/components/file-handler/FileHandler.stories.jsx
katarzynatobis: of course :) We event can use a dog picture 🐶

src/components/file-handler/FileHandler.jsx
katarzynatobis: ok :)

coderitual (approved):

clxandstuff (approved):

PR #2241, opened 2021-12-22
Story for IconAsButton

katarzynatobis:


coderitual:
Probably we need story for chromatic (this could be generated using script @clxandstuff prepared)

clxandstuff:
@katarzynatobis
You can generate chromatic stories file from regular storybook stories. Try this script:

```sh
$ node scripts/generate-chromatic-stories.js $path-to-stories-file
```

clxandstuff (approved):

PR #2255, opened 2022-01-11
use prettier with eslint and stylelint

clxandstuff:
Goal of this change is to use prettier along with eslint and style lint.

List of changes:
- sass-lint replaced by style-lint
- style-lint configured similar to brainly-frontend for better contribution experience
- eslint reconfigured
- prettier configured
- SCSS, MDX, JS* files reformatted with prettier
- `prettier-check` & `prettier-write` npm scripts added
- updated docs


.prettierignore
```diff
@@ -0,0 +1,5 @@
+src/images/icons.js
```
coderitual, line 1: maybe glob pattern `src/images/**/*.js`?

coderitual (approved):

.prettierignore
clxandstuff: @coderitual 
hmm why not, but I'm thinking of someone manually adding script there for images manipulation will miss prettier formatting? I'm not sure if this can really happen?

coderitual: probably it means that `src/images/` is not the best folder to keep generated js files there

PR #2256, opened 2022-01-12
One webpack

clxandstuff:
We have 4 full webpack configs in the project, what makes any build changes difficult to apply. After this change we will have one base webpack config which is used in all build scripts. Of course some build specific config params e.g. `output` are overridden.


webpack.config.js
```diff
@@ -12,6 +12,7 @@ const SOURCE_DIR = path.join(__dirname, 'src');
 const DIST_DIR = path.join(__dirname, 'dist');
 const VERSIONED_DIST_DIR = path.join(DIST_DIR, VERSION);
 const SOURCE_DOCS_DIR = path.join(SOURCE_DIR, 'docs');
+const SOURCE_DOCS2_DIR = path.join(SOURCE_DIR, '_docs');
```
coderitual, line 15: how about `SOURCE_DOCS_STORYBOOK_DIR`

clxandstuff: Done

coderitual:
@clxandstuff I see a lot of warning after running `yarn build-storybook` and `yarn storybook` scripts:

```
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding  ["@babel/plugin-proposal-private-property-in-object", { "loose": true }]
to the "plugins" section of your Babel config.
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-proposal-private-methods.```

Do you know if that's something related to the new shared webpack or it was before that change?

**Comment by clxandstuff** [comment]
I think I've seen it before. I will check it

**Comment by clxandstuff** [comment]
@coderitual
Warnings fixed

**Comment by clxandstuff** [comment]
Team, please check also changes in `package.json`. This is just optional cleaning of unnecessary script arguments, but I can revert it if you don't like it

**Review by coderitual — APPROVED** [review]
great cleanup, thanks 💯

File: .gitignore [hunk 2]
```diff
@@ -17,7 +17,7 @@ node_modules
 coverage/
 package-lock.json
 yarn-error.log
-.storybook-static
+storybook-static
```
coderitual, line 20: This change is resulting in a lot of tracked files generated by `build-storybook` command. Is it on purpose?

coderitual: @clxandstuff oh it might be that I see previously generated files being tracked after I switched branches, no issue then

PR #2263, opened 2022-01-17
Add mobile icons

katarzynatobis:
Screenshot 2022-01-17 at 14 01 36


src/images/mobile-icons/circle_fill.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 512 512" xmlns="http://www.w3.org/2000/svg">
+  <title>circle fill</title>
+  <ellipse cx="256" cy="256.004" rx="157.538" ry="157.541" fill="black"/>
```
coderitual, line 3: I think we can remove fill from all icons so if they are moved to the web icons no one will forget about it.

coderitual (approved):

PR #2274, opened 2022-01-27
text components responsive props

clxandstuff:
Responsive props for:
- Headline
- Subheadline
- Text
- Link
- Textbit

CSS size:
before: 141 049 bytes / 17 532 bytes (gzipped)
after: 150 959 bytes / 18 517 byes (gzipped)


coderitual:
@clxandstuff could you please provide css size difference for not gzipped content? While loading time won't be affected much it's always good to see how it may affect parsing time.

src/components/utils/responsive-props.spec.js
```diff
@@ -0,0 +1,79 @@
+import {generateResponsiveClassNames} from './responsive-props';
```
coderitual, line 1: @clxandstuff 
I am not sure that this test is enough. We should always test public api in the first place, and while this helper might be working correctly, it doesn't mean it's being used correctly in components. 

I am aware that flex component has much better coverage in this regard. Maybe outside of this PR we may consider our approach to testing responsive props on components level so we can incorporate better solution in the future.

coderitual (approved):

src/components/utils/responsive-props.spec.js
clxandstuff: I agree. There are a lot of tests to cover, but I think I could write helper to make it less bloated. I'll give it a try in this PR.

clxandstuff: @coderitual check `Flex.spec.js file` how tests look. I'm gonna wrap common logic to helper but first would like to confirm if this is good enough?

coderitual:
@clxandstuff I only spotted small baseline differences in chromatic. Those might be caused by the fact that chromatic didn't work for some part of the last month.

clxandstuff:
@coderitual
Actually it was bug in `mergeStories` and fixed it.

PR #2279, opened 2022-02-02
add buildspecs for styleguide

clxandstuff:


scripts/tasks/root-redirect-page.js
```diff
@@ -8,12 +8,20 @@ module.exports = function (gulp, plugins, consts) {
       'utf8'
     );
 
+    let version;
+
+    if (consts.VERSION.includes('-beta')) {
```
coderitual, line 13: does it mean that we always put to the same folder?

coderitual: I see, thx

coderitual (approved):

PR #2293, opened 2022-02-08
Fix text related mixins

matzimowski:
**Note**:

This changes do not affect style-guide itself from visual perspective and are related only to the specific case: importing exposed mixins to style custom elements.

It's partially related to the reported production bug: https://brainly.atlassian.net/browse/PBUGS-1444

**Case:**

when more than 1 text-related mixin was directly imported in a project:
```
@import '~brainly-style-guide/src/components/text/text';
@import '~brainly-style-guide/src/components/text/headlines';
```
due to usage of the same variable names, they were overriding each other and the values were taken from the one that was imported later.

as a result some body texts received the sizes of headlines:

## before

Screenshot 2022-02-08 at 07 51 50
Screenshot 2022-02-08 at 07 51 57

## after

Screenshot 2022-02-08 at 07 38 55
Screenshot 2022-02-08 at 07 38 59

### DIFF
--- a/src/components/text/_headlines.scss
+++ b/src/components/text/_headlines.scss
@@ -1,10 +1,10 @@
 @import './text-colors-mixin';
 
 $includeHtml: false !default;
-$sizes: (xxsmall xsmall small medium large xlarge xxlarge xxxlarge);
-$transforms: (uppercase lowercase capitalize);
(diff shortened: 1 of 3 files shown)


bartoszlorek (approved):

coderitual (approved):
❤️

PR #2299, opened 2022-02-14
Dialog - Fix Unimplemented TransitionEnd Event

bartoszlorek:
Currently both `onEntryTransitionEnd` and `onExitTransitionEnd` callbacks aren't fired when environment doesn't support transition events, e.g. the JSDOM. It requires additional mocking in every test which is not obvious. Furthermore, it doesn't work in older browsers (minor).

- https://github.com/jsdom/jsdom/issues/1781
- https://github.com/testing-library/dom-testing-library/pull/865

### DIFF
--- a/src/components/dialog/Dialog.spec.jsx
+++ b/src/components/dialog/Dialog.spec.jsx
@@ -43,7 +43,6 @@ describe('<Dialog>', () => {
 
   it('fires onDismiss callback on Escape key', () => {
     const onDismiss = jest.fn();
-
     const wrapper = mount(
       <Dialog onDismiss={onDismiss} open>
         content text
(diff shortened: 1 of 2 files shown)


matzimowski (approved):

src/components/dialog/Dialog.spec.jsx
```diff
@@ -68,16 +67,13 @@ describe('<Dialog>', () => {
...
+
+    mount(
       <Dialog onEntryTransitionEnd={onEntryTransitionEnd} open>
         content text
       </Dialog>
     );
 
-    wrapper.find('[role="dialog"]').simulate('transitionEnd', {
-      propertyName: 'transform',
-    });
```
bartoszlorek, line 79: ...and the simulation is no longer needed to just open/close a dialog.

coderitual (approved):
Great, thx @bartoszlorek

PR #2314, opened 2022-02-21
Dialog SG docs

domidomi:
Dialog section was created in the styleguide docs, though there is no Dialog visual representation. The reason is Dialog displays over the whole page and applies styling to the `body`, plus it doesn't show until transition is finished. There is a way of displaying it in the iframe, but:
1. We would have to make controls in interactive docs working, so it means passing data to the iframe.
2. We would have to apply SG styling and scripts.

We made a decision, that it is not worth to spend time on this, as we want to migrate fully to storybook docs in the next quarter. Plus we will have to tackle similar issue in storybook MDX docs - once this is achieved, then we can focus on translating the solution to the styleguide docs.


src/components/dialog/useFocusTrap.js
```diff
@@ -32,10 +32,7 @@ export function useFocusTrap({
     }
 
     function handleFocusTrap(event: FocusEvent) {
-      if (
-        event.target instanceof Node &&
-        dialogElement.contains(event.target)
-      ) {
+      if (event.target.nodeName && dialogElement.contains(event.target)) {
```
coderitual, line 35: This was to satisfy flow which probably won't get duck typing as a type guard I guess.

clxandstuff (approved):

coderitual:
@domidomi is that PR still relevant? If so could you please provide description?

domidomi:
> @domidomi is that PR still relevant? If so could you please provide description?

@coderitual yes it is relevant, I am going to add the description and merge it today :)

PR #2317, opened 2022-02-24
Added favicon with notification elipse for tutoring

mkudej:
Added new favicons to notify users about tuor found / tutor not found on brainly-frontend.
More informations can be found in this PR: https://github.com/brainly/brainly-frontend/pull/19475

image


domidomi:
How about other markets that don't use "B" favicon? Do we need to add these ones too?

clxandstuff:
@domidomi
All markets uses same favicon with "B".

clxandstuff (approved):

domidomi:
@clxandstuff we also have "Znanija", "Eodev" and "Nosdevoirs"... Am I wrong? 😅

coderitual:
@clxandstuff , @domidomi is actually right, we use different fav icons on other markets.

@mkudej I know you don't need other martkets but maybe it would be good idea if designer provide notification favicons for all markets to keep them consistent

mkudej:
@domidomi Right now this icon is only for tutoring - so just India and US markets :)
@coderitual Designer is currently on vacation, but I will inform her asap about other markets.

For now, feature is needed only for brainly, so that's fine.

coderitual:
@mkudej Sure, could you please create an issue in this repo which that information that other icons should be added so we don't forget. Then you can PR separate PR when designer is back. Thanks!

PR #2327, opened 2022-03-03
Added onItemSelect handler to dropdown

WJelen:
Added onItemSelect handler required for new footer implementation.


src/components/dropdowns/Dropdown.spec.jsx
```diff
@@ -63,4 +63,23 @@ describe('<Dropdown />', () => {
     component.update();
     expect(component.find('.sg-dropdown__items')).toHaveLength(0);
   });
+
+  it('should pass event and link to onSelectHanlder', () => {
```
clxandstuff, line 67: Could you change description to something like 'should call onSelectHanlder when link is clicked' ?

src/components/dropdowns/Dropdown.spec.jsx
```diff
@@ -63,4 +63,23 @@ describe('<Dropdown />', () => {
...
+    const onItemSelect = jest.fn();
+    const component = mount(
```
clxandstuff, line 69: Do you need `mount` instead of `shallow`?

clxandstuff (changes requested):

src/components/dropdowns/Dropdown.spec.jsx
WJelen: will check

WJelen: The only small disadvantage is, it will not test if the function received anything as the first argument (`event` object in this case), and we can only assert on the second one. (as we are missing event system in shallow, it will receive `undefined`)
If you are ok with that I can switch to `shallow` :)

WJelen: changed in 7a6cfdac

src/components/dropdowns/Dropdown.spec.jsx
WJelen: fixed in 7a6cfdac

src/components/dropdowns/Dropdown.spec.jsx
clxandstuff: Thanks, looks good

clxandstuff (approved):

Issue #2312, opened 2022-02-19
Dialog scrollbars causing unexpected shift during show animation


coderitual:
When dialog is animated (e.g. while showing) additional scrollbars are shown causing unexpected horizontal shift. This effect is only visible for with overlay scrollbar disabled (overlay scrollbars are enabled by default when using trackpad, magic mouse etc. and disabled using normal mouse.

When overlay scrollbars are disabled:

https://user-images.githubusercontent.com/8572321/154801159-ef6d3522-3c3b-488e-96df-72a831af345b.mp4

PR #2328, opened 2022-03-03
Prevent Dialog scrollbars from showing up during show transition

domidomi:
Related issue: https://github.com/brainly/style-guide/issues/2312


src/components/dialog/Dialog.jsx
```diff
@@ -97,6 +100,7 @@ function BaseDialog({
 
     if (!supportsTransitions()) {
       fireTransitionEndCallbacks();
+      setHasFinishedTransition(true);
```
coderitual, line 103: Why not to put `setHasFinishedTransition(true);` in `fireTransitionEndCallbacks`?

coderitual (approved):

src/components/dialog/Dialog.jsx
```diff
@@ -100,6 +108,20 @@ function BaseDialog({
...
   }, [open, fireTransitionEndCallbacks]);
 
+  React.useEffect(() => {
+    if (open) {
+      const dialogHeight =
+        containerRef.current &&
+        containerRef.current.getBoundingClientRect().height;
+
+      const overlayHeight =
+        overlayRef.current && overlayRef.current.getBoundingClientRect().height;
```
coderitual, line 118: maybe just `overlayRef?.current.getBoundingClientRect().height`?

domidomi: @coderitual `containerRef?.current?.getBoundingClientRect().height;` would work

coderitual: oh sry we probably don't need first condition, so just `containerRef.current?.getBoundingClientRect().height;`

src/components/dialog/Dialog.jsx
```diff
@@ -138,7 +156,9 @@ function BaseDialog({
   );
 
   const overlayClass = cx('js-dialog', 'sg-dialog__overlay', {
-    'sg-dialog__overlay--scroll': scroll === 'outside',
+    'sg-dialog__overlay--scroll':
+      (showScrollbarsInitially || hasFinishedTransition) &&
```
coderitual, line 160: @domidomi can we rename `showScrollbarsInitially` to something more meaningful? We can name it by reason or by effect but the next. condition is by reason and this one by effect. We shouldn't mix those 2 appraoches.

clxandstuff (approved):

PR #2329, opened 2022-03-04
Link fixes

katarzynatobis:
- onClick added to Link as "a"
- fixed onClick type

### DIFF
--- a/src/components/text/Link.jsx
+++ b/src/components/text/Link.jsx
@@ -59,7 +59,7 @@ export type LinkPropsType = {
   className?: ?string,
   inherited?: boolean,
   'aria-label'?: string,
-  onClick?: (e: KeyboardEvent | MouseEvent) => mixed,
+  onClick?: MouseEventHandler,
   ...
 };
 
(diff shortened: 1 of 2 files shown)


src/components/text/Link.jsx
```diff
@@ -59,7 +59,7 @@ export type LinkPropsType = {
   className?: ?string,
   inherited?: boolean,
   'aria-label'?: string,
-  onClick?: (e: KeyboardEvent | MouseEvent) => mixed,
+  onClick?: (e: MouseEvent) => mixed,
```
coderitual, line 62: Hey @katarzynatobis , I think that this has been an old approach, the proper type would be something like:
`(SyntheticMouseEvent<HTMLButtonElement | HTMLAnchorElement >) => void`,  or `(SyntheticMouseEvent<HTMLElement>) => void` (I am not sure if the former will work)

Please be sure to run `yarn js-typecheck` and see if typescript is generated correctly.

coderitual (approved):
only types to be fixed

src/components/text/Link.jsx
katarzynatobis: fixed :)

PR #2332, opened 2022-03-07
Create Basic Transition Component

bartoszlorek:
https://brainly.atlassian.net/wiki/spaces/DesignSystem/pages/925270192/Transition+Component

todo:
- [x] component interface definitions
- [x] component mount/unmount on `transitionEnd`
- [x] `transitionEnd` callbacks with handling unsupported environment
- [x] parsing `PropertyObject` with predefined values (`s`, `m`, `l`, etc.)
- [x] CSS Transition based animator with limited additive motion
- [ ] WAAPI based animator (possible)
- [ ] delay


matzimowski (approved):
i am so impressed how you implemented all this @bartoszlorek  🙇 
cannot wait to use it in one of our features 🐎 ❤️ 

huge kudos for this one! fantastic job!

src/components/transition/classNamesRegistry.js
```diff
@@ -0,0 +1,22 @@
+// @flow strict
+
+export interface ClassNamesRegistryType {
+  register(key: string, value: string): void;
+  toString(): string;
+}
+
+export function createClassNamesRegistry(): ClassNamesRegistryType {
```
coderitual, line 8: Where is this function being used?

src/components/transition/CSSTransitionAnimator.js
```diff
@@ -0,0 +1,205 @@
...
+    if (willChangeProps.height) {
+      element.style.height = height.value === 'auto' ? '' : height.value;
+    }
+
+    if (willChangeProps.opacity) {
+      element.style.opacity = opacity.value;
+    }
+
+    // repaint in case of further changes
+    forceRepaint(element);
```
coderitual, line 154: I am not sure why we need to sync reflow here. Comment above is not really helpful

bartoszlorek: It should be called between sync `addElementStyles()`. I can move it where they are called for better understanding.

coderitual: Thx

src/components/transition/classNamesRegistry.js
coderitual: Sure, i just wasn't able to find any import of this function, neither in examples nor in tests.

coderitual: ok I see it, diff was hidden loaded, sorry

coderitual:
The last missing part is probably exposing public interface of this component in top level `index.js`.`Transition` component is pretty straightforward but I am not sure about `predefinedEffect` name.

We can sync about if that after holiday. Good work anyway :)

src/components/transition/Transition.stories.jsx
```diff
@@ -0,0 +1,70 @@
...
+  <Transition {...args}>
+    <Box color="blue-40" padding="m">
+      <Text color="text-white" weight="bold">
+        hello world
+      </Text>
+    </Box>
+  </Transition>
+);
+
+export * from './examples';
```
coderitual, line 70: Could you please rename this folder to `stories`. It's convention we use so it's clear that those file are related.

src/components/transition/CSSTransitionAnimator.js
bartoszlorek: @coderitual btw. I just found a bug related to this part, but the fix should be soon (this week).

src/components/transition/Transition.stories.jsx
bartoszlorek: Especially as there aren't only "examples".

bartoszlorek:
@coderitual `predefinedEffects` are exposed as `Transition.createEffect()` and it's the only way how it should be used. BTW. I would be strongly into adding an `index.js` file to this component to establish a public interface (both logic and types) on the component level not somewhere outside in the repo.

coderitual:
@bartoszlorek I am not sure if we need additional index file. If there is only one piece to be exported to public api (`Transition` component itself) maybe it's not needed. Public API of the style guide is defined inside top level `index.js` not inside specific components folder (we expose `src` to the legacy part of the app which used flow). 

Regarding typescript types. If there is any type that need to be specifically exported additionally to the ones u have used inside code, it need to be exported on root index as well. It's rarely needed. Usually all the types  will be available normally through autogenerated TS declaration file (they will be available to import directly from `brainly-style-guide`). 

btw, something is broken in types generation step, I will take a look after holiday. We can also have short sync just before merge to make sure we're on the same page.

src/components/transition/CSSTransitionAnimator.js
```diff
@@ -217,6 +219,18 @@ export function createCSSTransitionAnimator(
       pushState(DEFAULT_PARSED_PROPS);
       removeElementStyles(element);
     },
-    finished: () => --remainingPropsToChange <= 0,
+    transitionEnd: (event: TransitionEvent) => {
```
coderitual, line 222: @bartoszlorek imho we shouldn't mix event code inside this function. It should be simple 

```
if (--remainingPropsToChange === 0 && finishCallbackRef) {
        finishCallbackRef();
      }
```

the part 
```
 // ignores bubbling events of its own descendants
      if (event.target !== event.currentTarget) {
        return;
      }
```
should be directly. 

Now I understand why called it transitionHandler in the first place, but imho checking `event.target !== event.currentTarget` has nothing to do with animator component. It strongly relies how we add listener to the node (we can e.g stop bubbling with different event)

coderitual: This is not blocker to this PR, but I think we somehow mixing those 2 responsibilities too much.

coderitual: thx!

bartoszlorek: https://github.com/brainly/style-guide/pull/2332/commits/421b9e594fef7679e3475491547be5ef959d82ca

src/components/transition/Transition.jsx
```diff
@@ -208,8 +209,22 @@ function BaseTransition({
   const onTransitionStartRef = React.useRef();
   const onTransitionEndRef = React.useRef();
 
-  onTransitionStartRef.current = onTransitionStart;
-  onTransitionEndRef.current = onTransitionEnd;
+  React.useLayoutEffect(() => {
```
coderitual, line 212: We're gonna need isomorphic version of this effect to make it work in SSR env:

```
export const isBrowser = typeof window !== "undefined";
export const useIsomorphicLayoutEffect = isBrowser ? useLayoutEffect : useEffect;
```

bartoszlorek: True, but I believe we should use such approach with every `useLayoutEffect` in this component.

coderitual: true, anytime we need `useLayoutEffect` we should use `useIsomorphicLayoutEffect`. The only reason we fallback to `useEffect` is that it has the same interface as `useLayoutEffect` and we're sure it won't execute on server.

bartoszlorek: Second question, what should be the result for SSR because there won't be any transition?

- Transitions should NOT apply any styles to the element.
- Transition should apply styles of the target phase, the phase after the transition.
- Transitions without delay should invoke both `transitionStart` and `transitionEnd` callbacks synchronously.
- Transitions with delay should NOT call any callbacks.

bartoszlorek: Ok, then I understand that it won't be a case because `useEffect` won't be executed on SSR.

bartoszlorek: https://github.com/brainly/style-guide/pull/2332/commits/3624afa16d7a4c95b783491fc45d90a26f260fb5

src/components/transition/stories/common/useTransformationState.js
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * The useLayoutEffect hook is required to catch transformation
+   * changes before they will be painted. This could be a problem
+   * during the first render when mounting both the container and
+   * its deep descendant, causing the containerRef to be null.
+   */
+  const didTheFirstRender = React.useRef(false);
+
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
bartoszlorek: The parent node isn't defined in `useLayoutEffect` when the parent element and its children on different tree levels are mounting. This implementation requires additional rethinking and could be moved to the application or exported from the Style-Guide as `useReducedMotion` initiated.

PR #2337, opened 2022-03-09
v210.0.0

domidomi:
## Breaking changes:
- Change Label solid color from indigo-50 to indigo-40 (#2335)
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "209.2.0",
+  "version": "209.3.0",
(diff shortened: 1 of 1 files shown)


katarzynatobis (approved):

coderitual:
Hey @domidomi , isn't that a breaking change considering that the class might be used in php/twig?

domidomi:
> Hey @domidomi , isn't that a breaking change considering that the class might be used in php/twig?

At the moment it is not used in twig, but I think you are right, it could cause visible changes in the app

coderitual (approved):

PR #2341, opened 2022-03-10
Remove icon titles in browser tooltips

### LINKED ISSUE #2316: SVG's hardcoded title is "leaking" for Icon
It might be a duplicate of #1651 but there are not many details there, so it's hard to say.

I noticed that the tooltip shown when the mouse pointer is placed over the Icon element does not always display the correct title.

**How to reproduce:**
Just use any `Icon` and set a custom title: `<Icon title="Not a heart" type="heart" color="icon-black" size={40} />`

Placing the pointer at the edge of `<svg>` element will display a tooltip saying `Not a heart`, but placing the pointer over the `<use>` element will show a `heart` tooltip.

The svgs that are used, have [title](https://github.com/brainly/style-guide/blob/master/src/images/icons/heart.svg?short_path=ea04438#L2) "hardcoded", so when they are "embeded" by [use](https://github.com/brainly/style-guide/blob/master/src/components/icons/Icon.jsx#L606), those titles "leak".

**Examples** from tutoring:
image
image

image
image
image
https://github.com/brainly/style-guide/blob/master/src/images/icons/drawing_mode.svg?short_path=7bd5689#L2

katarzynatobis:
Fixes #2316


src/components/icons/Icon.jsx
```diff
@@ -601,7 +601,10 @@ const Icon = ({
           aria-labelledby={labelledBy}
           focusable="false"
         >
-          <title id={titleId}>{title || titleFallback}</title>
+          <text id={titleId} hidden>
+            {title || defaultTitle}
```
coderitual, line 605: i think this condition will break intention when I set `title=""`. `??` instead?

katarzynatobis: Could you explain why we want the title to be empty? `title=""` makes Icon inaccessible, it doesn't remove it from the accessibility tree. If you want to hide it, `arria-hidden` has to be added. I don't see any other intention to do that.

katarzynatobis: From discussion on deque:
Screenshot 2022-03-10 at 18 51 35

coderitual: It's not about accessibility. As you mentioned for that case I should use aria-hidden. 

We need to provide a way to hide native tooltip. There are 2 cases
- we don't want any tooltip on a element
- we provide custom tooltip solution and label

in both cases I may apply a11y label on an outer element to satisfy accessibility requirement and not be forced to display native tooltip at the same time.

coderitual (approved):

PR #2365, opened 2022-04-01
New tab indicator in link

katarzynatobis:
Indicator as an unicode icon and description (available for screen readers).
Screenshot 2022-04-01 at 14 34 02


coderitual (commented):
hey @katarzynatobis , I think it might be a good idea to add additional story covering that case. It will be also handled in visual testing that way.

katarzynatobis:
@coderitual 
good idea! I already added one

domidomi (approved):

coderitual (approved):
I left small comment in chromatic, just thinking about better link text than "some text" (that we unfortunately used in other stories as well)

eg smth like we have in current style guide taken from production.

CleanShot 2022-04-07 at 17 37 26

PR #2366, opened 2022-04-04
MDX docs

domidomi:
Converted Box, Card, Avatar, Bubble and Counter stories into MDX


src/components/bubble/Bubble.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+        Just wondering if you have any problems with your school work.
+      </span>
+    ),
+  }}
+/>
+
+# Bubble
+
+- [Stories](#stories)
+- [Accesibility](#bubble-accessibility)
```
coderitual, line 37: @domidomi why we need this special anchor (with additional `bubble` prefix?) I think main sections names should be constant between stories.

I am aware that  that this anchor is embedded in separate a11y story but main mdx should be responsible for defining main sections so we can move `accessibility` headline out of a11y.mdx file.

domidomi: I just followed format that Kasia has prepared. But you are right, we could move the accessibility headline to the main MDX!

clxandstuff (approved):

PR #2381, opened 2022-04-08
Filehandler to mdx

katarzynatobis:
- convert Filehandler documentation to mdx
- fix typo in mdx files (Accesibility -> Accessibility)
- move a11y.mdx headinh to stories.mdx file

Screenshot 2022-04-13 at 17 58 26


src/components/file-handler/FileHandler.stories.mdx
```diff
@@ -0,0 +1,106 @@
...
+    },
+    iconType: {
+      control: {
+        type: 'select',
+        options: TYPE,
+      },
+    },
+  }}
+  args={{
+    children: 'userAvatar.jpg',
```
coderitual, line 33: let's not use camel cases for file names :)

src/components/file-handler/FileHandler.stories.mdx
```diff
@@ -0,0 +1,107 @@
...
+      </Flex>
+    )}
+  </Story>
+</Canvas>
+
+### With thumbnail
+
+<Canvas>
+  <Story name="With thumbnail">
+    {args => <FileHandler {...args} thumbnailSrc={src} />}
```
domidomi, line 95: Can we rather try with such approach?:
```
<Story name="With thumbnail" args={{thumbnailSrc: src}}>
    {args => <FileHandler {...args} />}
```
What do you think? I followed such convention, so I guess we should decide to stick to one approach
@coderitual @clxandstuff

domidomi: I just checked how args setup affects the story itself and I think we should follow my approach. See this example from Dropdown:
https://github.com/brainly/style-guide/pull/2384/files#diff-7cb061b0877f860733a6f0d92ad01130d44d2b48e2f3ad7bbfaeb446e3f27ec8R82

When an arg is being set on the Story, rather than component, it renders controls properly!
Screenshot 2022-04-13 at 16 38 36

katarzynatobis: nice catch!

coderitual: @domidomi @katarzynatobis I wouldn't say this is a strict requirement for stories different than default. We have many stories where changing args don't make a lot of sense (e.g. in buttons) but we can agree that we can pass via args if it makes sense.

coderitual:
@katarzynatobis please be sure to update PR description to cover all the changes, I was quite surprised by so many files changed :)

domidomi (approved):

PR #2383, opened 2022-04-08
Footer docs to mdx

katarzynatobis:
Convert Footer documentation to mdx

Screenshot 2022-04-12 at 11 23 51


coderitual (approved):

coderitual:
@katarzynatobis Probably we need better visual example of this component, even though I am not sure if this component is even used on prod. I think we need better footer component than this one at some point.
CleanShot 2022-04-11 at 15 31 01

katarzynatobis:
@coderitual I totally agree! Footer is not used on prod, but it definitely needs some better presentation. I added a story with some 'real' footer content, but without any other styleguide component/

src/components/footer/Footer.stories.mdx
```diff
@@ -0,0 +1,52 @@
...
+<ArgsTable story="Default" />
+
+## Stories
+
+### Full
+
+<Canvas>
+  <Story name="Full">
+    {args => (
+      <Footer>
```
domidomi, line 43: @katarzynatobis Can you please pass `args` to Footer?

domidomi (changes requested):

PR #2393, opened 2022-04-15
[WIP] Dialog decoposition PoC

dpietrzyk:
The purpose of this PR is to split a `Dialog` component into multiple parts in order to be able to add custom elements to it without adding new props and still have all a11y improvements from the original dialog.

The default usage with just a `Dialog` component shouldn't change at all but in the case of more advanced cases, it will be a possibility to compose your own "dialog" fully compliant with the "Style Guide" dialog.

image

image

### DIFF
--- a/src/components/dialog/Dialog.jsx
+++ b/src/components/dialog/Dialog.jsx
@@ -52,13 +52,33 @@ const DialogContext = React.createContext<DialogContextType>({});
 
 export function DialogContextProvider({
   children,
+  open,
   size = 'm',
   scroll = 'outside',
   zIndex = 'auto',
+  onExitTransitionEnd,
   ...otherProps
 }: DialogPropsType) {
   const overlayRef = React.useRef(null);
   const containerRef = React.useRef(null);
+
+  /* HANDLE MOUNTING */
+  const [mounted, setMounted] = React.useState<boolean>(open);
+
+  if (open && !mounted) {
+    setMounted(true);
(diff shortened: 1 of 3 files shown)


coderitual:
Thanks for proposal @dpietrzyk! 

As we discussed before we going fruther:
- making sure we have everything working visually (all properties, specifically fullscreen and inside/outside scroll)
- making sure we don't break any accessibility feature we currently have.
- build and tests are passing
- converting flow to typescript generates correct types (we will the details how to do it)

Could you also please add a story to this PR which simulates usage of the dialog in the same way as it would be used in the app? We'll be also able to preview it directly from this PR (by clicking details on preview step as on screenshot below)

image

cc @clxandstuff @domidomi @katarzynatobis

dpietrzyk:
> is PR which simulates usage of the dialog in the same way as it would be used in the app? We'll be able to preview it directly from this PR (by clicking details on preview step as on screenshot below)

Sure, I'll add stories with our usage :) 

Also, I want to clarify that this is only a PoC and not a "final" proposition, so I focused on "HOW" we can handle this situation, but the code still needs some polishing.

coderitual:
Hey @bartoszlorek, if you have time, take a look at the direction suggested in this PR in terms of possibly using the Transition component inside the dialog as a potential replacement for the current animation implementation.

coderitual:
conversation moved to internal channels

PR #2400, opened 2022-04-27
New Checkbox component

domidomi:
Screenshot 2022-05-11 at 10 13 37
Screenshot 2022-05-11 at 10 13 51
Screenshot 2022-05-11 at 10 14 24
Screenshot 2022-05-11 at 10 14 37


src/components/checkbox/_checkbox.scss
```diff
@@ -0,0 +1,196 @@
+$checkboxWidth: 16px;
+$checkboxMargin: 8px;
+$clickableArea: 32px;
+
+.sg-checkbox-new {
+  @include component();
+
+  display: flex;
```
coderitual, line 8: Shouldn't be `inline-flex`?

clxandstuff:
There is an error when editing "children" control
Screenshot 2022-05-12 at 13 36 45

clxandstuff:
`defaultChecked` prop control is not working. It is only used when going out of story page and back or refreshing page. I think this is because how storybook watch works. We can fix it or consider adding explanation, how to use it.

clxandstuff:
`description` prop control breaks when editing.
Screenshot 2022-05-12 at 13 43 35

src/components/form-elements/checkbox/_checkbox.scss
```diff
@@ -0,0 +1,190 @@
...
+      }
+
+      &:checked,
+      &:indeterminate {
+        & + .sg-checkbox__icon {
+          border-color: $checkedIconColor;
+          background: $checkedIconColor;
+        }
+
+        &[aria-invalid] + .sg-checkbox__icon {
```
katarzynatobis, line 72: <3

src/components/form-elements/checkbox/Checkbox.spec.jsx
```diff
@@ -0,0 +1,113 @@
...
+  it('checks/unchecks when either input or label is clicked', () => {
+    const checkbox = renderCheckbox({children: 'my label'});
+    const checkboxInput = checkbox.getByRole('checkbox');
+
+    expect(checkboxInput.checked).toBe(false);
+    expect(checkbox.queryByLabelText('my label')).toBeTruthy();
+    checkboxInput.click();
+    expect(checkboxInput.checked).toBe(true);
+    checkbox.queryByLabelText('my label').click();
+    expect(checkboxInput.checked).toBe(false);
```
katarzynatobis, line 27: It would be great to test if focus is moved to checkbox when label is clicked

katarzynatobis (changes requested):

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,232 @@
...
+  checked,
+  children,
+  className,
+  color = 'dark',
+  defaultChecked,
+  description,
+  disabled = false,
+  errorMessage,
+  id,
+  indeterminate = false,
```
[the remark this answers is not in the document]
katarzynatobis: Actually `indeterminate` is an IDL attribute of checkbox: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

src/components/form-elements/checkbox/CheckboxIcon.jsx
```diff
@@ -0,0 +1,72 @@
...
+  >
+    {/* eslint-disable max-len */}
+    <path d="M3 8C3 7.44772 3.44772 7 4 7H12C12.5523 7 13 7.44772 13 8C13 8.55228 12.5523 9 12 9H4C3.44772 9 3 8.55228 3 8Z" />
+    {/* eslint-enable max-len */}
+  </svg>
+);
+
+export type CheckboxIconPropsType = {
+  checked?: boolean,
+  className?: string,
```
clxandstuff, line 34: `classname` is not used. Maybe we can remove it?

clxandstuff (changes requested):

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,232 @@
...
+      ids.push(descriptionId);
+    }
+    return ids.join(' ');
+  }, [errorTextId, descriptionId, invalid, errorMessage, description]);
+
+  return (
+    <div {...props} className={checkboxClass}>
+      <label
+        className="sg-checkbox__wrapper"
+        htmlFor={checkboxId}
```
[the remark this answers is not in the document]
domidomi: Resolving this conv, as because of recent changes input is no longer nested inside the label

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -66,6 +66,11 @@ export type CheckboxPropsType = {
    * @example <Checkbox invalid />
    */
   invalid?: boolean,
+  /**
+   * ID of a custom label, that describes the checkbox input.
+   * @example <Checkbox customLabelId="my-custom-label" />
+   */
+  customLabelId?: string,
```
katarzynatobis, line 73: I would stay with `aria-labelledby`

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -186,23 +188,25 @@ const Checkbox = ({
...
-            type="span"
-            weight="bold"
-            className="sg-checkbox__label"
-          >
-            {children}
-          </Text>
+        {!customLabelId && children !== undefined && children !== null && (
+          <label htmlFor={checkboxId}>
+            <Text
+              size="medium"
```
katarzynatobis, line 200: `Text` can have `type="label"` in props

katarzynatobis (dismissed):

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,236 @@
...
+            name={name}
+            onChange={onInputChange}
+            required={required}
+            value={value}
+            aria-checked={indeterminate ? 'mixed' : isChecked}
+            aria-invalid={invalid ? true : undefined}
+            aria-describedby={describedbyIds}
+            aria-labelledby={customLabelId}
+          />
+          <CheckboxIcon checked={isChecked} indeterminate={indeterminate} />
```
coderitual, line 195: I think that CheckboxIcon component should be simple UI element without repeating the same interface as checkbox.

```
checked={isChecked} indeterminate={indeterminate}
```

is interface that doesn't belong to icon and CheckboxIcon should provide simple `type` prop

src/components/form-elements/checkbox/CheckboxIcon.jsx
```diff
@@ -0,0 +1,67 @@
...
+    if (indeterminate) return <IndeterminateIcon />;
+    if (checked) return <CheckIcon />;
+
+    return null;
+  };
+
+  const checkboxIcon = React.useMemo(
+    () => getIcon(indeterminate, checked),
+    [indeterminate, checked]
+  );
```
coderitual, line 52: This code is incorrect. We don't need memoization which is based on all props. For handling derived state we should use simple condition (btw this logic shouldn't be in the icon, it belongs to the parent component). The code above is equivalent of:

```
let icon;
if (checked) icon = <CheckIcon />;
if (indeterminate) icon = <IndeterminateIcon />;
```

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,247 @@
...
+
+  React.useEffect(() => {
+    // If it's a controlled checkbox, set isChecked value
+    if (checked !== undefined) setIsChecked(checked);
+  }, [checked]);
+
+  const onInputChange = React.useCallback(
+    e => {
+      // If it's an uncontrolled checkbox
+      if (checked === undefined) setIsChecked(!isChecked);
```
coderitual, line 139: I would introduce special variable `isControlled = checked === undefined` and reuse it. It will work better than comment imho.

domidomi: @coderitual great idea!

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,247 @@
...
+  );
+  const inputRef = React.useRef(null);
+
+  React.useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.indeterminate = indeterminate;
+    }
+  }, [inputRef, indeterminate]);
+
+  React.useEffect(() => {
```
coderitual, line 131: I am wondering if can get rid of this specific effect

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,247 @@
...
+  const [isChecked, setIsChecked] = React.useState(
+    checked !== undefined ? checked : defaultChecked
+  );
+  const inputRef = React.useRef(null);
+
+  React.useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.indeterminate = indeterminate;
+    }
+  }, [inputRef, indeterminate]);
```
coderitual, line 129: we shouldn't need to check inputRef, is it something eslint is complaining about?

domidomi: you are right, it is not needed in this case

src/components/form-elements/checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,245 @@
...
+    inputRef.current.indeterminate = indeterminate;
+  }, [inputRef, indeterminate]);
+
+  React.useEffect(() => {
+    if (isControlled) setIsChecked(checked);
+  }, [checked, isControlled]);
+
+  const onInputChange = React.useCallback(
+    e => {
+      if (isControlled) setIsChecked(!isChecked);
```
coderitual, line 135: this is quaite dangerous if there are other places which use `isChecked`, better use `setIsChecked(val => !val);`

clxandstuff (approved):

src/components/form-elements/checkbox/Checkbox.stories.mdx
```diff
@@ -0,0 +1,335 @@
...
+    )}
+  </Story>
+</Canvas>
+
+### With custom label
+
+<Canvas>
+  <Story name="With custom label">
+    {args => (
+      <label htmlFor="myCheckboxID">
```
katarzynatobis, line 315: There is no need to duplicate the same function implementation. `htmlFor`, `aria-labelledby` and nesting do exactly the same - link label to the input. So nesting is enough in that case

PR #2401, opened 2022-04-28
Expose use reduce motion from Styleguide

przytrzask:
<Meta title="Utilities/useReducedMotion" />

# useReducedMotion

`useReducedMotion` is a custom hook for detecting the users motion preference.

## Import

```jsx
import {useReducedMotion} from 'brainly-style-guide';
```

## Usage

```jsx
const shouldReduceMotion = useReducedMotion()

<Accordion reduceMotion={shouldReduceMotion}>
    <AccordionItem title={'example'} id="accordion-item-1">
      example body
    </AccordionItem>
  </Accordion>

```


src/components/utils/useReducedMotion/useReducedMotion.stories.mdx
```diff
@@ -0,0 +1,42 @@
+import {ArgsTable, Meta, Story, Canvas} from '@storybook/addon-docs';
+import Accordion from '../../accordion/Accordion';
+import AccordionItem from '../../accordion/AccordionItem';
+
+<Meta title="Utilities/useReducedMotion" />
```
coderitual, line 5: Currently Utilities is mainly for css level utilities. We're going to review the whole documentation structure during the curent storybook but I maybe for this hook we can use `Accessibility/useReducedMotion`

@brainly/design-infra wdyt guys?

coderitual:
@przytrzask please don't forget about description when you'll open PR for the review, thx!

src/components/utils/useReducedMotion/pages/useReducedMotion.jsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
```
coderitual, line 1: you don't need to create separate folder for `useReducedMotion`. Following the pattern we use:
```
utils/
├─ useReducedMotion.jsx
├─ useReducedMotion.stories.mdx
├─ pages/
│  ├─ reduced-motion.jsx

```

src/components/utils/useReducedMotion/useReducedMotion.stories.mdx
coderitual: cc @katarzynatobis @domidomi @clxandstuff

coderitual: @przytrzask you can leave it as it is for now. We will change it to more appropriate group once we internally finalise the structure.

coderitual: @przytrzask You can leave the documentation content as is. We'll take care of moving things to the appropriate sections separately once we adjust the structure. On your part I would just ask you to follow this comment https://github.com/brainly/style-guide/pull/2401#discussion_r860842619

src/components/utils/useReducedMotion.stories.mdx
```diff
@@ -0,0 +1,42 @@
...
+    </AccordionItem>
+  </Accordion>
+
+```
+
+## Example (using styleguide Accordion)
+
+To see in action on MAC open system preferences then accessibility and check "reduce motion" option
+
+![To see in action on MAC open system preferences then accessibility and check "reduce motion" option](/reduced-motion-setting.png)
```
coderitual, line 34: I think `/reduced-motion-setting.png` is not working on preview (sb is published to subfolder). Let's change it to `reduced-motion-setting.png`

przytrzask: @coderitual  After change It doesn't work either (works locally)

coderitual: oh maybe it's something handled by mdx (so `/` is added always). Fortunately we can fix it separately.

src/components/utils/__mocks__/mq.js
```diff
@@ -0,0 +1,114 @@
+// @flow strict
```
coderitual, line 1: Please let me express my doubts about this mock:
- I don't know if it's working correctly (we're testing this mock by our test which relies on it at the same time, this mock doesn't have dedicated tests to cover its behaviour)
- I don't know if it's working the same way as in browsers.

This is generic issue with any kind of mock but in our case we effectively couple mock and hook together.

This is not a blocker but rather something we need to consider more carefully in the future. Anyway thanks for covering at least happy path.

coderitual (approved):

src/components/utils/__mocks__/mq.js
przytrzask: I share your opinion

- I was considering writing test for this mock but I have doubts to the approach itself
- Yes because it is a monkey patching 

But this mock helps to make sure that at leas that hook itself along with listeners are working correctly.

I think the best solution to test this util will be E2E cypress test thus we can force "prefers-reduced-motion" and relay on it.  As we do not have cypress we need to test story manually

PR #2408, opened 2022-05-12
SkipLink component

przytrzask:
# Skip Link

## Showcase

https://user-images.githubusercontent.com/20127089/176418881-0347bfb8-2a4e-4d8d-83f7-b8a843ecf463.mov

##overview
Screenshot 2022-06-29 at 12 48 51


coderitual:
@przytrzask Could we please discuss it on a meeting with me and @katarzynatobis . This is slightly different idea from what we've discussed (`SkipLink`/`SkipToContent` only visual element) and I would like to better understand proposed approach.

Please

coderitual (approved):
hey @przytrzask , we'll fix chromatic story in separeta pr, thanks for contrib!

PR #2410, opened 2022-05-15
Turn Transition docs into MDX

coderitual:
Documentation for the `Transition` component has been moved to MDX (the standard way of providing documentation within the style guide) In the context of the `Transition` component, it is even more important to have fine-grained control over the documentation and how individual elements and examples are presented.

---

CleanShot 2022-05-15 at 23 30 35


bartoszlorek (approved):

src/components/transition/stories/FillMode.jsx
```diff
@@ -27,7 +27,7 @@ const shrinkFadeEffect = {
 };
 
 export const FillMode = () => {
-  const [active, setActive] = React.useState(false);
+  const [show, setShow] = React.useState(false);
```
bartoszlorek, line 30: IMO, it should be `active` as it explains differences between fill modes when the `active` prop changes and it's not related to any widget or UI element like show/hide dialog.

bartoszlorek: Probably you have an issue with button text since "active" and "inactive" aren't verbs and they don't match the current state in a delayed transition. I have the same issue with this example. I think we should leave `active/setActive` variables and change the button text to a verb related to activation.

bartoszlorek (approved):

clxandstuff (approved):

PR #2413, opened 2022-05-16
New checkbox component accessibility

katarzynatobis:
Screenshot 2022-05-16 at 14 06 06


coderitual:
@katarzynatobis please change base to the one with the current implementation so we can see correct changes

PR #2414, opened 2022-05-17
Transition - Create Story Code Snippets

bartoszlorek:


src/components/transition/stories/SharedAxis.jsx
```diff
@@ -8,7 +8,7 @@ import Icon from '../../icons/Icon';
 import DummyBox from './common/DummyBox';
 import Stage from './common/Stage';
 
-const colorsOrder = ['red', 'yellow'];
+const colorsOrder = ['red', 'yellow', 'blue'];
```
bartoszlorek, line 11: it fixes a bug after introducing new button colors, and it moves away from relying on them in favor of icon colors.

coderitual (approved):

PR #2417, opened 2022-05-23
Counter accessiblity improvements

katarzynatobis:
- a11y documentation
- a11y tests
- new prop `aria-label` for the Counter

Screenshot 2022-05-23 at 08 54 53

### DIFF
--- a/src/components/counters/Counter.a11y.spec.jsx
+++ b/src/components/counters/Counter.a11y.spec.jsx
@@ -7,7 +7,7 @@ describe('Counter a11y', () => {
     await testA11y(<Counter aria-label="Label">32</Counter>);
   });
 
-  it('should have no a11y violations whenan icon is provided', async () => {
+  it('should have no a11y violations when an icon is provided', async () => {
     await testA11y(
       <Counter aria-label="Label" icon="points">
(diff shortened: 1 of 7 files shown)


src/components/counters/Counter.a11y.spec.jsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import {testA11y} from '../../axe';
+import Counter from './Counter';
+
+describe('Counter a11y', () => {
+  it('should have no a11y violations when a label is provided', async () => {
+    await testA11y(<Counter aria-label="Label">32</Counter>);
+  });
+
+  it('should have no a11y violations whenan icon is provided', async () => {
```
szemek, line 10:
```suggestion
  it('should have no a11y violations when an icon is provided', async () => {
```

katarzynatobis: good catch, thank you!

coderitual (approved):

coderitual:
Hey @katarzynatobis , I just realised it don't see those colors in handoff template. Were those changes discussed with designer?

PR #2449, opened 2022-06-22
enable adding tooltips to AccordionItem

wrona-d:
After open/close animation change overflow value to display tooltip elements.
Proposed implementation allow to add tooltip inside `Accordion-Item`. 
image


coderitual:
hey @wrona-d , please don't forget to fill description. At the moment I am a bit confused about the proposed implementation (places where you change overflow behaviour).

src/components/accordion/AccordionItem.jsx
```diff
@@ -88,6 +88,7 @@ const AccordionItem = ({
       if (reduceMotion) {
         contentRef.current.style.height = `${0}px`;
         contentRef.current.hidden = true;
+        contentRef.current.style.overflow = 'visible';
```
clxandstuff, line 91: Shouldn't be `hidden` for collapse?

clxandstuff (changes requested):

src/components/accordion/AccordionItem.jsx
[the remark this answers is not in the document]
wrona-d: oh, here should be hidden, and inside expand() overflow value should be set on visible

wrona-d: yes, my bad

clxandstuff:
@coderitual Do you think that it would be good to add story for this case to chromatic?

coderitual:
@clxandstuff If you don't think it's too complicated I am fine. Alternatively this should be covered in unit tests - looking at our comments we expect override prop to have specific value in certain state.

clxandstuff:
@coderitual Overflow is purely visual so I thought that it would be better to test it with snapshot. Just putting some overflowing stuff in accordion item and checking would be sufficient. I would avoid testing visual stuff with unit tests because it violates black-box testing.

I proposed chromatic story because I guess we don't want it in regular storybook. If you're ok with that I can help @wrona-d  adding this story.

clxandstuff (approved):

coderitual (approved):

PR #2451, opened 2022-06-23
remove postfix to React namespace generated by api extractor

clxandstuff:
We're having errors in frontend app, when using class components.
Api extractor is adding to styleguide types React_2 empty namespace. When using this namespace, all members are 'any' type causing ts errors. Class components are not React.Component anymore failing typechecks.


coderitual:
I see that the bug was introduced in recent version of API Extractor:
https://github.com/microsoft/rushstack/issues/2479

We started using newer version of it in this PR https://github.com/brainly/style-guide/pull/2403 after upgrading `flow-to-ts` which had to be upgraded due to not supported types needed in `Transition` component. We can see if there is a version of API extractor without this bug but supported in newest `flow-to-ts`

coderitual (approved):

PR #2459, opened 2022-06-30
change position of storybook sidebar carets

clxandstuff:


clxandstuff:
@coderitual
I think it looks better now 😄

coderitual:
Do you think it's worth keeping this PR @clxandstuff , I think future improvements will cover this case, but I'm not sure if the code from this pull request will be particularly useful. wdyt?

PR #2463, opened 2022-07-05
Transition Component - Fire Callbacks After Instant Animation

bartoszlorek:
When `effect` does not define one of the optional phases like `initial`, `animate`, or `exit`, they become "instant" via default values. This is correct behavior for a CSS Transition. For example, when a user doesn't care about the exit transition they skip it but the `onTransitionEnd` callback should be called anyway. The current implementation relies on CSS Transition that doesn't trigger events for transition with a zero duration. It produces a bug when skipped phases won't trigger proper callbacks.

### before

https://user-images.githubusercontent.com/13873576/177271413-fcee004a-b023-40c8-b387-9ab00a7357be.mp4

### after

https://user-images.githubusercontent.com/13873576/177271445-0bd8c1a1-6363-45b2-b6b6-71e6d92eb9a4.mp4

### DIFF
--- a/src/components/transition/CSSTransitionAnimator.js
+++ b/src/components/transition/CSSTransitionAnimator.js
@@ -38,6 +38,13 @@ export function createCSSTransitionAnimator(
   const hasLastChangedValue = (prop: string) =>
     history[0][prop].value !== history[1][prop].value;
 
+  const getWillChangeProps = () => ({
+    transform: hasLastChangedValue('transform'),
+    width: hasLastChangedValue('width'),
+    height: hasLastChangedValue('height'),
+    opacity: hasLastChangedValue('opacity'),
+  });
+
   /**
    * A single animation may include multiple CSS transitions
    * of different properties and the animator should invoke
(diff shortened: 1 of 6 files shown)


src/components/transition/stories/OffsetBehavior.jsx
```diff
@@ -47,6 +47,11 @@ const contentCounterSlideEffect = {
     transform: {scaleY: 1, duration: 'gentle1', origin: 'left top'},
     easing: 'entry',
   },
+  exit: {
+    opacity: 0,
+    duration: 'quick2',
+    easing: 'exit',
+  },
```
bartoszlorek, line 54: This problem with missing the `exit` phase wasn't visible with an instant animation bug.

src/components/transition/Transition.jsx
```diff
@@ -258,7 +258,7 @@ function BaseTransition({
     }
 
     if (supportsTransitions() && isFillModeBackwards(fillMode)) {
-      animator.animate(container, rules.from);
+      animator.apply(container, rules.from);
```
bartoszlorek, line 261: It's about "prefilling" the element before actual animation (related to filling modes) so from the beginning, it should be a different method than `animate`.

src/components/transition/CSSTransitionAnimator.js
```diff
@@ -213,8 +224,45 @@ export function createCSSTransitionAnimator(
...
+
   return {
-    animate,
+    animate: (
+      element: HTMLElement,
+      from?: PropertyObjectType,
+      to?: PropertyObjectType,
+      speed?: number
+    ) => {
+      const {instant} = executeTransition(element, from, to, speed);
```
bartoszlorek, line 253: We only care about callbacks in the `animate` not `apply` method.

coderitual (approved):
Yea, I totally agree, thx!

Malinicz (approved):

src/components/transition/CSSTransitionAnimator.js
```diff
@@ -213,18 +163,117 @@ export function createCSSTransitionAnimator(
...
+          transitioned: true,
+          parsedProps: toParsedProps,
+          willChangeProps,
+          speed,
+        });
+      }
+
+      if (instant && finishCallbackRef) {
+        finishCallbackRef();
+      }
```
bartoszlorek, line 248: actual change

PR #2471, opened 2022-07-07
Extend dialog for ads

clxandstuff:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "218.0.1",
+  "version": "218.0.1-beta-ads-dialog.1",
```
coderitual, line 3: Marking to not forget removing that

clxandstuff: Thanks. I updated it right now to not forget :)

coderitual (approved):

PR #2484, opened 2022-07-15
Spinner a11y improvements

katarzynatobis:
Accessibility improvements for Spinner:
- `role="status"`
- `aria-live="assertive"` to announce the status
- accessible information about the status
- documentation & tests
- Spinner story moved to mdx

Screenshot 2022-07-15 at 17 17 56


src/components/spinner/Spinner.jsx
```diff
@@ -56,7 +58,13 @@ const Spinner = ({
     className
   );
 
-  return <div {...props} className={spinnerClassNames} />;
+  return (
+    <div {...props} className={spinnerClassNames}>
+      <span className="sg-visually-hidden" aria-live="assertive" role="status">
```
coderitual, line 63: I am not sure aria-live="assertive" should be used on spinner level. This is pure visual element without app logic. Can we discuss?

clxandstuff (approved):

PR #2489, opened 2022-07-21
Checkbox & Radio UX motion

domidomi:
Added animations to Checkbox and Radio


coderitual:
hey @domidomi I just spotted that Checkbox is displayed as flex instead of inline-flex. I must have missed it on the last pr.

coderitual (approved):

PR #2504, opened 2022-08-09
vanilla extract research [for test only]

clxandstuff:


coderitual:
I think we can close it by now, wdyt?

clxandstuff:
@coderitual yess!

PR #2506, opened 2022-08-10
Accessibility improvements for Button

katarzynatobis:
- documentation and tests added
- switched to default cursor
- Button stories rewritten to mdx
- has accessible information about loading state
- native button `type` can be set as `nativeType`
- with href (Button with href should have a role `link` and meet the accessibility requirements for the link):
    - has accessible information about opening a new tab when `target="_blank"` is provided
    - `disabled` is respected
 - documentation for toggle button
 
Screenshot 2022-08-16 at 11 41 04


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "218.0.1",
+  "version": "v218.0.1-beta-a11y-buttons.1",
```
coderitual, line 3: just placing comment to not forget reverting this change

src/components/buttons/Button.spec.jsx
```diff
@@ -16,13 +16,6 @@ test('type', () => {
   expect(button.hasClass(`sg-button--${buttonType}`)).toEqual(true);
 });
 
-test('button with href', () => {
```
coderitual, line 19: let's discuss about our plans removing enzyme entirely. As far as I remember we've been talking about possibly moving both a11y and classic tests to one `.spec` but let's bring back this topic and make a decision.

cc @domidomi @clxandstuff

coderitual (changes requested):
Hey @katarzynatobis , I am requesting removing changes to the package.json, that's the only blocking comment going further.

package.json
katarzynatobis: it is reverted :)

coderitual (approved):

PR #2510, opened 2022-08-16
Add unit test for nested Dialogs

azajac-brl:
~~Allow Brainiac usage for Dialog onDismiss prop by passing event target to callback. Because we want to use events we shouldn't stop their propagation, so as part of this PR I've proposed another solution for dismissing nested Dialogs.~~

Added unit test to ease regression of nested Dialogs closing

### DIFF
--- a/src/components/dialog/Dialog.a11y.spec.jsx
+++ b/src/components/dialog/Dialog.a11y.spec.jsx
@@ -1,5 +1,9 @@
 import * as React from 'react';
-import {render, waitForElementToBeRemoved} from '@testing-library/react';
+import {
+  screen,
+  render,
+  waitForElementToBeRemoved,
+} from '@testing-library/react';
 import DialogHeader from './DialogHeader';
 import DialogBody from './DialogBody';
 import DialogCloseButton from './DialogCloseButton';
@@ -94,5 +98,26 @@ describe('Dialog a11y', () => {
 
     userEvent.click(dialog.getByText(buttonText));
     waitForElementToBeRemoved(dialog.queryByRole('dialog'));
+  });
+
+  it('correctly handles escape key for nested Dialogs', () => {
+    // TODO: improve this test to adhere to react-testing-library standards
+    // so to check modal visibility instead of onDismiss callback
+    const onDismissOuter = jest.fn();
+    const onDismissInner = jest.fn();
+
+    render(
+      <Dialog open onDismiss={onDismissOuter}>
+        <Dialog open onDismiss={onDismissInner} data-testid="inner-dialog">
+          Inner content
+        </Dialog>
+      </Dialog>
+    );
+
+    screen.getByText('Inner content').focus();
+    userEvent.keyboard('{esc}');
+
+    expect(onDismissInner).toBeCalled();
+    expect(onDismissOuter).not.toBeCalled();
   });
 });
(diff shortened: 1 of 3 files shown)


src/components/dialog/Dialog.jsx
```diff
@@ -69,6 +76,7 @@ function BaseDialog({
   const overlayRef = React.useRef(null);
   const containerRef = React.useRef(null);
   const [exiting, setExiting] = React.useState<boolean>(false);
+  const [id] = React.useState(`dialog-${getRandomId()}`);
```
azajac-brl, line 79: ℹ️ Instead of using regular `const` here I'm using state to avoid regenerating id here if component state changes. This wouldn't affect how proposed solution works, but I think the expected behaviour here would be to have one constant id until component is unmounted

coderitual: useRef should be used here since we don't need reset the value during component lifetime.

src/components/dialog/Dialog.jsx
```diff
@@ -46,6 +50,9 @@ Dialog.defaultProps = ({
   scroll: 'outside',
 }: $Shape<DialogPropsType>);
 
+const getRandomId = (length = 9) =>
```
coderitual, line 53: please use `import {generateId} from '../utils';`

src/components/dialog/Dialog.jsx
```diff
@@ -196,6 +208,7 @@ function BaseDialog({
       onClick={onDismiss ? handleOverlayClick : undefined}
       onKeyUp={handleKeyUp}
       ref={overlayRef}
+      id={id}
```
coderitual, line 211: why do we need this?

coderitual: oh I see now, because of stop propagation we weren't able to catch event outside dialog.

azajac-brl:
As discussed with @coderitual this PR includes only adding new unit test for nested Dialog components. Previously this PR enabled passing Brainiac callbacks to `onDismiss` but due to changes in PR in my project this requirement became redundant, hence the change is needed. If, however, this need will arise again two findings from our research:
1. `stopPropagation` used in `onDismiss` does _not_ interfere with Brainiac
2. If, for some reason, one would like to change `stopPropagation` we've concluded that `ev.target.closest('.js-dialog') === ev.currentTarget` added to `handleKeyUp` is the way to go.

TLDR:
FY2ZNIWWAAAQopM

src/components/dialog/Dialog.a11y.spec.jsx
```diff
@@ -95,4 +99,23 @@ describe('Dialog a11y', () => {
     userEvent.click(dialog.getByText(buttonText));
     waitForElementToBeRemoved(dialog.queryByRole('dialog'));
   });
+
+  it('correctly handles escape key for nested Dialogs', () => {
+    const onDismissOuter = jest.fn();
```
coderitual, line 104: Instead of testing calling function the better approach would be to test effect. The effect should be that inner dialog closes while outer dialog stays.

azajac-brl: Discussed on Slack, we've decided to stay with this version for the time being until better solution will be found

coderitual (approved):

PR #2533, opened 2022-09-09
Rename Button type to variant

katarzynatobis:
`type` prop renamed to `variant` and `nativeType` renamed to `type`. This change allows the developer to use the `type` attribute of the native `button` element.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "219.1.2-beta.0",
+  "version": "219.1.2-beta-button-naming.0",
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "brainly-style-guide",
-  "version": "219.1.2",
+  "version": "219.1.2-button-naming-beta.0",
```
coderitual, line 3: to not forget removing before merge

coderitual (approved):

package.json
katarzynatobis: 👍🏻

PR #2544, opened 2022-09-27
dont play Accordion animations on default expanded items

### LINKED ISSUE #2348: Accordion plays expand animation on first render causing layout shifts
Rendering accordion component for the first time (on mount) when some of the elements are expanded by default, results in playing "expand" animation on them. Expected behaviour is to render expanded items on mount without animation even without reduced motion property set.

Current behaviour causing layout shift
CleanShot 2022-03-24 at 13 30 27

clxandstuff:
This changes turns off animations for default expanded items. It fixes layout shifts on brainly pages.

Closes #2348


domidomi:
@clxandstuff what do you think about naming it something like `hasRenderedInitially` instead of `shouldUseAnimations`?

katarzynatobis (approved):

PR #2549, opened 2022-09-30
doc: fix spacing doc

reod:
Change readme to match [real values](https://github.com/brainly/style-guide/blob/master/src/sass/_config.scss#L45).


coderitual (approved):
Thanks @reod !

matzimowski (approved):

PR #2561, opened 2022-10-18
dialog bugfixing

clxandstuff:
Fixes:
- inner dialog is not closing when user clicks on the overlay.
- slots are now invisible when using a `fullscreen` size
- focus trap is working
- dialog content is rendered as last element to avoid potential problems when offsetting children elements out of the content box


coderitual (approved):
Let's merge it and handle focus trap issue separately

PR #2579, opened 2022-11-15
implement metrics CLI

clxandstuff:


bin/metrics/main.js
```diff
@@ -0,0 +1,106 @@
...
+const tsxParser = require('jscodeshift/parser/tsx');
+const {Lambda} = require('@aws-sdk/client-lambda');
+
+/* eslint-disable no-console */
+
+const lambda = new Lambda({
+  region: 'eu-west-1',
+});
+
+exports.main = async function (commitID, commitDate) {
```
coderitual, line 14: I should be able to call cli locally so it might be needed to decouple it from lambda call

bin/metrics/cli.js
```diff
@@ -0,0 +1,6 @@
+const {main} = require('./main');
```
coderitual, line 1: Please use shebang `#!/usr/bin/env node` in the first line

package.json
```diff
@@ -33,6 +34,8 @@
     "path": "^0.12.7"
   },
   "devDependencies": {
+    "@aws-sdk/client-dynamodb": "^3.202.0",
+    "@aws-sdk/util-dynamodb": "^3.204.0",
```
coderitual, line 38: why those deps?

bin/metrics/cli.js
clxandstuff: 👍

package.json
clxandstuff: We don't need them, I forgot to remove it.

bin/metrics/main.js
clxandstuff: I rewritten cli with `meow` and now you can display help message:

```
$ sg-metrics --help
```

To be able to get metrics but use it locally CLI now have flag `--dry` which logs result to `stdout`

test/main.spec.js
```diff
@@ -0,0 +1,124 @@
+import {main} from '../bin/metrics/main';
```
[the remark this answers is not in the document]
clxandstuff: good idea :)

bin/metrics/main.js
```diff
@@ -0,0 +1,124 @@
...
+      packageParsed.dependencies['brainly-style-guide']) ||
+    null
+  );
+}
+
+function getComponents() {
+  const elements = [];
+  const filePaths = glob.sync('**/*.{js,ts}x', {
+    ignore: 'node_modules/**',
+    cwd: process.cwd(),
```
coderitual, line 70: does it treat cwd as a root folder where search starts? We sometimes my need to search specific folder only (`src` etc)

coderitual (approved):

bin/metrics/main.js
clxandstuff: Yes. Now it search all folders from cwd. I can add option for providing search root paths. What do you think?

coderitual: Probably I would add it, in codemod cli we have it (but it is not immediately visible since it's passed to internal command)

coderitual: it can be normal argument even so:
`npx sg-metrics --dry src/*`

clxandstuff: cli currently requires 2 arguments, so it would be better to keep it as parameter like `--roots /root1 /root2`

clxandstuff: I'm also thinking about automatically retrieving commit hash and date. This could open way to have path as single parameter

coderitual: ok, it might be named parameter if you prefer that way, i would use `--path` rather than --roots though

bin/metrics/cli.js
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
+
+const {main} = require('./main');
+const meow = require('meow');
+
+const command = meow({
+  description: 'CLI tool for collecting styleguide metrics.',
+  help: `
+    Usage
+      $ yarn sg-metrics <...options> paths
```
coderitual, line 10: Following bash conventions it should be:
`yarn sg-metrics [options] <paths>`

[] - optional param
<> - mandatory param

coderitual: @clxandstuff

coderitual (approved):

bin/metrics/cli.js
clxandstuff: 👍

PR #2590, opened 2022-11-29
Fix textarea losing focus after error message is displayed

matzimowski:
It's the same [issue](https://github.com/brainly/style-guide/pull/2550/) we spotted and fixed for `Input` component.

Preview:

https://user-images.githubusercontent.com/1231144/204576617-24321a16-03fb-43ba-bb68-aa52d63fe0f3.mov

### DIFF
--- a/src/components/form-elements/Textarea.jsx
+++ b/src/components/form-elements/Textarea.jsx
@@ -138,7 +138,7 @@ const Textarea = (props: TextareaPropsType) => {
   const errorMessageDisplayed =
     invalid === true && errorMessage !== undefined && errorMessage !== '';
 
-  return errorMessageDisplayed ? (
+  return (
     <div className={wrapperClass}>
       <Type
         {...additionalProps}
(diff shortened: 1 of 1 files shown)


Malinicz (approved):

coderitual (approved):

przytrzask (approved):

src/components/form-elements/_textarea.scss
```diff
@@ -130,6 +130,8 @@ $includeHtml: false !default;
     }
 
     &__wrapper {
+      max-width: min-content;
```
matzimowski, line 133: the same property was set while using  `sg-input__wrapper`

coderitual:
hey @matzimowski I think we should to keep inline-block element behaviour in place (so it follows other form elements), I guess inline-flex would cause the same issue but I am wondering if this also happens to other controls - if so we can keep inline-block behaviour.

coderitual:
I realized that we have used display `block` in text are all the time (which is kind of wrong if we speak about default)

matzimowski:
@coderitual agree, i reverted my change.

coderitual (approved):

coderitual:
Chromatic stuck for some reason (bug), i will try to manually rebuild it

PR #2595, opened 2022-12-06
metrics cli add support for string globs

clxandstuff:
CLI can throw an error when globs are expanded and script gets too many arguments. For those cases we can use quotet globs so they will be expanded by script.

`--ignore "pattern"` option to ignore files, default is "node_modules"


test-project/package.json
```diff
@@ -0,0 +1,200 @@
+{
```
coderitual, line 1: why package json heree?

coderitual (changes requested):
package.json in test folder?

test-project/package.json
clxandstuff: Test code 😄

PR #2599, opened 2022-12-09
Styleguide code sandbox

clxandstuff:
Code sandbox:
 - new story under `Tools/Live Editor` with sandpack
 - build scripts refactored, deploy URLs are parameters and not hardcoded like it was in `.babelrc` and `scripts/build-s3.js`
 - new webpack build for sandbox assets (CSS, JS, icons etc.)
 - clean up `src` folder from generated files, they are stored in `dist` folder (I couldn't get it done for icons, so they are still in `src/images`)


coderitual:
Can we hide this button until it works (it redirects to code sandbox with the code, but without proper installation of deps it doesn't give much of a value)
CleanShot 2022-12-13 at 09 24 23

Small issue on the left panel
CleanShot 2022-12-13 at 09 23 44

coderitual (approved):

PR #2608, opened 2022-12-16
Add useId hook to prevent SSR hydration ID mismatch

matzimowski:
**Issue:**

Server vs client mismatch of generated id

Screenshot 2022-12-16 at 10 06 05

**Notes:**
- slightly 😉  inspired by `@reach/auto-id` lib
- released as `v220.7.2-alpha.1` and tested in web app


coderitual:
Closing because #2616

PR #2626, opened 2023-01-05
Change Rating component - change icon type, spacings and UI of active state

ElaLukaszewicz:
- change spacing between number of votes and stars and between rating and stars from 6px to 4px
- change star type to star_outlined when empty
- fix bug -> '0' visible before rating when it equals 0

**BUG FIX**
Before:
image

After:
image

**CHANGED SPACINGS**
Before:
image

After:
image

**ACTIVE STATE**
Before:

https://user-images.githubusercontent.com/39903772/210737805-d37681fb-02d3-48e0-be81-088af2b57048.mov

After (UI for active is done as it's in the custom `Rating` component on QPage currently):

https://user-images.githubusercontent.com/39903772/210737818-edda2c7f-8df9-4e92-8117-3f22e7ca3737.mov

### DIFF
--- a/src/components/rating/Rating.jsx
+++ b/src/components/rating/Rating.jsx
@@ -174,7 +174,7 @@ class Rating extends React.Component<RatingPropsType> {
     }
 
     return (
-      <div className={ratingClass} aria-label={label}>
+      <div className={ratingClass} aria-label={label} role="group">
         <p className="sg-rate-box__rate">
           {!noLabel && <span aria-hidden>{rateString}</span>}
           {Boolean(rate) && (
--- a/src/components/rating/Rating.jsx
+++ b/src/components/rating/Rating.jsx
@@ -5,7 +5,6 @@ import classnames from 'classnames';
 import Star from './subcomponents/Star';
 import RateCounter from './subcomponents/RateCounter';
 import {__DEV__, invariant} from '../utils';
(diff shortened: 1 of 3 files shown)


src/components/rating/Rating.jsx
```diff
@@ -174,10 +174,12 @@ class Rating extends React.Component<RatingPropsType> {
     }
 
     return (
-      <div className={ratingClass} aria-label={label}>
+      <div className={ratingClass} aria-label={label} role="group">
```
ElaLukaszewicz, line 177: not sure it's the correct role for this component, wdyt?

coderitual: I don't think we need group here

ElaLukaszewicz: a11y tests are failing without this 'role' 
image

coderitual: Oh I see, i am wondering why it didn't scream previously, maybe @katarzynatobis has some idea

katarzynatobis: `group` doesn't seem to fit here. I would try a different approach - `<span className="sg-visually-hidden">{label}</span>` as a first child of that div

katarzynatobis (changes requested):

katarzynatobis (approved):

PR #2632, opened 2023-01-13
fix Checkbox unnecessary render

### LINKED ISSUE #2601: Checkbox (and radio?) animation played on 1 render / rerender
There is one case found where animation of checkbox transition is played and it shouldn't. It might be specific case in application causes that but it seems it has something to do with internal animation mechanism of component. This case might be also present in radio component but it needs to be investigated.

https://user-images.githubusercontent.com/8572321/207032855-71480b18-139e-45b4-91ab-97c8cabe0370.mp4

clxandstuff:
This change fixes unnecessary animation on multiple checkboxes.

Fixes #2601

### DIFF
--- a/src/components/form-elements/checkbox/Checkbox.stories.mdx
+++ b/src/components/form-elements/checkbox/Checkbox.stories.mdx
@@ -1,3 +1,4 @@
+import {useState} from 'react';
 import {ArgsTable, Meta, Story, Canvas} from '@storybook/addon-docs';
 import {StoryVariant, StoryVariantTable} from '../../../docs/utils';
 import {styled} from '@storybook/theming';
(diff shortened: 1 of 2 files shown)


domidomi (approved):

src/components/form-elements/checkbox/Checkbox.stories.mdx
```diff
@@ -432,6 +434,69 @@ In the following example, className `sg-checkbox--custom-theme` was applied to a
...
+        const newData = {...data};
+        newData[id] = {...data[id], value: !data[id].value};
+        setData(newData);
+      };
+      return (
+        <Flex direction="column">
+          <div>
+            <Button onClick={() => onChange(101)} size="s">
+              click to toggle first checkbox
+            </Button>
```
katarzynatobis, line 483: I think this button is unnecessary - you can just click the first checkbox :D

domidomi: @katarzynatobis it is to change the active checkbox without clicking it

PR #2634, opened 2023-01-17
Radio and RadioGroup initial animation fixes on controlled state. Checkbox cleanup

domidomi:
Checkbox:
- Remove unnecessary isFirstRender check

RadioGroup and Radio:
- Do not apply Radio animation until either Radio checked state or RadioGroup value is changed
- Add stories with controlled RadioGroup

useIsFirstRender:
- remove hook


src/components/form-elements/radio/_radio.scss
```diff
@@ -142,6 +142,7 @@ $radioCheckedBorderWidth: 5px;
     height: $radioSize;
     border-radius: 50%;
     box-shadow: 0px 0px 0px var(--radioBorderWidth) var(--radioRingColor) inset;
+    transition: box-shadow 0s;
```
clxandstuff, line 145: Why do we need this?

domidomi: We don't have exit animation on the radio (when radio becomes unchecked), so I had to disable transition on the ring thickness change.

clxandstuff (approved):

PR #2665, opened 2023-01-30
New Select component

domidomi:


clxandstuff (changes requested):

src/components/select/Select.spec.tsx
```diff
@@ -0,0 +1,166 @@
...
+    expect(select.queryByRole('listbox')).toBeTruthy();
+  });
+
+  it('can select single option', async () => {
+    const select = render(<RenderSelect />);
+    const selectElement = select.getByRole('combobox') as HTMLElement;
+
+    expect(select.getByText('Select...')).toBeTruthy();
+    userEvent.click(selectElement);
+    await waitFor(() => expect(select.queryByRole('listbox')).toBeTruthy());
```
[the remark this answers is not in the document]
coderitual: findBy should be used here.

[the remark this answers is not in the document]
clxandstuff: You're right, `findBy` is better. 👍

coderitual:
how about removing user selection from items?
https://user-images.githubusercontent.com/8572321/225870565-2093769c-c122-4a2d-9c16-ea98c863b7be.mp4

coderitual:
I found similar issue to the ano above - cursor changes to text inside select input
https://user-images.githubusercontent.com/8572321/225872134-98a9220d-c90f-41b8-9af6-5be2981c5227.mp4

coderitual:
I think this story (positioning) is a bit confusing. Especially second and third row. wdyt?
https://user-images.githubusercontent.com/8572321/225873190-2993ee0a-12ed-465a-9a2e-5a8f6cbd9dea.mp4

src/components/select-menu/useSelectMenuAnimations.tsx
```diff
@@ -0,0 +1,199 @@
...
+      const initialContainerSize = lastRef.current;
+      const selectElementSize = selectRef.current;
+
+      popupContent.style.width = `${initialContainerSize.width}px`;
+      popupContent.style.height = `${initialContainerSize.height}px`;
+      popupContent.classList.add(SCROLL_HIDE_CLASSNAME);
+
+      if (!hasReduceMotion) {
+        // Reset the popup height to the pre-appear position
+        popupContainer.style.height = `1px`;
```
coderitual, line 111: why 1 px?

src/components/select-menu/useSelectMenuAnimations.tsx
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        floatingContainer.classList.add(OPEN_CLASSNAME);
+
+        popupContainer.style.opacity = `1`;
+
+        popupContainer.style.height = `${initialContainerSize.height}px`;
+        const popupWidth: number = Math.max(
+          initialContainerSize.width,
+          selectElementSize.width * 0.7,
```
coderitual, line 137: 0.7 should be named

src/components/select-menu/useSelectMenuAnimations.tsx
```diff
@@ -0,0 +1,199 @@
...
+            cleanupAfterTransition(popupContent);
+
+            popupContainer.removeEventListener(
+              'transitionend',
+              handleTransitionEnd
+            );
+          }
+        };
+
+        popupContainer?.addEventListener('transitionend', handleTransitionEnd);
```
coderitual, line 184: why null coalescing in `popupContainer?`

src/components/select-menu/useSelectMenuAnimations.tsx
domidomi: hmm I had some issues with the animation in the past when it was set to 0, but it seems it works all fine now with 0! Thanks for catching this

src/components/select-menu/useSelectMenuAnimations.tsx
```diff
@@ -0,0 +1,199 @@
...
+type UseSelectMenuAnimationsPropsType = {
+  selectId: string;
+  popupClassName: string;
+  popupContentClassName: string;
+  floatingContainerClassName: string;
+  selectElementClassName: string;
+};
+
+const MIN_POPUP_WIDTH = 120;
+const ANIMATE_CLASSNAME = 'animate-on-transforms';
```
coderitual, line 13: hmm not sure if we shouldn't use `sg-*` prefix here as in ohter global classes

domidomi:
> I think this story (positioning) is a bit confusing. Especially second and third row. wdyt?
> 
> do we need this story for chromatic or smth? It's aslo not covering most of the cases. There is one more story when we can test all positionings - with space constraints.
> 
> I think we should change the name of "with space constraints." to mentioned "positioning" and maybe remove the current positioning unless we need it in chromatic

@coderitual right, I don't think we actually need this story now, as it was more useful for debugging during development stage

src/components/select-menu/SelectMenu.chromatic.stories.tsx
```diff
@@ -0,0 +1,152 @@
...
+export const DifferentPopupLenghts = args => {
+  return (
+    <Flex direction="row" style={{gap: '40px'}}>
+      <div style={{width: '200px'}}>
+        <Text>Popup with wide content:</Text>
+        <SelectMenu
+          options={[
+            {
+              value: 'tv',
+              label: 'Usłyszałem o Brainly w reklamie telewizyjnej w telewizji',
```
coderitual, line 37: Why PL lang?

src/components/select-menu/SelectMenuOption.tsx
```diff
@@ -0,0 +1,109 @@
...
+import SubjectIcon from '../subject-icons/SubjectIcon';
+import Icon from '../icons/Icon';
+import Checkbox from '../form-elements/checkbox/Checkbox';
+import Text from '../text/Text';
+
+export type SelectOptionElementPropsType = {
+  option: SelectMenuOptionType;
+  isSelected?: boolean;
+  withIcon?: boolean;
+  multiSelect?: boolean;
```
coderitual, line 17: this shouldn't be an option of select option but should be defined on menuselect level and passed via context

coderitual (changes requested):

src/components/select-menu/useSelectMenuAnimations.tsx
domidomi: It was used in the figma file, but I can totally change it to eng

src/components/select-menu/SelectMenu.chromatic.stories.tsx
domidomi: It was used in the figma file, but I can totally change it to eng

src/components/select-menu/SelectMenu.tsx
```diff
@@ -0,0 +1,503 @@
...
+
+export type SelectMenuSizeType = 's' | 'm' | 'l';
+
+export const SIZE = {
+  S: 's',
+  M: 'm',
+  L: 'l',
+} as const;
+
+const DEFAULT_SIZE = SIZE.M;
```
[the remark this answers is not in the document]
domidomi: it's also used for a check in line :398

src/components/select-menu/SelectMenu.chromatic.stories.tsx
coderitual: I see, please ask designer to change it accordingly in figma. Our base lang is always eng. We should only use different lang if we prepare something specific to one market.

src/components/select-menu/SelectMenuOption.tsx
coderitual: Ok i see why is that. In the near future we should migrate json based configuration to composition so the component is used more like:

```tsx
 <Select>
    <Select.Group label="Label">
      <Select.Option value="myvalue">Value</Select.Option>
    </Select.Group>
    <Select.Group label="Code">
      <Select.Option value="a">A/Select.Option>
      <Select.Option value="b">B</Select.Option>
      <Select.Option value="c">C</Select.Option>
    </Select.Group>
  </Select>
```

src/components/select-menu/SelectMenu.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+  });
+
+  it('can select multiple options in multi select Select', async () => {
+    const select = render(<RenderSelectMenu multiSelect />);
+    const selectElement = select.getByRole('combobox') as HTMLElement;
+
+    expect(selectElement.getAttribute('aria-multiselectable')).toBeTruthy();
+    userEvent.click(selectElement);
+    await select.findByRole('listbox');
+    fireEvent.click(select.getByText('Physics'));
```
coderitual, line 111: why fire event and not userEvent?

src/components/select-menu/SelectMenu.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+    const select = render(<RenderSelectMenu />);
+    const selectElement = select.getByRole('combobox') as HTMLElement;
+
+    userEvent.click(selectElement);
+    await select.findByRole('listbox');
+    fireEvent.click(select.getByText('Physics'));
+
+    await waitFor(() => expect(select.queryByRole('listbox')).toBeFalsy());
+    expect(select.queryByText('Select...')).toBeFalsy();
+    expect(select.getByText('Physics')).toBeTruthy();
```
coderitual, line 89: Why not using `toBeInTheDocument`?

src/components/select-menu/SelectMenuOption.tsx
```diff
@@ -0,0 +1,109 @@
...
+          <Icon
+            size={24}
+            color={isSelected || isHovered ? 'icon-black' : 'icon-gray-50'}
+            type={regularIconName}
+          />
+        );
+      }
+    }, [isSelected, isHovered, withIcon, icon]);
+
+    const optionState = React.useMemo(() => {
```
coderitual, line 73: why use memo?

src/components/select-menu/_select-menu.scss
```diff
@@ -0,0 +1,329 @@
+@if ($includeHtml) {
+  .sg-select-menu {
+    --select-background: #{$gray-20};
```
coderitual, line 3: we should start using css variables as a values for tokens instead of scss variables. This shift will be needed in the future to support dark mode.

cc @clxandstuff @katarzynatobis

src/components/select-menu/SelectMenu.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+window.HTMLElement.prototype.scroll = jest.fn();
+
+describe('<SelectMenu />', () => {
+  it('renders SelectMenu in the default state', () => {
+    const select = render(<RenderSelectMenu />);
+    const selectElement = select.getByRole('combobox') as HTMLElement;
+
+    expect(select.getByText('Select...')).toBeTruthy();
+    expect(selectElement.getAttribute('aria-expanded')).toBe('false');
+    expect(selectElement.getAttribute('aria-disabled')).toBeFalsy();
```
coderitual, line 60: As a general rule we should be as explicit as possible. Please don't overuse `toBeFalsy` and `toBeTruthy` and use `toBe`, `toBeUndefined`, `toEqual`, `toHaveAttribute` etc.

In this specific case - testing `aria-disabled` feels inconsitend across all tests.

src/components/select-menu/SelectMenu.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+    const selectElement = select.getByRole('combobox') as HTMLElement;
+
+    userEvent.click(selectElement);
+    expect(selectElement.getAttribute('aria-expanded')).toEqual('true');
+    select.findByRole('listbox');
+    expect(select.getByRole('listbox').getAttribute('id')).toBe(
+      select.getByRole('combobox').getAttribute('aria-controls')
+    );
+    expect(
+      within(select.getByRole('listbox')).getAllByRole('option').length
```
coderitual, line 75: I don't understand this test. Are we checking if length is equal , true, 1,2,3,4 etc?

src/components/select-menu/SelectMenu.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+    userEvent.click(selectElement);
+    await select.findByRole('listbox');
+    fireEvent.click(select.getByText('History'));
+    await waitFor(() => expect(select.queryByRole('listbox')).toBeFalsy());
+    expect(select.getByText('History')).toBeTruthy();
+
+    userEvent.click(selectElement);
+    await waitFor(() => expect(select.queryByRole('listbox')).toBeTruthy());
+    expect(
+      within(select.getByRole('option', {selected: true})).getByText('History')
```
coderitual, line 100: I think this test might be more readable? If we get particular element and compare it to the ohter it's good practise to keep those elements referenced in variables.

e.g
```ts
const option = select.getByRole('option', {selected: true});
const item = within(option).getByText('History');
expect(item).toBeVisible();
```

src/components/select-menu/SelectMenu.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+    await waitFor(() =>
+      expect(selectElement.getAttribute('aria-expanded')).toEqual('true')
+    );
+    const selectedOption1 = select.getByRole('option', {name: 'Physics'});
+
+    expect(selectedOption1).toHaveFocus();
+    expect(selectedOption1).toEqual(document.activeElement);
+    userEvent.keyboard('{space}');
+    expect(selectedOption1.getAttribute('aria-selected')).toEqual('true');
+    await waitFor(() => expect(select.queryByRole('listbox')).toBeFalsy());
```
coderitual, line 164: waitFor is used on one expect and missing for the other `expect(selectedOption1.getAt...`, is it on purpose?

src/components/select-menu/SelectMenu.spec.tsx
coderitual: it's better to use toHaveLength in this case or toBe in equivalent form

src/components/select-menu/SelectMenu.spec.tsx
```diff
@@ -0,0 +1,220 @@
...
+}
+
+window.HTMLElement.prototype.scroll = jest.fn();
+
+describe('<SelectMenu />', () => {
+  it('renders SelectMenu in the default state', () => {
+    const select = render(<RenderSelectMenu />);
+    const selectElement = select.getByRole('combobox') as HTMLElement;
+
+    expect(select.getByText('Select...')).toBeTruthy();
```
coderitual, line 58: Maybe it's better to just start from this part?

We can use:
```
const select = screen.getByText('Select...');
const selectElement = screen.getByRole('combobox') as HTMLElement;
```

let's discuss @domidomi @clxandstuff @katarzynatobis

package.json
```diff
@@ -27,6 +27,7 @@
     "@aws-sdk/client-lambda": "^3.210.0",
     "@codesandbox/sandpack-react": "^1.17.0",
     "@codesandbox/sandpack-themes": "^1.17.0",
+    "@floating-ui/react": "^0.18.1",
```
coderitual, line 30: Probably we can upgrade to `0.21.1`

src/components/select-menu/useSelectMenuAnimations.tsx
domidomi: Some leftover, thanks for catching!

coderitual (approved):
All points were discussed and are non blocking

clxandstuff (approved):

PR #2675, opened 2023-02-13
Sparks component ✨

coderitual:
sparks

https://user-images.githubusercontent.com/8572321/220912887-a21dfaaa-2c62-4b03-95f0-98720ebce158.mp4

https://user-images.githubusercontent.com/8572321/220912907-e1306516-96fe-44aa-a01e-1e6872866c48.mp4

https://user-images.githubusercontent.com/8572321/220912916-931653f0-3774-499f-b47e-e7a8b5c1e3d2.mp4

CleanShot 2023-03-01 at 11 11 45

### DIFF
--- a/.eslintrc.rules.js
+++ b/.eslintrc.rules.js
@@ -507,7 +507,7 @@ module.exports = {
     // require method and property shorthand syntax for object literals
     'object-shorthand': ['error', 'always'],
     // suggest using of const declaration for variables that are never modified after declared
-    //'prefer-const': 'error',
+    'prefer-const': 'error',
     // require destructuring from arrays and/or objects
     'prefer-destructuring': 'off',
     // disallow parseInt() in favor of binary, octal, and hexadecimal literals
--- a/src/components/sparks/Particle.tsx
+++ b/src/components/sparks/Particle.tsx
@@ -3,7 +3,6 @@ import * as React from 'react';
 export interface ParticleCSSProperties extends React.CSSProperties {
   '--particle-dir'?: number;
   '--particle-offset'?: number;
-  '--index'?: number;
 }
 
 export interface ParticleProps {
--- a/src/components/sparks/Particle.tsx
+++ b/src/components/sparks/Particle.tsx
@@ -45,5 +45,4 @@ const Particle = React.forwardRef<HTMLDivElement, ParticleProps>(
   }
 );
 
-Particle.displayName = 'Particle';
 export default Particle;
--- a/src/components/sparks/Particle.tsx
+++ b/src/components/sparks/Particle.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 export interface ParticleCSSProperties extends React.CSSProperties {
   '--particle-dir'?: number;
   '--particle-offset'?: number;
+  '--rotation-dir'?: number;
 }
 
 export interface ParticleProps {
--- a/src/components/sparks/Sparks.spec.tsx
+++ b/src/components/sparks/Sparks.spec.tsx
@@ -41,7 +41,7 @@ describe('<Sparks />', () => {
 
     userEvent.click(button);
 
-    expect(onClick).not.toHaveBeenCalled();
+    expect(onClick).toHaveBeenCalled();
   });
 
   it('has no a11y violations', async () => {
--- a/src/components/sparks/Sparks.spec.tsx
+++ b/src/components/sparks/Sparks.spec.tsx
@@ -3,6 +3,17 @@ import {render} from '@testing-library/react';
 import {testA11y} from '../../axe';
 import Sparks from './Sparks';
 
+// mock all animations WAAPI so tests run correctly in jsdom environment (no window)
+jest.mock('./animation', () => ({
+  useAnimation: jest.fn(() => ({
+    animation: {
+      play: jest.fn(),
+      cancel: jest.fn(),
+    },
+    animationRef: React.createRef(),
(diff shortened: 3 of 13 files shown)


src/components/sparks/Particle.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+
+export interface ParticleCSSProperties extends React.CSSProperties {
+  '--particle-dir'?: number;
+  '--particle-offset'?: number;
+  '--index'?: number;
```
clxandstuff, line 6: What do you think about prefixing it like other vars `--particle-index` ?

BTW I can't see the usage of this var,  let's remove it then.

src/components/sparks/Particle.tsx
```diff
@@ -0,0 +1,50 @@
...
+          viewBox="0 0 512 512"
+        >
+          {shapes[shape]}
+        </svg>
+      </div>
+    );
+  }
+);
+
+Particle.displayName = 'Particle';
```
clxandstuff, line 49: I think we can remove it due `@babel/plugin-transform-react-display-name` is doing that for us. We have it as a part of [@babel/preset-react](https://babeljs.io/docs/babel-preset-react)

src/components/sparks/Sparks.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Sparks>
+        <button type="button">Click me</button>
+      </Sparks>
+    );
+
+    expect(sparks.container.firstChild).toBeTruthy();
+    expect(sparks.getByRole('button', {name: /click me/i})).toBeTruthy();
+  });
+
+  it('renders around interactive element and this element still captures all the events', () => {
```
clxandstuff, line 30: I would rephrase it a little bit 'renders around interactive element and this element still can be clicked' because we actually testing it but not all events.

src/components/sparks/Sparks.tsx
```diff
@@ -0,0 +1,173 @@
...
+              />
+            )
+          )}
+        </div>
+      </div>
+    );
+  }
+);
+
+Sparks.displayName = 'Sparks';
```
clxandstuff, line 175: I think we can remove it due @babel/plugin-transform-react-display-name is doing that for us. We have it as a part of [@babel/preset-react](https://babeljs.io/docs/babel-preset-react)

src/components/sparks/Sparks.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+describe('<Sparks />', () => {
+  it('renders correctly and displays its children', () => {
+    const sparks = render(
+      <Sparks>
+        <button type="button">Click me</button>
+      </Sparks>
+    );
+
+    expect(sparks.container.firstChild).toBeTruthy();
```
clxandstuff, line 26: If you add role on svg ('img') , you can use getByRole and check more accurately sparks.

clxandstuff (changes requested):

src/components/sparks/Particle.tsx
coderitual: Yes it is leftover, thx for catching

src/components/sparks/Sparks.tsx
coderitual: True

src/components/sparks/Sparks.spec.tsx
coderitual: Yea, honestly the meaning of this test is vague (checks if element just renders anything)

I will remove it once I figure out chromatic tests

src/components/sparks/Particle.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+
+export interface ParticleCSSProperties extends React.CSSProperties {
+  '--particle-dir'?: number;
```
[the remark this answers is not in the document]
coderitual: It controls direction of the particle - up or down represented  by 1/-1. Thanks to that direction is controlled entirely within key frame (—particle-dir * <translateinpx>). Some particles are going from top to bottom and some the opposite

.eslintrc.rules.js
```diff
@@ -506,7 +507,7 @@ module.exports = {
     // require method and property shorthand syntax for object literals
     'object-shorthand': ['error', 'always'],
     // suggest using of const declaration for variables that are never modified after declared
-    'prefer-const': 'error',
```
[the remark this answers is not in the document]
coderitual: Temp, I want to discuss with you next week

src/components/sparks/Sparks.spec.tsx
coderitual: As for role img it is pointless here since all contaniner is hidden in accesibility tree. The only thing I care about is that component renders. Expect is just used to satisfy test.

prpbably this test can be skipped because existence of 2 other tests, it was just the only test for some time.

clxandstuff: That's the best case if any other tests cover this we can get rid of this artificial "it renders" tests

src/components/sparks/Particle.tsx
clxandstuff: You missed it or should we keep it here? I see that in Sparks same part has been removed.

coderitual: Sorry, missed it, fixed now

clxandstuff (approved):

PR #2686, opened 2023-03-07
add PageHeader to addon-docs config

katarzynatobis:
`PageHeader` added to addon-docs config in `.storybook/preview.js` to avoid mistakes associated with missing import.


coderitual:
Can we discuss tomorrow? I am not entirely sure about this direction.

PR #2712, opened 2023-03-31
SelectMenu a11y tests

katarzynatobis:


src/components/select-menu/SelectMenu.tsx
```diff
@@ -421,10 +420,9 @@ const SelectMenu = React.forwardRef<HTMLDivElement, SelectMenuPropsType>(
           tabIndex={!disabled ? 0 : -1}
           aria-disabled={disabled}
           aria-invalid={invalid ? true : undefined}
-          aria-controls={`${id}-listbox`}
+          aria-controls={interactionsFloatingProps.id as string}
```
coderitual, line 423: what's the type of `id` so we must cast?

coderitual (approved):

PR #2726, opened 2023-04-19
Fix #2713 - search unmounts unexpectedly

### LINKED ISSUE #2713: Search component's button unmounts unexpectedly
## Problem
In a search component we have following conditional rendering 
```ts
{withRoundButton ? (
        <div className={`${baseClassName}__icon`}>
          <Button
            variant="solid"
            className={cx({
              'sg-search-button--s': size === 's',
            })}
            icon={
              <Icon
                type="search"
                size={size === 'l' ? 24 : 16}
                color="adaptive"
              />
            }
            iconOnly
            size={size === 'l' ? 'm' : 's'}
          />
        </div>
      ) : (
        <button className={`${baseClassName}__icon`}>
          <Icon
            type="search"
            color={ICON_COLOR['icon-gray-50']}
            size={size === 'l' ? 24 : 16}
          />
        </button>
      )}
```

and since we use this component based on whether it is focused or not like this 
```ts
turn (
    <Search
      withRoundButton={focused}
```

then because blur happens first, the button unmounts next submit is never called

CleanShot 2023-04-03 at 09 25 40

### Proposed solution

I think it will be rather easy to style this button using classes only instead of conditional rendering

katarzynatobis:
closes #2713

### DIFF
--- a/src/components/search/Search.tsx
+++ b/src/components/search/Search.tsx
@@ -77,8 +77,8 @@ const Search = ({
         variant={withRoundButton ? 'solid' : 'transparent-light'}
         className={cx([
           `${baseClassName}__icon`,
+          `sg-search-button--${size}`,
           {
-            'sg-search-button--s': size === 's',
             [`${baseClassName}__icon--transparent`]: !withRoundButton,
           },
         ])}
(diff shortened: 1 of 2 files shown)


clxandstuff:
Since you removed outer `div`, you should add label https://github.com/brainly/style-guide/labels/Change%3A%20Internal%20Layout

clxandstuff (approved):

PR #2732, opened 2023-04-25
New Tooltip component

domidomi:
Screenshot 2023-05-23 at 12 32 14
Screenshot 2023-05-23 at 12 32 07
Screenshot 2023-05-23 at 12 31 59
Screenshot 2023-05-23 at 12 31 33


src/components/tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,292 @@
...
+    <Tooltip defaultOpen={props.defaultOpen}>
+      <Tooltip.Element label="Copy to clipboard" />
+      <Tooltip.Trigger>{props.children}</Tooltip.Trigger>
+    </Tooltip>
+  );
+}
+
+// JSDom (which is used by jest) does not implement layout/rendering.
+// we create this mock to simply simulate a desktop view with a width of 1000
+function useResizeObserverMock() {
```
coderitual, line 27: let's keep "use*" prefix in naming strictly for hooks.

clxandstuff: This is a mock of `useResizeObserver`, so use* is fine in this case.

Have you tried to mock ResizeObserver instead? It would be more flexible and we can have full control over it.

Something like this:

```suggestion
global.ResizeObserver = jest.fn().mockImplementation(() => ({
    observe: jest.fn(),
    unobserve: jest.fn(),
    disconnect: jest.fn(),
}))
```

clxandstuff:
I've found strange behaviour when scrolling `Text Lengths` chromatic story. Tooltips sticked to bottom of the screen but I believe they should be hidden till we scroll to their trigger elements.

https://github.com/brainly/style-guide/assets/25419830/7069ddcc-0f27-464e-ad4a-314f7d48c610

src/components/tooltip/Tooltip.spec.tsx
coderitual: true, missed the part it's a mock for a hook, agree that it can be mocked deeper.

coderitual:
How about some background for `<p>` tag so it's more clear it's a block
CleanShot 2023-05-26 at 10 01 14

coderitual:
maybe different style here (minor)

image

⬇️
CleanShot 2023-05-26 at 10 06 18

src/components/tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+    const tooltipElement = tooltip.getByRole('tooltip') as HTMLElement;
+
+    expect(tooltipElement).toBeInTheDocument();
+
+    fireEvent.blur(button);
+    expect(tooltipElement).not.toBeInTheDocument();
+  });
+
+  it('hides tooltip when Esc key is hit', () => {
```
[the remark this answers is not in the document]
katarzynatobis: Esc should work also for the Tooltip triggered by hover, it is [WCAG success criterion 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)

katarzynatobis (approved):

PR #2744, opened 2023-05-11
new useIsFirstRender

clxandstuff:
useIsFirstRender rewritten and used in Checkbox and Radio

### DIFF
--- a/src/components/utils/useIsFirstRender.ts
+++ b/src/components/utils/useIsFirstRender.ts
@@ -1,14 +1,17 @@
-import {useRef} from 'react';
+import React from 'react';
 
-const useIsFirstRender = () => {
-  const isFirstRender = useRef(true);
+export const useIsFirstRender = () => {
+  const [isFirstRender, setIsFirstRender] = React.useState(true);
 
-  if (isFirstRender.current === true) {
-    isFirstRender.current = false;
-    return true;
-  }
+  React.useLayoutEffect(() => {
+    const raf = window.requestAnimationFrame(() => {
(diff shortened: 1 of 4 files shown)


src/components/utils/useIsFirstRender.ts
```diff
@@ -1,14 +1,17 @@
...
-const useIsFirstRender = () => {
-  const isFirstRender = useRef(true);
+export const useIsFirstRender = () => {
+  const [isFirstRender, setIsFirstRender] = React.useState(true);
 
-  if (isFirstRender.current === true) {
-    isFirstRender.current = false;
-    return true;
-  }
+  React.useLayoutEffect(() => {
```
coderitual, line 6: hey @clxandstuff , we need isomorphic version here

coderitual:
i see that you're still using state to communicate about first render (inside first render hook). I thin this may cause some side effects. Let's take a look into tests.

src/components/utils/useIsFirstRender.ts
clxandstuff: Done

clxandstuff:
@coderitual

Take a look now. I removed states related to animation from `isFirstRender`, `Checkbox`, and `Radio`.

coderitual:
hey @clxandstuff I see 2 problems with the example below:
- using ref directly in render (we can discuss)
- is naming of this ref correct?

CleanShot 2023-05-17 at 11 16 59

coderitual (approved):

PR #2777, opened 2023-09-15
Replace title tag in SVG with aria-labelledby

iamjasinski:
The only thing that this PR does is to replace title tag in SVGs with aria-labelledby. 
Why? We had [complaints from SEO team](https://docs.google.com/document/d/1KzIjV0STv5JiXzpdeVMgaYQhcbItBWEldUgnPU6ciyI/edit#heading=h.3c0es9m2ixg2) about this and their recommendation was to use one unified title tag or delete it. 
I investigated that and [official documentation ](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title)says that title tag is not recommended and we should use aria-labelledby so I replaced it in every SVG with title tag.


src/images/icons/add_more.svg
```diff
@@ -1,7 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
-  <title>
-    add more
-  </title>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512" aria-labbeledby="add more">
```
daria-marczak, line 1: this should not be `aria-labelledby` (btw. there's a typo here) but `aria-label`. [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) needs another item with the id provided between "" to be able to label this element

jasinskidev: That what's happening when you blindly copy paste from the docs without thinking 😄 Yes I know, I will change it, thanks!

daria-marczak: but to be honest I'm not sure if the svg does actually need any label - it just has a presentational role. I'd suggest to give it `alt=""` as this suggests to the screen readers that it doesn't matter for the page content

matzimowski:
i'd would advise to remove it and do not replace it with another attribute, since we already remove it on a build step: https://github.com/brainly/style-guide/pull/2578/files

PR #2794, opened 2024-01-15
Dialog - useBodyOnScroll - Fix for nested dialogs case

matzimowski:
## before

two dialogs are open. one is nested in another. both are closed at the same time.  `"sg-dialog-no-scroll"` class is not removed from the document body and the whole page cannot be scrolled (due to `manyDialogsOpened` condition).

## after

two dialogs are open. one is nested in another. both are closed at the same time.  `"sg-dialog-no-scroll"` class is removed from the document body and the whole page is working as expected.

**Note:**
An [Issue ](https://github.com/brainly/style-guide/issues/2795) created to introduce an ultimate solution for nested components

### DIFF
--- a/src/components/dialog/Dialog.stories.mdx
+++ b/src/components/dialog/Dialog.stories.mdx
@@ -517,6 +517,10 @@ import {fireEvent, within} from '@testing-library/react';
       const handleDismissAnother = () => {
         setAnotherOpen(false);
       };
+      const handleCloseAll = () => {
+        setAnotherOpen(false);
+        setOpen(false);
+      };
       const headerId = 'dialog-header';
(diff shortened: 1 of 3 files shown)


bartoszlorek (approved):
I would implement the `useBodyNoScroll` with the context to be aware of the nested state. https://www.aleksandrhovhannisyan.com/blog/react-context-nested-components/

krzotki (approved):

matzimowski:
> I would implement the `useBodyNoScroll` with the context to be aware of the nested state. https://www.aleksandrhovhannisyan.com/blog/react-context-nested-components/

👌 i created this https://github.com/brainly/style-guide/issues/2795 issue to handle it the way you suggest

PR #2806, opened 2024-07-04
upgrade React to 17.0.2

krzotki:


package.json
```diff
@@ -23,12 +23,15 @@
     "node": ">=8.0.0",
     "yarn": "^1.0.0"
   },
+  "resolutions": {
+    "@types/react": "^17.0.40"
+  },
```
kopach, line 28: Why do we need this? What libraries do require this override?

src/components/accordion/Accordion.spec.tsx
```diff
@@ -226,7 +226,14 @@ describe('<Accordion>', () => {
 
     expect(item.getAttribute('aria-expanded')).toEqual('false');
     expect(accordion.queryByRole('region')).toBeNull();
+
+    // This one does not trigger onFocus method in React 17
     accordion.getByRole('button').focus();
+
+    // This one does not set document.activeElement
+    // https://github.com/jsdom/jsdom/issues/2586
+    fireEvent.focus(accordion.getByRole('button'));
```
kopach, line 235: what do you think we can do about this?

src/components/rating/Rating.tsx
```diff
@@ -83,6 +83,7 @@ class Rating extends React.Component<RatingPropsType> {
     (arg0: React.MouseEvent<HTMLSpanElement>) => unknown
   > = [];
 
+  // It's a typo, should be componentWillReceiveProps (UNSAFE_componentWillReceiveProps)
```
kopach, line 86: can we fix this? what is the solution here?

package.json
krzotki: https://stackoverflow.com/a/73005024

src/components/rating/Rating.tsx
krzotki: The solution would be to use `UNSAFE_componentWillReceiveProps` but this will change behaviour of the component as the previous `componentWillReciveProps` was never called.
Other solution would be to remove it completely

src/components/accordion/Accordion.spec.tsx
krzotki: No idea, I was not able to fix this differently.
The component works correctly though, it collapes on Enter/Space when motion is reduced:

https://github.com/brainly/style-guide/assets/52164548/7cb61829-9804-478b-be32-f66ddb8e82ae

package.json
kopach: yes yes, I understand what it does. the problem is, that basically now we override all the dependencies to use this specific react types. proper solution would be to identify which dependency doesn't use `@types/react@17` and possibly update that dependency

krzotki: Aight, lemme check

krzotki: I removed it, and it still works, maybe something with `yarn.lock`

src/components/accordion/Accordion.spec.tsx
kopach, line 235: I would suggest to move this into helper function (put it in the end of this file), use it here as 1 line and describe all the comments in that function. e.g.
```
focusElement(accordion.getByRole('button'));
```

and later

```js
function focusElement(element) {
    // we need both those calls for now due to defect in fireEvent.focus https://github.com/jsdom/jsdom/issues/2586
    // This alone does not trigger onFocus method
    element.focus(); 
    // This alone does not set document.activeElement due to bug
    fireEvent.focus(element);
}
```

This way it will be elegant in place of usage (and easy to replace later once defect fixed). And explanatory in the function if someone would like to dig deeper

krzotki: Fixed in https://github.com/brainly/style-guide/pull/2806/commits/768d6977da3d29ffcdaefd0cd91b83dd0b0d7c0c

PR #2807, opened 2024-07-09
React 18

krzotki:
Upgrade to react 18 and fix tests


krzotki (commented):
This has to be adjusted before proper release:
https://github.com/brainly/style-guide/blob/ce89f5797654776b19fb16d6133f3af1038fb337/src/components/dialog/Dialog.tsx#L53C8-L53C20

Because of:
```js
next-dev.js:20 Warning: Dialog: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
    at Dialog (webpack-internal:///../../../node_modules/brainly-style-guide/esm/components/dialog/Dialog.js:246:20)
```

src/components/utils/useReducedMotion.spec.ts
```diff
@@ -1,6 +1,7 @@
 import useReducedMotion from './useReducedMotion';
-import {act, renderHook} from '@testing-library/react-hooks';
+import {renderHook} from '@testing-library/react';
 import MatchMedia from './__mocks__/mq';
+import {act} from 'react';
```
kopach, line 4: I think, we should do similar refactoring as it was done here https://github.com/brainly/frontend-monorepo/pull/516
I mean, import `act` from `'@testing-library/react';` and remove `@testing-library/react-hooks` from the codebase

package.json
```diff
@@ -28,7 +28,7 @@
     "@codesandbox/sandpack-react": "^1.17.0",
     "@codesandbox/sandpack-themes": "^1.17.0",
     "@floating-ui/react": "^0.21.1",
-    "@types/react": "17.0.40",
+    "@types/react": "18.3.3",
```
kopach, line 31: why is this in `dependencies`? can we move this into `devDependencies`?

krzotki: Moved to dev deps

src/components/utils/useReducedMotion.spec.ts
kopach, line 4: get this `act` from `@testing-library/react';` instead, please

kopach (approved):

radekwarisch (approved):

PR #2810, opened 2024-07-19
Adjust FileHandler by removing target blank attribute.

ElaLukaszewicz:
# Remove target blank in FileHandler
According to a11y rules, link which downloads a file should not have target blank attribute. 

Before:
Screenshot 2024-07-19 at 12 45 37

After:
Screenshot 2024-07-19 at 12 37 07


matzimowski (approved):
👏

krzotki (approved):

PR #2812, opened 2024-07-26
Relax dependencies

kopach:


package.json
```diff
@@ -68,6 +67,7 @@
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^13.1.0",
     "@testing-library/user-event": "^13.5.0",
+    "@types/react": "18.3.3",
```
kopach, line 70: moved to devDependencies, otherwise we're forcing to install this exact version during install in dependent project

krzotki (approved):

